PR #1, opened 2021-07-07
Draft of README content of new polyfill repo

justingrant:
Added some initial README content

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,1 +1,61 @@
 # Temporal Polyfill
+
+## Polyfill for [TC39 Proposal: Temporal](https://github.com/tc39/proposal-temporal)
+
+This polyfill is maintained by the champions of the [Temporal proposal](https://github.com/tc39/proposal-temporal) and is intended be ready for production use by the time the Temporal proposal reaches Stage 4.
+
+This polyfill is compatible with Node.js 12 or later.
+
+## Roadmap
+
+* [x] Fork non-production polyfill from [tc39/proposal-temporal repo](https://github.com/tc39/proposal-temporal/tree/main/polyfill)
+* [x] Release initial pre-alpha to NPM at [@js-temporal/polyfill](https://www.npmjs.com/package/@js-temporal/polyfill)
+* [ ] Sync up the code in this repo with the handful of polyfill changes that have recently been made in the [tc39/proposal-temporal](https://github.com/tc39/proposal-temporal) repo
+* [ ] Release alpha version to NPM 
+* [ ] Deprecate all other earlier Temporal polyfills
+* [ ] Convert to TypeScript for better maintainability
+* [ ] (Maybe) Optimize performance of slow operations, notably non-ISO calendar calculations
+
+## Bug Reports and Feedback
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1 +1,61 @@
 # Temporal Polyfill
+
+## Polyfill for [TC39 Proposal: Temporal](https://github.com/tc39/proposal-temporal)
+
+This polyfill is maintained by the champions of the [Temporal proposal](https://github.com/tc39/proposal-temporal) and is intended be ready for production use by the time the Temporal proposal reaches Stage 4.
```
ptomato, line 5:
```suggestion
This polyfill is maintained by the champions of the [Temporal proposal](https://github.com/tc39/proposal-temporal) and is intended to be ready for production use by the time the Temporal proposal reaches Stage 4.
```
In general I think we ought to avoid committing to specific timelines. We could consider saying something else here, although "by Stage 4" seems OK enough.

We had talked in one of our previous meetings (can't remember when) about how most of the champions don't have time to do extensive maintenance on it. Maybe we should at least say it's maintained by "some of" the champions, but I wonder if there's a phrase that would be more encouraging towards people reading this to join up and help maintain it.

Consider also mentioning here that this polyfill is not associated with TC39.

README.md
```diff
@@ -1 +1,61 @@
...
+
+This polyfill is compatible with Node.js 12 or later.
```
ptomato, line 7: I think we can drop Node 12 in the new polyfill since 14.x is the current LTS.

README.md
```diff
@@ -1 +1,61 @@
...
+
+## Documentation
+
+Reference documentation and examples for the Temporal API can be found [here](https://tc39.es/proposal-temporal/docs/index.html).
+
+A cookbook to help you get started and learn the ins and outs of Temporal is available [here](https://tc39.es/proposal-temporal/docs/index.html)
+
+If you find a bug in the documentation, please file a bug over in the [tc39/proposal-temporal issue tracker](https://github.com/tc39/proposal-temporal/issues) issue tracker.
+
+Note that the Temporal documentation is in the process of being migrated to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript).
```
ptomato, line 33: As a bonus, we could link to the MDN issue in the tc39 issue tracker.

README.md
```diff
@@ -1 +1,61 @@
...
+To install:
+
+```bash
+$ npm install @js-temporal/polyfill
+```
+
+CJS Usage:
+
+```javascript
+const { Temporal } = require('@js-temporal/polyfill');
```
ptomato, line 46: Should we mention in the examples how to import the full set of functionality (off the top of my head, it's something like `const { Temporal, Intl, toTemporalInstant } = ...` followed by `Date.prototype.toTemporalInstant = toTemporalInstant`?

ptomato (approved):
This is great!

README.md
justingrant: Sounds good. I revised this text. Take a look at the new text and let me know what you think.

README.md
justingrant: Good idea. Done.

README.md
justingrant: 👍

PR #5, opened 2021-07-08
Prepare for alpha (0.1.0) release

justingrant:
* update version to 0.1.0
* copy recent code & test changes from original repo
* switch CI tests to run Node 14 and 16 (remove 15)
* run prettier --fix
* update dependencies
* update .gitignore

### DIFF
--- a/lib/ecmascript.mjs
+++ b/lib/ecmascript.mjs
@@ -150,6 +150,13 @@ const ES2020 = {
 
 export const ES = ObjectAssign({}, ES2020, {
   ToPositiveInteger: ToPositiveInteger,
+  ToFiniteInteger: (value) => {
+    const integer = ES.ToInteger(value);
+    if (!NumberIsFinite(integer)) {
+      throw new RangeError('infinity is out of range');
+    }
+    return integer;
+  },
   IsTemporalInstant: (item) => HasSlot(item, EPOCHNANOSECONDS) && !HasSlot(item, TIME_ZONE, CALENDAR),
   IsTemporalTimeZone: (item) => HasSlot(item, TIMEZONE_ID),
   IsTemporalCalendar: (item) => HasSlot(item, CALENDAR_ID),
--- a/lib/ecmascript.mjs
+++ b/lib/ecmascript.mjs
@@ -377,18 +377,8 @@ export const ES = ObjectAssign({}, ES2020, {
     return { years, months, weeks, days, hours, minutes, seconds, milliseconds, microseconds, nanoseconds };
   },
   ParseTemporalInstant: (isoString) => {
-    const {
-      year,
-      month,
-      day,
-      hour,
-      minute,
-      second,
-      millisecond,
-      microsecond,
-      nanosecond,
-      offset
-    } = ES.ParseTemporalInstantString(isoString);
+    const { year, month, day, hour, minute, second, millisecond, microsecond, nanosecond, offset } =
+      ES.ParseTemporalInstantString(isoString);
 
     const epochNs = ES.GetEpochFromISOParts(
       year,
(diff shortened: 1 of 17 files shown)


test/intl.mjs
```diff
@@ -600,8 +599,7 @@ describe('Intl', () => {
         });
         const withValues = getValues('with');
         const withErrorExpected =
-          withValues === RangeError ||
-          ((nodeVersion === '14' || nodeVersion === '12') && withValues.nodeBefore15 === RangeError);
+          withValues === RangeError || (nodeVersion === '12' && withValues.node12 === RangeError);
```
ptomato, line 602: Note: these we can probably eliminate if we're not supporting Node 12 anymore, but it's fine to leave as-is for now since we're just copying over the changes to the old polyfill.

test/plaintime.mjs
```diff
@@ -18,255 +18,6 @@ import * as Temporal from '@js-temporal/polyfill';
 const { PlainTime, PlainDateTime } = Temporal;
 
 describe('Time', () => {
-  describe('Structure', () => {
```
ptomato, line 21: I would suggest not porting over these removals (here and in PlainDate) — these are tests that Ms2ger converted into test262-style, but I think we should keep them in the production polyfill, at least for the time being. (The idea is that the Demitasse tests will gradually disappear from the old polyfill)

lib/ecmascript.mjs
```diff
@@ -2290,7 +2187,7 @@ export const ES = ObjectAssign({}, ES2020, {
   },
   GetIANATimeZonePreviousTransition: (epochNanoseconds, id) => {
     const lowercap = BEFORE_FIRST_DST; // 1847-01-01T00:00:00Z
-    let rightNanos = bigInt(epochNanoseconds).minus(1);
+    let rightNanos = epochNanoseconds;
```
ptomato, line 2190: Hmm, this was actually a bug that I fixed when I copied over the new polyfill, because I had already fixed it in a pull request on the old polyfill — and this change would revert it back to the old state. Maybe it somehow never got merged in the old repo? I'll investigate this.

ptomato (changes requested):
There's something I need to check out re. a bug fix in the getNextTransition/getPreviousTransition code, and I'd recommend not removing the big blocks of PlainDate/PlainTime tests, but otherwise this looks good.

lib/ecmascript.mjs
ptomato: Ugh, it's in https://github.com/tc39/proposal-temporal/pull/1370 which stalled because of the example being low priority. I had forgotten to split out the bug fix in a separate PR. Maybe we could undo these changes here for now, and then I'll work on a new PR in the proposal repo with the proper test262 tests?

test/plaintime.mjs
justingrant: Oh, OK. I just copy/pasted the test files from the old repo because there were test failures in the new repo. I can undo the deletions and just leave the test changes. This does increase the pain of syncing any changes form old=>new, but hopefully that won't be too much going fwd?

test/intl.mjs
justingrant: Yeah, I'd prefer to leave as-is for now. I'll file an issue to remove Node 12 tests at some point.

test/plaintime.mjs
ptomato: Yes, that sounds like the right thing to do.

test/intl.mjs
justingrant: Filed #6 to track

lib/ecmascript.mjs
justingrant: Your call. LMK what you'd like to do and I'll make any needed changes later this afternoon. Or you could make them. I'm easy either way.

justingrant: Oh, never mind-- I forgot we agreed in chat that you'd make these changes. Got my PRs mixed up. ;-)

ptomato: I ended up having to copy the commits in order to know exactly what should stay and what should go. It took longer, but at least I know it's correct now.

ptomato (approved):

PR #10, opened 2021-07-09
Cache Intl.DateTimeFormat to reduce memory footprint

### LINKED ISSUE #7: Reducing memory footprint by reusing Intl.DateTimeFormat instances

fer22f:
Creating Intl.DateTimeFormat instances in V8 is very slow and memory heavy. GetFormatterParts and GetCanonicalTimeZoneIdentifier are functions that are called many times when using Temporal, and they used to create new instances of Intl.DateTimeFormat for each call. In this commit, we cache them using the time zone identifier as the key.

It should be noted that doing the same to SystemTimeZone was deliberately avoided. This is due to the fact that the user's time zone may change during the execution of a program. An example is calling Temporal.now.zonedDateTimeISO() which should always output the correct time zone. This shouldn't be a problem for server-sided code which usually doesn't (or rather, shouldn't) use the time zone from the environment for calculations.

This fixes #7.

### DIFF
--- a/lib/ecmascript.mjs
+++ b/lib/ecmascript.mjs
@@ -148,6 +148,27 @@ const ES2020 = {
   Type
 };
 
+const IntlDateTimeFormatEnUsCache = new Map();
+
+function getIntlDateTimeFormatEnUsForTimeZone(timeZoneIdentifier) {
+  let instance = IntlDateTimeFormatEnUsCache.get(timeZoneIdentifier);
+  if (instance === undefined) {
+    instance = new IntlDateTimeFormat('en-us', {
+      timeZone: String(timeZoneIdentifier),
+      hour12: false,
+      era: 'short',
+      year: 'numeric',
+      month: 'numeric',
+      day: 'numeric',
+      hour: 'numeric',
+      minute: 'numeric',
+      second: 'numeric'
+    });
+    IntlDateTimeFormatEnUsCache.set(timeZoneIdentifier, instance);
+  }
+  return instance;
+}
+
 export const ES = ObjectAssign({}, ES2020, {
   ToPositiveInteger: ToPositiveInteger,
   ToFiniteInteger: (value) => {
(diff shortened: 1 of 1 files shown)


ptomato (approved):
I think @justingrant already reviewed this in the issue thread, but it looks good to me too.

Issue #4, opened 2021-07-08
How to run Test262 on this polyfill?


justingrant:
There should be a way to run Test262 tests against this polyfill, as a complement (and verification) to the existing non-Test262 tests.

Ideally, we wouldn't have to duplicate the Test262 test code from proposal-temporal. Could a script check out the latest Test262 code from proposal-temporal and run it locally against this repo?

PR #12, opened 2021-07-09
Speed up `toLocaleString` ~2.5x by reducing creation of `Intl.DateTimeFormat` instances

justingrant:
This PR speeds up `toLocaleString` and other operations that depend on the polyfilled `Intl.DateTimeFormat`. The fix was to prevent unnecessary creation of built-in `Intl.DateTimeFormat` objects, because the constructor of that built-in class is slooooooow. For more details about the underlying issue see: https://bugs.chromium.org/p/v8/issues/detail?id=6528

In local testing, speedup is about 2.5x for ZDT toLocaleString calls.  The sample code below runs in about 1800ms in the playground of the current main branch, and about 700ms in the playground of the PR branch. 

```js
(() => {
  start = Date.now();
  for (i = 0; i < 1000; i++) {
    Temporal.ZonedDateTime.from('2020-01-01[America/Los_Angeles]').toLocaleString();
  }
  total = Date.now()-start;
  console.log(`${total}ms`);
}) ()
```

### DIFF
--- a/lib/intl.mjs
+++ b/lib/intl.mjs
@@ -28,6 +28,8 @@ const ORIGINAL = Symbol('original');
 const TZ_RESOLVED = Symbol('timezone');
 const TZ_GIVEN = Symbol('timezone-id-given');
 const CAL_ID = Symbol('calendar-id');
+const LOCALE = Symbol('locale');
+const OPTIONS = Symbol('options');
 
 const descriptor = (value) => {
   return {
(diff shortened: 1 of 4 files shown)


lib/intl.mjs
```diff
@@ -41,21 +43,44 @@ const descriptor = (value) => {
...
+      obj[prop] = val;
+    }
+  } else if (typeof val === 'function') {
+    val = new IntlDateTimeFormat(obj[LOCALE], val(obj[OPTIONS]));
+    obj[prop] = val;
+  }
+  return val;
+}
+
+export function DateTimeFormat(locale, options) {
```
Ms2ger, line 64:
```suggestion
export function DateTimeFormat(locale = undefined, options = {}) {
```

Ms2ger (approved):

lib/intl.mjs
justingrant: Why is this change better? I'm not opposed, just wondering.

Ms2ger: It affects `DurationFormat.length`, which should be zero because the arguments are marked as optional in the spec

ryzokuken: s/`DurationFormat`/`DateTimeFormat`

lib/intl.mjs
```diff
@@ -41,21 +43,44 @@ const descriptor = (value) => {
 const IntlDateTimeFormat = globalThis.Intl.DateTimeFormat;
 const ObjectAssign = Object.assign;
 
-export function DateTimeFormat(locale = IntlDateTimeFormat().resolvedOptions().locale, options = {}) {
+// Construction of  built-in Intl.DateTimeFormat objects is sloooooow,
+// so we'll only create those instances when we need them.
+// See https://bugs.chromium.org/p/v8/issues/detail?id=6528
+function getPropLazy(obj, prop) {
+  let val = obj[prop];
+  if (prop === TZ_RESOLVED) {
```
ptomato, line 51: I'd probably split this part into a separate `getResolvedTimeZoneLazy()` function, so that we have two lazy property retrieval functions that have clear input and return types. That will probably help when converting the code to TypeScript.

ptomato (approved):
Thanks, this looks good!

lib/intl.mjs
justingrant: Ahhh, makes sense. What's funny is that I was waiting to send the equivalent PR to proposal-temporal until this one was approved... but the Test262 tests would have caught the problem over there. Any got ideas (see #4) for easily running Test262 against this repo? Each time I think I understand ECMAScript, I log onto this website and am reminded that I am still a padawan. 

My latest commit fixes the problem you found above, but while fixing it I found another problem: I wasn't fully protecting against the caller mutating `options` because I was only shallow-cloning it, so if someone did a wacky thing like {month: {toString: . Also I wasn't converting the `locale` param into a primitive to guard against later mutation. 

Working on a fix for that problem now.

lib/intl.mjs
justingrant: Good idea. Done.

lib/intl.mjs
```diff
@@ -41,21 +43,44 @@ const descriptor = (value) => {
...
+    obj[prop] = val;
+  }
+  return val;
+}
+
+export function DateTimeFormat(locale = undefined, options = {}) {
   if (!(this instanceof DateTimeFormat)) return new DateTimeFormat(locale, options);
+  const original = new IntlDateTimeFormat(locale, options);
+  const ro = original.resolvedOptions();
+  options = typeof options === 'undefined' ? {} : ObjectAssign({}, options);
```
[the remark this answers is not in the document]
justingrant: Yep, I made exactly this change in my latest checkin.  I also realized that shallow-cloning of `options` was not enough, because if `options` property values or `locale` were not primitives, then they could also be mutated by the caller before lazy-init happens. Take a look at my last checkin and let me know if the hacky JSON-based deep clone is sufficient.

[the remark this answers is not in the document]
justingrant: I also added tests to verify that mutating constructor inputs after the constructor returns won't break anything.

Still not sure that using a JSON round-trip is the right way to deep-clone these inputs.

lib/intl.mjs
```diff
@@ -41,21 +43,60 @@ const descriptor = (value) => {
...
+function getResolvedTimeZoneLazy(obj) {
+  let val = obj[TZ_RESOLVED];
+  if (typeof val === 'string') {
+    val = new TimeZone(val);
+    obj[TZ_RESOLVED] = val;
+  }
+  return val;
+}
+
+function deepClone(val) {
```
ptomato, line 67: Would it be possible to avoid the deep cloning altogether, and store `ro` instead of `options` for lazily creating new instances? (and use `ro.locale` as the locale?)

I ask because either deep- or shallow-cloning the object is going to be very tricky to do while matching the observable behaviour in [ToDateTimeOptions](https://tc39.es/ecma402/#sec-todatetimeoptions). But it seems like no matter whether we create a DateTimeFormat with the originally-given or the resolved options/locale, it should have the same result.

ptomato (commented):
Similarly to you, I am feeling a bit iffy about the deep-cloning, but maybe it can be avoided altogether?

lib/intl.mjs
justingrant: @ryzokuken @sffc - @ptomato's idea above sounds like a good one. Will there be any difference in behavior between `i1` and `i2` below?

```js
i1 = new DateTimeFormat(locale, options);
ro = i1.resolvedOptions();
i2 = new DateTimeFormat(ro.locale, ro);
```

justingrant: Update: using `resolvedOptions()` output to replace `options` didn't work, because the polyfill adds additional options if the user's options don't contain unit options like `year` or `hour`. 

What *does* seem to work is a small variation of this approach: clone `resolvedOptions()`and delete all properties that weren't present in the original options object.

@ptomato - take a look at my latest commit and LMK if you think this is a promising approach.

lib/intl.mjs
```diff
@@ -41,21 +43,64 @@ const descriptor = (value) => {
...
+  // that will be coerced to strings), we rely on `resolvedOptions()` to do the
+  // coercion and cloning for us. Unfortunately, we can't just use the resolved
+  // options as-is because our options-amending logic adds additional fields if
+  // the user doesn't supply any unit fields like year, month, day, hour, etc.
+  // Therefore, we limit the properties in the clone to properties that were
+  // present in the original input.
+  if (typeof options !== 'undefined') {
+    const clonedOptions = ObjectAssign({}, ro);
+    for (const prop in clonedOptions) {
+      if (typeof options[prop] === 'undefined') delete clonedOptions[prop];
```
ptomato, line 87: I would use `ES.HasOwnProperty()` here, to avoid calling getters a second time. That will still not quite be spec-compliant, since Proxy objects would be able to observe the HasProperty operation, but it's probably fine. Although that might mean we shouldn't backport this to proposal-temporal.

ptomato (approved):
I think this solution is good, though maybe one of the Intl folks could chip in?

lib/intl.mjs
justingrant: > I would use `ES.HasOwnProperty()` here, to avoid calling getters a second time. That will still not quite be spec-compliant, since Proxy objects would be able to observe the HasProperty operation, but it's probably fine. Although that might mean we shouldn't backport this to proposal-temporal.

@ptomato I looked at the current polyfill code and I think that it has the multi-getter-call problem too. Look at the `amend` function.  I don't think the current polyfill has the Proxy problem because it `ObjectAssign`s the options object before doing anything with it.

Anyway, I just pushed a new commit that does the options cloning (like the old polyfill) before reading, and also uses `ES.HasOwnProperty()` to avoid making the proxy problem any worse than it already is. I'll leave it for someone else to fix `amend` later.  Is this OK?

Also, I think this should be safe for back-porting because it doesn't make the current polyfill worse. What do you think?

LMK if you (or @ryzokuken @sffc) have any ideas how to improve this further.

Ms2ger: Yeah, I was planning to take a look at the handling in the spec polyfill at some point. I guess we can accept this as "don't pass silly arguments" here, though.

justingrant:
I think this PR is ready to merge, but I'm going to hold off on merging it for a day or so so that we can port and merge the now=>Now rename (see https://github.com/tc39/proposal-temporal/pull/1645) first.  As soon as that's merged in the old repo, I'll prepare a port of it for this repo.

PR #14, opened 2021-07-12
Fix division issues caught by tsc in RoundDuration.

12wrigja:
While migrating the source of the polyfill to TypeScript, tsc caught cases where native math operations were being used on big-integer polyfill objects as if they were native BigInts or numbers (`The left-hand side of an arithmetic operation must be of type 'any', 'number', 'bigint' or an enum type.ts(2362)`). 

This PR fixes this error by converting them back to JS numbers, but should these be turned back into JS numbers, or should these operations be done using the big-integer library?

### DIFF
--- a/lib/ecmascript.mjs
+++ b/lib/ecmascript.mjs
@@ -3898,8 +3898,8 @@ export const ES = ObjectAssign({}, ES2020, {
         oneYearDays = MathAbs(oneYearDays);
         const divisor = bigInt(oneYearDays).multiply(dayLengthNs);
         nanoseconds = divisor.multiply(years).plus(bigInt(days).multiply(dayLengthNs)).plus(nanoseconds);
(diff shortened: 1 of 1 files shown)


ptomato (changes requested):
There is currently a bit of an inconsistent state in the Temporal proposal spec text about this because of tc39/proposal-temporal#1604. However, according to the guidance in tc39/proposal-temporal#1413, internal calculations should be done as mathematical values and only converted to JS Numbers at boundaries where that is required (RoundNumberToIncrement is one such boundary.) So I believe this should do the same, e.g.
```js
const rounded = ES.RoundNumberToIncrement(nanoseconds.toJSNumber(), divisor.multiply(increment).toJSNumber(), roundingMode);
```
(Although, `total = nanoseconds.toJSNumber() / divisor.toJSNumber()` seems correct, since it is the best representation we can get in JS of the non-integer total)

ptomato (approved):

PR #21, opened 2021-07-26
Typescript

12wrigja:
This PR begins the process of migrating the Temporal polyfill's sources to TypeScript.

I'm less familiar with the impact of this change on the distributed NPM bundle / changes to package.json and package-lock.json, so a review from someone with that expertise would be greatly appreciated.

In follow-up PRs I will migrate various parts of the source to more canonical representations (e.g. Calendars should probably be based on an abstract class hierarchy).


lib/ecmascript.ts
```diff
@@ -122,7 +121,7 @@ const ALLOWED_UNITS = [
   'microsecond',
   'nanosecond'
 ];
-const SINGULAR_PLURAL_UNITS = [
+const SINGULAR_PLURAL_UNITS: Array<[string, string]> = [
```
shicks, line 124: `ReadonlyArray<readonly [string, string]>` ?

What about just adding `as const` to the end of the array literal?

lib/init.js
```diff
@@ -3,7 +3,7 @@ if (module.parent && module.parent.id === 'internal/preload') {
   globalThis.__debug__ = true;
 }
 
-import('./index.mjs')
+import('../dist/index.js')
```
shicks, line 6: Do we need Closure Compiler to be able to handle this?  Does it?

12wrigja: We don't - this file is only used to launch the polyfill Playground.

lib/calendar.ts
```diff
@@ -352,14 +351,15 @@ function resolveNonLunisolarMonth(calendarDate, overflow = undefined, monthsPerY
...
+  now: number;
+  hits = 0;
+  misses = 0;
+  constructor(cacheToClone: OneObjectCache = undefined) {
     this.now = globalThis.performance ? globalThis.performance.now() : Date.now();
-    this.hits = 0;
-    this.misses = 0;
     if (cacheToClone !== undefined) {
-      let i = cacheToClone.length;
+      let i = 0; // TODO why was this originally cacheToClone.length ?
```
ptomato, line 362: Maybe @justingrant can review this part?

tsconfig.json
```diff
@@ -0,0 +1,46 @@
...
+    ],
+    "preserveConstEnums": false,
+    "skipDefaultLibCheck": true,
+    "sourceMap": false,
+    // "strictBindCallApply": true,
+    // "strictFunctionTypes": true,
+    // "strictNullChecks": true,
+    // "strictPropertyInitialization": true,
+    "stripInternal": true,
+    "target": "es5",
```
ptomato, line 42: This is probably my unfamiliarity with TypeScript configuration, but how is it outputting ESM files for an ES5 target?

rollup.config.js
```diff
@@ -10,9 +11,10 @@ const libName = 'temporal';
 
 export default [
   {
-    input: 'lib/index.mjs',
+    input: 'lib/index.ts',
     plugins: [
-      replace({ exclude: 'node_modules/**', __debug__: !isProduction }),
+      typescript(),
+      replace({ exclude: 'node_modules/**', 'globalThis.__debug__': !isProduction, preventAssignment: true }),
```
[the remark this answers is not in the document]
12wrigja: Yes, because this replacement happens over the emitted js, not the source ts file.

I tested this by running `NODE_ENV=production npm run build`, and grepping the output file for `_repr_`, which is one of the common properties that is only set in debug builds - it is still removed from the build with this change.

lib/calendar.ts
```diff
@@ -1042,7 +1096,7 @@ const helperHebrew = ObjectAssign({}, nonIsoHelperBase, {
     } else {
       // When called without input coming from legacy Date output, simply ensure
       // that all fields are present.
-      this.validateCalendarDate(calendarDate);
+      (this as unknown as NonIsoHelperBase).validateCalendarDate(calendarDate);
```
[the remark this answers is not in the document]
12wrigja: The error is 

```
Conversion of type <omitted for brevity> to type 'NonIsoHelperBase' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
```

I believe they don't overlap because the `this` in that context is being inferred as the collection of properties only within that object literal (and doesn't include all the properties that are present due to the `nonIsoHelperBase` within the Object.assign call). This seems to be a limitation of TypeScript's type inference, but I think it could be resolved by rewriting the calendars using classes (where this sort of extension is better understood in the type system).

tsconfig.json
12wrigja: https://github.com/Microsoft/TypeScript/issues/6319
https://github.com/microsoft/TypeScript/issues/5790

These imply to me that the options here are asking tsc to emit ES5 code using the ES6 module format instead of defaulting to commonjs. It seems the module format is only loosely tied to the language output level.

I too don't know for sure if this is "right" - I was basing it off the internal tsconfig used at Google but with modifications to get the existing test suite to run (which needed ESM). If we were to change a value here, I think it would be to change target to something higher and let the rest of the rollup config handle that output.

lib/ecmascript.ts
12wrigja: Done, though this required a small adjustment to the implementation of ES.DefaultTemporalLargestUnit to compile again (though I think this change would be needed anyways as the existing implementation was not renaming-safe for Closure Compiler).

tsconfig.json
12wrigja: That being said, I did adjust the module output to explicitly be es6 instead of the unstable esnext.

justingrant:
I'll review this PR tomorrow. One immediate suggestion is to make sure to rebase on top of the most recent polyfill code from proposal-temporal because a few minor PRs have landed there recently. Also, are there any changes that you made that should be back-ported to the old MJS polyfill?

12wrigja:
I opened https://github.com/js-temporal/temporal-polyfill/pull/22 to port over the normative changes I saw in the proposal repo, though I'm not entirely sure how they will translate over to TS (unsure how you'd assign those properties to what are currently module exports). Once that is merged then I'll rebase this.

Regarding backporting: there's a fair number of changes that could be backported, but I'd prefer to wait for this to be merged before starting that. I'd probably consider backporting:
-  a fair number of let->const changes that the linter decided it wanted fixing (though why it didn't seem to trigger on the existing mjs source is confusing to me).
 - I adjusted how the `__debug__` worked to make tsc happy (it doesn't understand the global comments, and I'd prefer not to pollute global typings by re-opening the window/global namespace to include that symbol).
 - various copies of the `bigIntIfAvailable` function needed to check for `BigInt` off of `globalThis` (instead of doing so implicitly)
 - renaming the default timezone function (this seems to greatly confuse tsc, but was clearly valid code before)

Everything else are changes that are in TypeScript type space (or are adjustments to use tsc modules instead of Node mjs), so I don't think can be backported unless we moved that impl to TS source as well (which could be done if people felt it was useful?).

ptomato:
I'm afraid backporting anything that wasn't clearly a bug fix to the old impl would end up being contentious within TC39.

12wrigja:
Ah right. In that case, I don't think there are any bug-fix-like changes to backport.

lib/now.ts
```diff
@@ -1,40 +1,40 @@
...
-import { GetIntrinsic } from './intrinsicclass.mjs';
+import { ES } from './ecmascript';
+import { GetIntrinsic } from './intrinsicclass';
 
 const instant = () => {
   const Instant = GetIntrinsic('%Temporal.Instant%');
   return new Instant(ES.SystemUTCEpochNanoSeconds());
 };
-const plainDateTime = (calendarLike, temporalTimeZoneLike = timeZone()) => {
+const plainDateTime = (calendarLike, temporalTimeZoneLike = defaultTimeZone()) => {
```
justingrant, line 8: Out of curiosity, why was this renaming needed?

lib/calendar.ts
justingrant: Yeah, this looks like a bug. Probably a holdover from a previous merge-based implementation.

lib/calendar.ts
```diff
@@ -1099,7 +1148,10 @@ const helperIslamic = ObjectAssign({}, nonIsoHelperBase, {
   calendarType: 'lunar',
   inLeapYear(calendarDate, cache) {
     // In leap years, the 12th month has 30 days. In non-leap years: 29.
-    const days = this.daysInMonth({ year: calendarDate.year, month: 12, day: 1 }, cache);
+    const days = (this as unknown as NonIsoHelperBase).daysInMonth(
```
justingrant, line 1151: Instead of these casts, would it make sense to put a `this: NonIsoHelperBase` into the param declaration like you did in other files?

lib/calendar.ts
```diff
@@ -1794,7 +1852,7 @@ const nonIsoGeneralImpl = {
     cache.setObject(result);
     return result;
   },
-  yearMonthFromFields(fields, options, calendar) {
+  yearMonthFromFields(this: typeof nonIsoGeneralImpl & { helper: NonIsoHelperBase }, fields, options, calendar) {
```
justingrant, line 1855: Would it make sense to abstract this type into a global type declaration to avoid repeating `typeof nonIsoGeneralImpl & { helper: NonIsoHelperBase }` in lot of places?

justingrant (approved):
Overall looks good. Thanks! I made a bunch of minor suggestions and asked a few questions. 

The `${dt}` thing (see my comment) is concerning though... if that really breaks in all TS code, then we may want to discuss changing `valueOf` behavior to avoid throwing every time a TS user tries to console.log a template string with a Temporal instance!

rollup.config.js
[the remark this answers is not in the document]
justingrant: While you're changing this, could you also fix the rollup warning with this plugin?  Here's what I see when I run `npm build` on this branch:
```
(!) Plugin replace: @rollup/plugin-replace: 'preventAssignment' currently defaults to false. It is recommended to set this option to `true`, as the next major version will default this option to `true`.
```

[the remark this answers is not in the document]
justingrant: Er, never mind the comment above-- I was on the wrong branch.  Build works fine.

justingrant (changes requested):
I changed my mind about approval. ;-)

There's a problem with sourcemaps that will make it impossible to set breakpoints or step into this polyfill's original source using a the debugger of browser dev tools or an IDE debugger like VS Code. The problem is that the output sourcemap has incorrect relative paths for Temporal's own source files. 

If you look at /dist/index.js.map, you'll see this at the end of the `files` array: 

```
"../../lib/intrinsicclass.ts","../../lib/slots.ts","../../lib/calendar.ts","../../lib/regex.ts","../../lib/ecmascript.ts","../../lib/timezone.ts","../../lib/intl.ts","../../lib/instant.ts","../../lib/plaindate.ts","../../lib/plaindatetime.ts","../../lib/duration.ts","../../lib/plainmonthday.ts","../../lib/now.ts","../../lib/plaintime.ts","../../lib/plainyearmonth.ts","../../lib/zoneddatetime.ts","../../lib/legacydate.ts"
```

These have one too many `../` which will make the sources unfindable when installed from NPM. 

instead, the `files` list should have paths like this: `"../lib/intrinsicclass.ts"` which reflects the correct relative path between `lib` and `dist` folders. 

The root cause is this line in tsconfig.json:

```json
 "outDir": "dist/esm",
```

The rollup typescript plugin is presumably building a sourcemap assuming that output files will be living in `dist/esm` when in reality they're living in `dist`.  Simply changing tsconfig.json to "outDir": "dist" will break tests, so there's some research req'd to figure this out.

justingrant:
Here's another, more minor sourcemap isssue: I use a webpack plugin to detect differences between code stored inline in the  sourcemap vs original source files. My plugin complained about the two lines below. Any idea what's going on here?  

```
/Users/justingrant/Documents/hdev/h3/web/node_modules/tslib/tslib.es6.js doesn't match sourcemap, starting at line 20, col 48
Original Line 19:         ({ __proto__: [] } instanceof Array && function (d, b) { d.__proto__ = b; }) ||
     Map Line 19:         ({ __proto__: [] } instanceof Array && function (d, b) { d.__proto__ = b; }) ||
Original Line 20:         function (d, b) { for (var p in b) if (b.hasOwnProperty(p)) d[p] = b[p]; };
     Map Line 20:         function (d, b) { for (var p in b) if (Object.prototype.hasOwnProperty.call(b, p)) d[p] = b[p]; };
Original Line 21:     return extendStatics(d, b);
     Map Line 21:     return extendStatics(d, b);

/Users/justingrant/Documents/hdev/h3/web/node_modules/is-date-object/index.js doesn't match sourcemap, starting at line 15, col 54
Original Line 14: var dateClass = '[object Date]';
     Map Line 14: var dateClass = '[object Date]';
Original Line 15: var hasToStringTag = typeof Symbol === 'function' && !!Symbol.toStringTag;
     Map Line 15: var hasToStringTag = typeof Symbol === 'function' && typeof Symbol.toStringTag === 'symbol';
Original Line 16: 
     Map Line 16: 
```

justingrant:
A final sourcemap problem is related to the fact that our polyfill when installed in a project will be 2 levels under node_modules (first `@js-temporal` and seond `polyfill`).  So either the sourcemap should be rewritten to add an extra `../` for all references to node_modules code, or we could try rewriting the path to `~/` in place of `../node_modules`.  The `~/` is AFAICT an undocumented synonym for "the app's node_modules folder", although I don't know if it has universal support. See https://github.com/justingrant/amplify-js/blob/768ad63b2689d377eddafe3802c50a31f509e5ab/packages/aws-amplify/webpack-utils.js#L56-L60 for an example of this rewriting that I did an an AWS PR a while ago. 

The other line of investigation on this issue is why we're including code in our sourcemaps from our dependency libraries. Does this mean that rollup and/or TS includes es-abstract, big-integer, and their transitive dependencies in our bundle?  That's probably not what we want. Is there a rollup and/or TS setting to prevent that?

justingrant:
> Does this mean that rollup and/or TS includes es-abstract, big-integer, and their transitive dependencies in our bundle?

BTW, here's a relatively modern TS library and the rollup config they use: https://github.com/mongodb/js-bson/blob/master/rollup.config.js.

If you `npm i bson` and look at the sourcemaps (and likely the bundled JS files too) you'll see that bson.bundle.js.map contains `"../node_modules/buffer/index.js" (it bundles its dependencies) but bson.esm.js.map does not bundle dependencies. The latter output file comes from a rollup config with `format: 'es'` that also uses the `external` config option which presumably excludes dependencies.

Would a similar ESM config be worth adding to this polyfill?

lib/ecmascript.ts
```diff
@@ -1326,7 +1326,9 @@ export const ES = ObjectAssign({}, ES2020, {
     if (offsetOpt === 'reject') {
       const offsetStr = ES.FormatTimeZoneOffsetString(offsetNs);
       const timeZoneString = ES.IsTemporalTimeZone(timeZone) ? GetSlot(timeZone, TIMEZONE_ID) : 'time zone';
-      throw new RangeError(`Offset ${offsetStr} is invalid for ${dt} in ${timeZoneString}`);
+      // TODO: The tsc emit for this line rewrites to invoke the PlainDateTime's valueOf method, NOT
```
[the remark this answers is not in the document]
12wrigja: `throw new RangeError(`Offset ${offsetStr} is invalid for ${dt} in ${timeZoneString}`);` is transpiled by tsc to `throw new RangeError('Offset ' + offsetStr + ' is invalid for ' + dt + ' in ' + timeZoneString);`, and the '+' operator invokes `valueOf`.

This would be an issue for either language given that the problem occurs regardless of who is using tsc to compile the sources - us when we bundle the sources for packaging or others I suppose if they directly compile against our TS source.

Is having `valueOf` throwing something officially part of the Temporal spec, and thus something that the native object will do as well?

lib/ecmascript.ts
```diff
@@ -372,9 +370,9 @@ export const ES = ObjectAssign({}, ES2020, {
     const weeks = ES.ToInteger(match[4]) * sign;
     const days = ES.ToInteger(match[5]) * sign;
     const hours = ES.ToInteger(match[6]) * sign;
-    let fHours = match[7];
+    let fHours: number | string = match[7];
```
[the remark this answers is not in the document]
12wrigja: It isn't a number, but it is typed as such to "cheat" and allow that variable to be reassigned to a number below (presumably to save on memory allocation?).

Naively, I would have pulled these out in `const`s and then create a new variable to store the numeric conversion of this value, but I was attempting to avoid actual code changes (vs just type-space additions) so the runtime behaviour wouldn't change very much.

lib/ecmascript.ts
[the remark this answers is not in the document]
justingrant: > Is having `valueOf` throwing something officially part of the Temporal spec, and thus something that the native object will do as well?

Yes. https://tc39.es/proposal-temporal/#sec-temporal.plaindate.prototype.valueof. That's why I'm concerned about this.

I'll add this issue to the agenda for the next Temporal Champions meeting. We may want to keep the behavior as-is, but transpiling string interpolation to `+` might be popular enough (perhaps not only in TS, maybe in Babel too for regular JS?) that we may want to revisit.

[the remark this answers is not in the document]
justingrant: To clarify, this isn't an issue for this PR-- it's a more general issue for Temporal overall. See https://github.com/tc39/proposal-temporal/issues/1681. Feel free to resolve this conversation.

lib/ecmascript.ts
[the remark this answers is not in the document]
justingrant: Sounds good, thanks.

lib/now.ts
12wrigja:
```
Parameter 'temporalTimeZoneLike' cannot reference identifier 'timeZone' declared after it.ts(2373)
Block-scoped variable 'timeZone' used before its declaration.ts(2448)
now.ts(9, 9): 'timeZone' is declared here.
```

It seems that tsc manages to confuse the timeZone function and the constant declared within the function body, so I renamed the function.

lib/calendar.ts
12wrigja: When I tried this I got an error: `Type alias 'NonIsoGeneralImplThis' circularly references itself.ts(2456)`.

Example:
```
type NonIsoGeneralImplThis = typeof nonIsoGeneralImpl & { helper: NonIsoHelperBase };

... 

const nonIsoGeneralImpl = {
  dateFromFields(this: NonIsoGeneralImplThis, fields, options, calendar) {
    const overflow = ES.ToTemporalOverflow(options);
```

lib/ecmascript.ts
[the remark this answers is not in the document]
12wrigja: Thanks for opening that issue. I've added a reference to the TypeScript repo issue so if someone stumbles across this fix in the future and finds out it's been fixed upstream we can revert this fix.

lib/ecmascript.ts
```diff
@@ -1913,7 +1915,7 @@ export const ES = ObjectAssign({}, ES2020, {
     if (timeZone !== undefined) timeZoneString = ES.BuiltinTimeZoneGetOffsetStringFor(outputTimeZone, instant);
     return `${year}-${month}-${day}T${hour}:${minute}${seconds}${timeZoneString}`;
   },
-  TemporalDurationToString: (duration, precision = 'auto', options = undefined) => {
+  TemporalDurationToString: (duration, precision: 'auto' | number = 'auto', options = undefined) => {
```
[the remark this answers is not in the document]
12wrigja: Done.

lib/calendar.ts
12wrigja: Done - replaced all the `as unknown as` with the equivalent `this` type.

lib/calendar.ts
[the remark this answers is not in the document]
12wrigja: I swapped this over to specifying the `this` type as part of the function's parameters to avoid the dangerous casts. Resolving.

12wrigja:
I've pushed another commit that I believe addresses your comments - it's heavily based on the commit you sent me with some small adjustments. In particular, we don't need to have two tsconfig files as the rollup TypeScript plugin can use the package tsconfig as a base and seems to adjust various settings automatically to work with the rest of the build tools.

I'll wait until #25 is merged before rebasing and squashing / cleaning up the commits.

lib/calendar.ts
12wrigja: In that case it should probably get backported right? If I understand the code right, in a hypothetical scenario where one was copying a "full cache" (has MAX_CACHE_ENTRIES things in it), there would be nothing in the new cache?

justingrant: The consequence of this is purely performance, not functional, so unless we're seeing aberrant perf caused by cache exhaustion (AFAIK we're not) then there's no need to back-port to the old polyfill that's just there for running tests. AFAIK there are no tests that fill up the cache, BTW.

tsconfig.json
```diff
@@ -33,14 +32,13 @@
...
     "skipDefaultLibCheck": true,
-    "sourceMap": false,
     // "strictBindCallApply": true,
     // "strictFunctionTypes": true,
     // "strictNullChecks": true,
     // "strictPropertyInitialization": true,
     "stripInternal": true,
     "target": "es5",
-    "outDir": "dist/esm",
+    "outDir": "ts-dist/",
```
justingrant, line 41: Changes look great. Sourcemaps and tests both work. Minor suggestion: the folder name `ts-dist` is too close to "dist" for my taste, given that the former is only for internal use while the latter is published to npm. Possible to pick another name with "test" or "internal" in the name and "dist" not in it?

12wrigja: I've moved this to ./tsc-out/ so it's very clear what is putting files there.

lib/now.ts
justingrant: In this case, I'd suggest renaming the local variables `tz` instead of renaming the `timeZone` function. This would be less change vs. JS, and also IMHO it'd be more clear that the default params are calling the implementation of `Now.timeZone` as opposed to some other function.

12wrigja: Done - renamed as you suggested.

lib/calendar.ts
justingrant: Ahhh, OK. Good thing for a future black-belt TS contributor to figure out. Not req'd now.

BTW, I apologize for the TS-unfriendly structure of this code. I wrote it this way to simplify adapting the code from the previous non-ISO-calendar implementation which used a similar class-like-but-not-really structure for these singletons. If I were starting all over again I'd probably do it differently. Sorry!

justingrant (approved):
Great work here!

justingrant:
@12wrigja - Actually, one more thing: I did some research about bundling dependencies, and I think CJS should *not* bundle its dependencies. I think that this will also solve any sourcemap paths issues for anyone using non-ESM bundles in node.

The change would be simple: 

```js
export default [
  {
    input,
    external,
    output: [
      outputEntry(pkg.module, 'es'),
      outputEntry(pkg.main, 'cjs')
    ],
    plugins
  },
  {
    input,
    output: [
      outputEntry(pkg.browser, 'umd'),
      outputEntry(pkg.module.replace('.esm', '.browser.esm'), 'es')
    ],
    plugins
  }
];
```

What do you think?

12wrigja:
@justingrant Done. Pity about the linter formatting though.

12wrigja:
I've rebased this on top of main and squashed in the adjustments to the rollup config. Is there anything left to resolve, or can we merge this now?

PR #23, opened 2021-07-29
Change let->const to satisfy lint checks for TS sources.

12wrigja:
The existing eslint config doesn't seem to mind the lets currently, but does once the files are migrated. Splitting this into a separate PR makes the TypeScript source migration PR easier to review.

### DIFF
--- a/lib/intl.mjs
+++ b/lib/intl.mjs
@@ -107,8 +107,8 @@ export function DateTimeFormat(locale = undefined, options = undefined) {
   this[INST] = instantAmend;
 }
 
-DateTimeFormat.supportedLocalesOf = function (...args) {
-  return IntlDateTimeFormat.supportedLocalesOf(...args);
+DateTimeFormat.supportedLocalesOf = function (locales, options) {
+  return IntlDateTimeFormat.supportedLocalesOf(locales, options);
 };
 
 const properties = {
@@ -194,7 +194,7 @@ function formatRangeToParts(a, b) {
 
 function amend(options = {}, amended = {}) {
   options = ObjectAssign({}, options);
-  for (let opt of [
+  for (const opt of [
     'year',
     'month',
     'day',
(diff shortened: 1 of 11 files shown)


lib/intl.mjs
```diff
@@ -107,8 +107,8 @@ export function DateTimeFormat(locale = undefined, options = undefined) {
   this[INST] = instantAmend;
 }
 
-DateTimeFormat.supportedLocalesOf = function (...args) {
-  return IntlDateTimeFormat.supportedLocalesOf(...args);
+DateTimeFormat.supportedLocalesOf = function (locales, options) {
```
ptomato, line 110: This change seems unrelated, but makes sense to me, because it makes `Intl.DateTimeFormat.supportedLocalesOf.length` conformant.

ptomato (approved):
Thanks! This will indeed make the other PR easier to review. I'm not sure why the eslint config didn't pick these up before migrating to TypeScript!

lib/intl.mjs
12wrigja: I changed this because tsc doesn't like using the argument spread to invoke supportedLocalesOf, but making it length conformant is handy :)

justingrant: @ptomato - does that mean that the proposal-temporal polyfill is *not* conformant here, and if so should this change be ported over there?

ptomato: Yes, but:
1. Intl.DateTimeFormat.supportedLocalesOf is not part of the Temporal proposal, so it won't be tested and probably doesn't matter. For this polyfill, it does matter, because people might replace their Intl object in production code.
2. I also realized it is still not conformant. This makes it have length 2 rather than 0, but it is actually supposed to have length 1. The signature is Intl.DateTimeFormat.supportedLocalesOf ( locales [ , options ] )

justingrant: @ptomato - so the right signature would be something like `function (locales, options = {})` ?

@12wrigja - I'd suggest splitting this change out into a new PR, and getting the conformance correct.

justingrant: Oops, just noticed this is already merged! Now we definitely need a separate PR to fix the conformance. ;-)

justingrant:
Great PR. Thanks @12wrigja !

PR #30, opened 2021-08-14
Add '.js' suffix to ecmascript imports

chrisveness:
File extensions are mandatory in ESM imports – q.v.
https://nodejs.org/api/esm.html#esm_mandatory_file_extensions.


ptomato (approved):
Thanks!

PR #31, opened 2021-08-15
Sync package-lock.json to package.json version

justingrant:
#27 didn't update the version in package-lock.json. I don't know what (if anything) this will break, but I assume they should be synced.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@js-temporal/polyfill",
-  "version": "0.1.0",
+  "version": "0.2.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@js-temporal/polyfill",
-      "version": "0.1.0",
+      "version": "0.2.0",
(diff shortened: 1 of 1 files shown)


ptomato (approved):
Thanks! I also don't know what this affects, but I agree, this should be probably synced.

12wrigja:
Is it possible to add a test / check for this as part of merging a PR?

ptomato:
This seems promising: https://stackoverflow.com/q/57998744/172999

justingrant:
I've noticed that a some OSS libraries simply don't ship a package-lock.json at all. Not sure what the pros/cons of doing this are.

ptomato:
The NPM utility itself encourages you to ship package-lock.json. It prevents you from getting different results locally / on CI / on users' machines, in the case where a newer version of a dependency has been released in between the last time you did `npm install` and the time when a user does `npm install`.

Personally I really hate checking in long autogenerated files that produce giant pull request diffs, but to me the above benefit weighs stronger than the dislike :smile:

PR #36, opened 2021-08-24
Fix non-ISO months arithmetic

justingrant:
Port of https://github.com/tc39/proposal-temporal/pull/1761

### DIFF
--- a/lib/calendar.ts
+++ b/lib/calendar.ts
@@ -829,10 +829,29 @@ const nonIsoHelperBase: NonIsoHelperBase = {
   addMonthsCalendar(calendarDate, months, overflow, cache) {
     const { day } = calendarDate;
     for (let i = 0, absMonths = MathAbs(months); i < absMonths; i++) {
-      const days = months < 0 ? -this.daysInPreviousMonth(calendarDate, cache) : this.daysInMonth(calendarDate, cache);
+      const { month } = calendarDate;
+      const oldCalendarDate = calendarDate;
+      const days =
+        months < 0
+          ? -Math.max(day, this.daysInPreviousMonth(calendarDate, cache))
+          : this.daysInMonth(calendarDate, cache);
       const isoDate = this.calendarToIsoDate(calendarDate, 'constrain', cache);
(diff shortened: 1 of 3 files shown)


ptomato:
Let's keep the review over on proposal-temporal for now, but I wanted to add that it might be worth restructuring this algorithm to better match [AddISODate](https://tc39.es/proposal-temporal/#sec-temporal-addisodate) which would involve generalizing BalanceISOYearMonth, BalanceISODate, and RegulateISODate to other calendars. Of course, it may not be worth it, if we change this polyfill to use, say, an implementation not based on Intl.DateTimeFormat...

justingrant:
> Of course, it may not be worth it, if we change this polyfill to use, say, an implementation not based on Intl.DateTimeFormat...

This would be my preferred solution. Most of the code in the current implementation is working around not having a good API to translate Calendar=>ISO dates (including eras & month codes) nor to get calendar-specific info like days-in-month or months-in-year. If we had that API, probably 80% of the non-ISO calendar code in the polyfill would go away. 

@sffc - once ICU4X is ready, would it be suitable for use in a polyfill? Is there some sort of Rust=>JS or Rust=>WASM=>JS transpilation possible?  If yes, could we ship parts of it we need (noted above) and not the whole thing?

> Let's keep the review over on proposal-temporal for now, but I wanted to add that it might be worth restructuring this algorithm to better match [AddISODate](https://tc39.es/proposal-temporal/#sec-temporal-addisodate) which would involve generalizing BalanceISOYearMonth, BalanceISODate, and RegulateISODate to other calendars

I won't have time to do this, but seems like a reasonable idea and I'd support it if someone wants to do the work.  Maybe a good time to do this could be if/when we get a better underlying API as noted above?

ptomato (approved):
I reviewed the equivalent code in proposal-temporal, assuming this is the same!

As for how to handle it in the future — I'm hoping that we won't need to continue doing that for too much longer. I'll pull in all the normative changes from the upcoming TC39 plenary all at once when they are approved, and after that we'll just need to watch for normative changes in the proposal, just like all the other implementors.

PR #45, opened 2021-09-09
Fix crash setting `day` outside current JPN era

justingrant:
Port of https://github.com/tc39/proposal-temporal/pull/1807

### DIFF
--- a/lib/calendar.ts
+++ b/lib/calendar.ts
@@ -1953,6 +1953,8 @@ const nonIsoGeneralImpl = {
   mergeFields(fields, additionalFields) {
     const fieldsCopy = { ...fields };
     const additionalFieldsCopy = { ...additionalFields };
+    // era and eraYear are intentionally unused
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
     const { month, monthCode, year, era, eraYear, ...original } = fieldsCopy;
     const {
       month: newMonth,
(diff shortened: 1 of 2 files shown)


lib/calendar.ts
```diff
@@ -1953,6 +1953,8 @@ const nonIsoGeneralImpl = {
   mergeFields(fields, additionalFields) {
     const fieldsCopy = { ...fields };
     const additionalFieldsCopy = { ...additionalFields };
+    // era and eraYear are intentionally unused
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
     const { month, monthCode, year, era, eraYear, ...original } = fieldsCopy;
```
12wrigja, line 1958: If these are really unused, can we just delete them from the destructuring?

12wrigja (approved):

lib/calendar.ts
justingrant: They're there to ensure that `...original` lacks those fields. I could have used `delete` instead (or assumed that `fields` object only contains a known set of fields and remove `...original`), but using unused variables seemed like the smallest change to existing code.  Also happy to do it a different way. What do you think?

12wrigja: I would suggest using delete and maybe moving some of the comments from below to explain why those fields in particular are being removed from `original`.

justingrant: @12wrigja Whoops, https://github.com/tc39/proposal-temporal/pull/1807 was already merged before I had a chance to make these updates in both repos. How'd you feel about merging this as-is to avoid diverging from proposal-temporal?

Then, if you think it's worth it, I'd be happy to do a refactoring PR to both repos to make the changes you recommended above. Or just leave as-is. Your call, I'm flexible either way.

ptomato: I'm fine with this either way as well.

ptomato (approved):

lib/calendar.ts
justingrant: OK I'll just merge this now and we can always follow up with a cleanup PR on both sides if there's interest.

PR #46, opened 2021-09-10
Add VS Code configs for IDE debugging of tests

justingrant:
This PR adds a launch.json debug configuration file that VS Code users can use for one-click IDE debugging of either Demitasse or (coming soon once #41 is merged) Test262 tests.  

Demitasse test debugging "just works".  You should be able to set breakpoints inside Demitasse tests and the debugger will break there.

Test262 debugging has more quirks.  Notably, breakpoints don't work inside the Test262 test runner because there's no sourcemap pointing from the executing code back to the original test sources. This may not be worth solving anyways, because I suspect that generating sourcemaps for every test file would take a loooooooong time!

What *does* work is to place a `debugger;` statement inside a Test262 test that you want to debug. Execution will pause at the `debugger` statement and then you can inspect variables, step in/over, and set breakpoints on later code. Just remember to remove the `debugger` statement before committing tests! 😄  

There are two Test262 debug configs:
1.  `Test262 (current file)` is the easiest. It just debugs the test file currently open in the editor.
2. `Test262 (subset in launch.json)` debugs a subset of tests hardcoded to a glob pattern in launch.json. This is useful for repeatedly debugging the same set of multiple tests. 

In addition to #41, tolerable Test262 debugging also depends on https://github.com/tc39/proposal-temporal/pull/1812 which contains several fixes to the debug experience.  It's not really usable without those changes.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,50 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "name": "Test262 (edit subset in launch.json)",
+      "request": "launch",
+      "type": "node",
+      "runtimeExecutable": "npm",
+      "runtimeArgs": ["run", "test262"],
+      "console": "integratedTerminal",
+      "internalConsoleOptions": "neverOpen",
+      "autoAttachChildProcesses": true,
+      "env": {
+        // Replace the glob pattern below with tests you want to debug.
+        // Comment out to run all tests, but will be very slow!
+        "TESTS": "PlainYearMonth/prototype/subtract/*.js",
+        // Extends timeouts to 1 hour (from 10 seconds default) so you can
+        // debug a test stopped at a breakpoint before the runner kills it.
+        "TIMEOUT": "3600000"
+      }
+    },
+    {
+      "name": "Test262 (current file)",
+      "request": "launch",
+      "type": "node",
+      "runtimeExecutable": "npm",
+      "runtimeArgs": ["run", "test262"],
+      "console": "integratedTerminal",
+      "internalConsoleOptions": "neverOpen",
+      "autoAttachChildProcesses": true,
+      "env": {
+        // Only test the currently-opened file in the IDE
+        "TESTS": "${file}",
+        // Extends timeouts to 1 hour (from 10 seconds default) so you can
+        // debug a test stopped at a breakpoint before the runner kills it.
+        "TIMEOUT": "3600000"
+      }
+    },
+    {
+      "name": "Demitasse tests",
+      "request": "launch",
+      "type": "node",
+      "runtimeExecutable": "npm",
+      "runtimeArgs": ["test"],
+      "console": "integratedTerminal",
+      "internalConsoleOptions": "neverOpen",
+      "autoAttachChildProcesses": true,
+    },
+ ]
+}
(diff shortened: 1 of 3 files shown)


12wrigja:
> Just remember to remove the debugger statement before committing tests! 😄

This sounds like a useful check to put in place somewhere.

.vscode/launch.json
```diff
@@ -0,0 +1,56 @@
...
+    {
+      "name": "Test262 (subset in launch.json)",
+      "request": "launch",
+      "type": "node",
+      "runtimeExecutable": "npm",
+      "runtimeArgs": ["run", "test262"],
+      "protocol": "inspector",
+      "console": "integratedTerminal",
+      "internalConsoleOptions": "neverOpen",
+      "disableOptimisticBPs": false,
```
12wrigja, line 13: What does this do?

.vscode/launch.json
```diff
@@ -0,0 +1,56 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "name": "Test262 (subset in launch.json)",
```
12wrigja, line 5: Maybe include the word "Change" in here somewhere?

12wrigja (approved):

justingrant:
> > Just remember to remove the debugger statement before committing tests! 😄
> 
> This sounds like a useful check to put in place somewhere.

There's already an eslint rule warning about not using `debugger`. Do eslint errors cause cause CI failures?  If not, then I agree this check would be good to add, but probably out of scope to this PR.

.vscode/launch.json
justingrant: Cool, we have an [answer from VSCode](https://github.com/microsoft/vscode-js-debug/issues/1102#issuecomment-917106779): `protocol` and  `disableOptimisticBPs` are not used anymore. I'll remove them. The others look good.

12wrigja:
If it already exists as an enabled lint rule, then it should break CI (as we do check for the lint returning successfully). Definitely out of scope for this PR - i'll test it shortly and send a PR to turn on the rule if it isn't enabled.

Is this ready to merge?

justingrant:
> Is this ready to merge?

Yep! (Assuming you're OK with my suggestion in https://github.com/js-temporal/temporal-polyfill/pull/41/files#r706482347)

12wrigja:
That seems fine - let's go with the environment variable. Thanks!

12wrigja:
R.e. the debugger check: this repo's linter is already configured to catch it:

```
./temporal-polyfill/test/tc39/polyfill/test/Calendar/prototype/dateAdd/balance-smaller-units.js
  15:1  error  Unexpected 'debugger' statement  no-debugger

./temporal-polyfill/test/tc39/polyfill/test/PlainDate/prototype/subtract/argument-string-negative-fractional-units.js
  13:1  error  Unexpected 'debugger' statement  no-debugger
  ```
  
  Also, this was very helpful for debugging the various test failures and narrowing down what was missing. Thank you so so much!

justingrant:
> Also, this was very helpful for debugging the various test failures and narrowing down what was missing.

So glad to hear!  I've been annoyed at the difficulty of debugging Temporal tests (esp. Test262) for a long time, so was good to have an excuse to make it easier.

PR #50, opened 2021-09-17
Revert "Copy options object for PlainYearMonth.{add,subtract} and Int…

12wrigja:
…erpretTemporalDateTimeFields to prevent user-modified objects from interfering with later operations."

This reverts commit bafa1bdf2dbfc28513d7e39b0c0d1c3d075d9db5 in order to get test262 to pass against this repo.

Once the corresponding change lands in the proposal repo(https://github.com/tc39/proposal-temporal/pull/1748), we can port that back over.


ptomato:
Oh, I didn't realize we were blocking on that, how about I fix it up and land it over in the other repo right now?

12wrigja:
Either works :) I just want to unblock landing the 262 test suite so we can be running it as part of PR CI.

ptomato:
It's merged on the other side now.

12wrigja:
Thanks! We don't need this anymore then.

PR #52, opened 2021-09-20
Remove usage of es-abstract Call, GetMethod, and HasOwnProperty.

12wrigja:


justingrant:
What's the reason to make these changes?  I don't disagree, just wanted to understand the context better.

12wrigja:
Several reasons:
 - the pattern of retrieving properties with string keys doesn't work with Closure Compiler's property renaming optimizations.
 - es-abstract is a very complex dependency to get working within Google's build system (which relies on being able to build almost entirely from source). For a dependency that could easily be dropped and replaced with native JS code inline it doesn't seem to pull its weight here (though definitely does if you are writing a polyfill to check the spec).

lib/ecmascript.ts
```diff
@@ -1532,107 +1527,94 @@ export const ES = ObjectAssign({}, ES2020, {
...
     if (ES.Type(result) !== 'Object') throw new TypeError('bad return from calendar.mergeFields()');
     return result;
   },
   CalendarDateAdd: (calendar, date, duration, options, dateAdd) => {
     if (dateAdd === undefined) {
-      dateAdd = ES.GetMethod(calendar, 'dateAdd');
+      dateAdd = calendar.dateAdd;
     }
-    const result = ES.Call(dateAdd, calendar, [date, duration, options]);
+    const result = calendar.dateAdd(date, duration, options);
```
ptomato, line 1541: This is not quite correct, because if `dateAdd` is undefined then `calendar.dateAdd` will be accessed twice, potentially calling a getter twice. We have to do the equivalent of `dateAdd.call(calendar, date, duration, options)` but also being robust in the case that user code does `delete Function.prototype.call` which probably means saving `Reflect.apply` in a variable at the top of the file and using that to call the function with `calendar` as the this-value.

lib/ecmascript.ts
```diff
@@ -1532,107 +1527,94 @@ export const ES = ObjectAssign({}, ES2020, {
...
     if (!ES.IsTemporalDate(result)) throw new TypeError('invalid result');
     return result;
   },
   CalendarDateUntil: (calendar, date, otherDate, options, dateUntil) => {
     if (dateUntil === undefined) {
-      dateUntil = ES.GetMethod(calendar, 'dateUntil');
+      dateUntil = calendar.dateUntil;
     }
-    const result = ES.Call(dateUntil, calendar, [date, otherDate, options]);
+    const result = calendar.dateUntil(date, otherDate, options);
```
ptomato, line 1549: (same)

lib/calendar.ts
```diff
@@ -1565,7 +1565,8 @@ const makeHelperGregorian = (id, originalEras) => {
       if (month === undefined) calendarDate = { ...calendarDate, month: monthCodeNumberPart(monthCode) };
       this.validateCalendarDate(calendarDate);
       calendarDate = this.completeEraYear(calendarDate);
-      calendarDate = ES.Call(nonIsoHelperBase.adjustCalendarDate, this, [calendarDate, cache, overflow]);
+      // TODO this can become `super` later.
+      calendarDate = nonIsoHelperBase.adjustCalendarDate.apply(this, [calendarDate, cache, overflow]);
```
ptomato, line 1569: Same as below, save `Reflect.apply` in a global variable and use it here

lib/ecmascript.ts
```diff
@@ -1742,11 +1721,11 @@ export const ES = ObjectAssign({}, ES2020, {
     );
   },
   GetOffsetNanosecondsFor: (timeZone, instant) => {
-    let getOffsetNanosecondsFor = ES.GetMethod(timeZone, 'getOffsetNanosecondsFor');
+    let getOffsetNanosecondsFor = timeZone.getOffsetNanosecondsFor;
     if (getOffsetNanosecondsFor === undefined) {
       getOffsetNanosecondsFor = GetIntrinsic('%Temporal.TimeZone.prototype.getOffsetNanosecondsFor%');
     }
-    const offsetNs = ES.Call(getOffsetNanosecondsFor, timeZone, [instant]);
+    const offsetNs = getOffsetNanosecondsFor.apply(timeZone, [instant]);
```
ptomato, line 1728: Same, use a saved `Reflect.apply` here

lib/intl.ts
```diff
@@ -90,7 +90,9 @@ export function DateTimeFormat(
   if (hasOptions) {
     const clonedResolved = ObjectAssign({}, ro);
     for (const prop in clonedResolved) {
-      if (!ES.HasOwnProperty(options, prop)) delete clonedResolved[prop];
+      if (!Object.prototype.hasOwnProperty.call(options, prop)) {
```
ptomato, line 93: Should save `Object.prototype.hasOwnProperty` at the top of the file in order to be robust against user code deleting it, and use a saved `Reflect.apply` to call it.

ptomato (approved):
I think this is all OK except for some inadvertent-looking semantic changes that I've marked.

Note that since `c = GetMethod(a, b)` throws a TypeError immediately if the gotten value is not undefined and not callable, when replacing it with `c = a.b`, if there is any chance that something else could throw an exception before calling `c()`, then we'd need to insert an extra check that `c` is callable. But I looked specifically for that and I don't think there are any instances of it.

My opinion is that it's fine to work towards inlining the es-abstract family of dependencies if that makes it easier to use. That will probably help reduce the bundle size, and I suspect es-abstract also contains some workarounds that we don't need because the environments that require them are too obsolete for this polyfill.

lib/ecmascript.ts
12wrigja: Gotcha. I think the same applies to mergeFields?

PR #53, opened 2021-09-21
Add @@toStringTag to TS types

justingrant:
We implement @@toStringTag, but it's not currently in the TS types. This PR fixes that.  There are some problems with the initial implementation that I'll discuss in follow-up comments.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -520,6 +520,7 @@ export namespace Temporal {
     toLocaleString(locales?: string | string[], options?: Intl.DateTimeFormatOptions): string;
     toJSON(): string;
     toString(options?: ToStringPrecisionOptions): string;
+    readonly [Symbol.toStringTag]: 'Temporal.Duration';
   }
 
   /**
@@ -616,6 +617,7 @@ export namespace Temporal {
     toJSON(): string;
     toString(options?: InstantToStringOptions): string;
     valueOf(): never;
+    readonly [Symbol.toStringTag]: 'Temporal.Instant';
   }
 
   export interface CalendarProtocol {
@@ -779,6 +781,7 @@ export namespace Temporal {
     fields(fields: Iterable<string>): Iterable<string>;
     mergeFields(fields: Record<string, unknown>, additionalFields: Record<string, unknown>): Record<string, unknown>;
     toString(): string;
+    readonly [Symbol.toStringTag]: 'Temporal.Calendar';
   }
 
   export type PlainDateLike = {
@@ -875,6 +878,7 @@ export namespace Temporal {
     toJSON(): string;
     toString(options?: ShowCalendarOption): string;
     valueOf(): never;
+    readonly [Symbol.toStringTag]: 'Temporal.PlainDate';
   }
 
   export type PlainDateTimeLike = {
@@ -1044,6 +1048,7 @@ export namespace Temporal {
     toJSON(): string;
     toString(options?: CalendarTypeToStringOptions): string;
     valueOf(): never;
+    readonly [Symbol.toStringTag]: 'Temporal.PlainDateTime';
   }
 
   export type PlainMonthDayLike = {
@@ -1078,6 +1083,7 @@ export namespace Temporal {
     toJSON(): string;
     toString(options?: ShowCalendarOption): string;
     valueOf(): never;
+    readonly [Symbol.toStringTag]: 'Temporal.PlainMonthDay';
   }
 
   // Temporal.PlainTime's `calendar` field is a Temporal.Calendar, not a
@@ -1215,6 +1221,7 @@ export namespace Temporal {
     toJSON(): string;
     toString(options?: ToStringPrecisionOptions): string;
     valueOf(): never;
+    readonly [Symbol.toStringTag]: 'Temporal.PlainTime';
   }
 
   /**
@@ -1272,6 +1279,7 @@ export namespace Temporal {
     getPossibleInstantsFor(dateTime: Temporal.PlainDateTime | PlainDateTimeLike | string): Temporal.Instant[];
     toString(): string;
     toJSON(): string;
+    readonly [Symbol.toStringTag]: 'Temporal.TimeZone';
   }
 
   export type PlainYearMonthLike = {
@@ -1329,6 +1337,7 @@ export namespace Temporal {
     toJSON(): string;
     toString(options?: ShowCalendarOption): string;
     valueOf(): never;
+    readonly [Symbol.toStringTag]: 'Temporal.PlainYearMonth';
   }
 
   export type ZonedDateTimeLike = {
@@ -1492,6 +1501,7 @@ export namespace Temporal {
     toJSON(): string;
     toString(options?: ZonedDateTimeToStringOptions): string;
     valueOf(): never;
+    readonly [Symbol.toStringTag]: 'Temporal.ZonedDateTime';
   }
 
   /**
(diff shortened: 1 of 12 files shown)


justingrant:
Currently this PR breaks the build because our implementation of @@toStringTag is implemented generically inside intrinsicclass.ts. So from TS's point of view, no Temporal types implement this property and the rollup TS plugin complains with errors like this:

```
lib/index.ts → ./dist/index.esm.js, ./dist/index.cjs...
[!] (plugin typescript) Error: @rollup/plugin-typescript TS2720: Class 'import("/Users/justingrant/Documents/hdev/temporal-polyfill/lib/calendar").Calendar' incorrectly implements class 'import("/Users/justingrant/Documents/hdev/temporal-polyfill/index").Temporal.Calendar'. Did you mean to extend 'import("/Users/justingrant/Documents/hdev/temporal-polyfill/index").Temporal.Calendar' and inherit its members as a subclass?
  Property '[Symbol.toStringTag]' is missing in type 'import("/Users/justingrant/Documents/hdev/temporal-polyfill/lib/calendar").Calendar' but required in type 'import("/Users/justingrant/Documents/hdev/temporal-polyfill/index").Temporal.Calendar'.
lib/calendar.ts (37:14)

37 export class Calendar implements Temporal.Calendar {
                ~~~~~~~~

  index.d.ts:784:14
    784     readonly [Symbol.toStringTag]: 'Temporal.Calendar';
                     ~~~~~~~~~~~~~~~~~~~~
    '[Symbol.toStringTag]' is declared here.
```

@12wrigja, any ideas for how to address this?

12wrigja:
I was going to look into the IntrinsicClass setup this week. I was hoping to make use of the TypeScript assertion function syntax, but I don't think this works given that the call happens after the class is defined.

Potentially one solution would be to create an abstract base class that accepts the string to use as the name and have every intrinsic class call the super constructor.

12wrigja:
https://github.com/12wrigja/temporal-polyfill/tree/to-string-tag-ts

I thought maybe a pattern like this might work, but apparently the TS emit strips out all the assignments to `[Symbol.toStringTag]`. Still trying to figure out why that happened.

I couldn't seem to get anything working that passed type-checking - so far as I understand it the type analysis in TS doesn't go "backwards", so having functions that adjust a class to add more properties can't adjust the type of that reference until after the call (e.g. using the assertion function syntax to indicate that something passed in has changed types).

justingrant:
Could something like below work?  I've tried this on my fork and just force-pushed it here so you can take a look. It passes TS. There's still a missing piece that `constructor.name` is wrong, but MakeIntrinsicClass could fix that. And it's possible that other Test262 tests will fail even after that's fixed, haven't tried yet. 

```ts
// eslint-disable-next-line @typescript-eslint/ban-types
export type WithoutToStringTag<T extends object> = Omit<T, typeof Symbol.toStringTag>;

class DurationImpl implements WithoutToStringTag<Temporal.Duration> {
. . . // (don't implement toStringTag here)
} 

MakeIntrinsicClass(DurationImpl, 'Temporal.Duration');

export const Duration = DurationImpl as unknown as typeof Temporal.Duration;
```

justingrant:
> There's still a missing piece that `constructor.name` is wrong, but MakeIntrinsicClass could fix that.

Arrgh, turns out that changing the `name` property of a class is not supported. See https://github.com/microsoft/TypeScript/pull/12065. So this approach won't work. But (good news) I did manage to get your idea working: implementing @@toStringTag inside the class itself. See the code I just pushed.

> I thought maybe a pattern like this might work, but apparently the TS emit strips out all the assignments to `[Symbol.toStringTag]`. Still trying to figure out why that happened.

The trick turned out to be one character: `!`.  From https://www.typescriptlang.org/docs/handbook/2/classes.html: 

> If you intend to definitely initialize a field through means other than the constructor (for example, maybe an external library is filling in part of your class for you), you can use the definite assignment assertion operator, `!`:

Example: 
```ts
class X {
    constructor() {
        console.log('foo');
    }
    readonly [Symbol.toStringTag]!: 'hello';
}
```

This still doesn't actually set the property (`MakeIntrinsicClass` is still needed for that) but it fixes all the TS errors during the build.

If we want to remove the property-setting from `MakeIntrinsicClass`, that's a harder problem. I tried a few options and the closest I got was to switch from a static property to a getter function in the class body.

12wrigja:
> Arrgh, turns out that changing the name property of a class is not supported. See microsoft/TypeScript#12065. So this approach won't work.

I'm guessing this is because a 262 test is expecting that the function name for `Duration` is `Duration`, and not `DurationImpl` like in your example? Shame.

I tried several variants on this pattern to try and reduce the number of duplications of this data, but I don't think there's any that I like. The cleanest is probably using a getter like you mentioned, but it's very easy to get wrong and seems wasteful to allocate an entire getter for something that is static.

This LG. I presume the implementation issues you mentioned were to do with `Temporal.Now` not having a `@@toStringTag`? Was there anything else you wanted to mention?

12wrigja (approved):

justingrant:
Thanks @12wrigja for pointing me towards the a solution! 

Per your proposal about tracking changes in both repos, should we wait to merge this until it's been merged in proposal-temporal?  (The change in that repo will only be the new .d.ts, not any of the other files.)  Or best to merge this one now and I can open a new PR to port it over to proposal-temporal?

> I'm guessing this is because a 262 test is expecting that the function name for `Duration` is `Duration`, and not `DurationImpl` like in your example? Shame.

Yep. But the `name` property is observable native behavior so the polyfill should match.

> I presume the implementation issues you mentioned were to do with `Temporal.Now` not having a `@@toStringTag`?

That was the last piece I figured out in this PR. Because Now was previously declared as a TS `namespace`, there was no way to declare its @@toStringTag because you can't export symbol values from a TS namespace, per https://github.com/microsoft/TypeScript/issues/36813.

So I switched Now to be declared as a `const` instead, which let me add @@toStringTag. AFAIK this will not change anything about how clients use Now, and I tried one of my own apps and it worked with no complaints.

> I tried several variants on this pattern to try and reduce the number of duplications of this data, but I don't think there's any that I like. The cleanest is probably using a getter like you mentioned, but it's very easy to get wrong and seems wasteful to allocate an entire getter for something that is static.

Yep, agreed. What's in this PR seems to be the path of least resistance, at least for now. 

I was able to get the behavior we want (where TS would emit a class field using the newly-adopted ES class fields syntax), but getting that output it requires changing TS config which I was reluctant to do here in this PR without understanding the implications of doing so. The changes required seem to be these: 
```json
useDefineForClassFields: true,
target: 'ESNext'
```

Here's an example if you want to play around with the various options.
https://www.typescriptlang.org/play?#code/MYGwhgzhAEAa0G8CwAoa7rAPYDsIBcAnAV2Hy0IAoBKRVDBzXCLEAUwDoQsBzSgcgBmWLP2oBuehgC+U9AG0AygE8AtgCNWHcoqIBLHDwAqYHgF1oAXmj8AFmxDd+ktBnUAuaDmIa2hK9AALABMLgwEYPh6wJgBwQBUIS6yKEA

I assume we could try these configs, but I'm not sure how (whether?) rollup would further transpile down to downlevel ES code, or whether it'd emit fancy ES2021+ code including class fields, and would that break older Node clients? Seems worthy of investigation... in another PR. :-)

12wrigja:
I tried out your example, but even with adjusting those emit options I don't think I can get the runtime retrieval of the value to work: `console.log(X.prototype[Symbol.toStringTag])` prints out `undefined`. I think this will have to wait for another PR (though I am going to explore getting some nicer typing into the intrinsic class container at some point this week).

R.e. the pr tracking: there's no test or structural changes here, so I think this can just be merged. My current litmus test for that is whether or not the 262 tests need to change - if so, then that should land in the proposal repo first.

PR #54, opened 2021-09-22
Type the intrinsics container, and remove usage of ESGetIntrinsic.

12wrigja:
ESGetIntrinsic is unused, and this now allows correct typing for types retrieved from the intrinsic container.

### DIFF
--- a/lib/intrinsicclass.ts
+++ b/lib/intrinsicclass.ts
@@ -1,9 +1,33 @@
-import ESGetIntrinsic from 'es-abstract/GetIntrinsic.js';
+import { Temporal } from '.';
 
 import { DEBUG } from './debug';
-const INTRINSICS = {};
 
-const customUtilInspectFormatters = {
+type TemporalIntrinsics = Omit<typeof Temporal, 'Now'>;
+type TemporalIntrinsicRegistrations = {
+  [key in keyof TemporalIntrinsics as `Temporal.${key}`]: TemporalIntrinsics[key];
+};
+type TemporalIntrinsicPrototypeRegistrations = {
+  [key in keyof TemporalIntrinsics as `Temporal.${key}.prototype`]: TemporalIntrinsics[key]['prototype'];
+};
+type TemporalIntrinsicRegisteredKeys = {
+  [key in keyof TemporalIntrinsicRegistrations as `%${key}%`]: TemporalIntrinsicRegistrations[key];
+};
+
+interface StandaloneIntrinsics {
+  'Temporal.Calendar.from': typeof Temporal.Calendar.from;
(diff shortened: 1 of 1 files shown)


justingrant (approved):
LGTM. I admit I don't fully understand why GetIntrinsic is there at all. Why not just import directly?  I'm sure there's a good reason, but was curious.

12wrigja:
My guess is it's another "written as per the spec thing", and we could potentially do away with it all so long as there wasn't real cyclic dependency between the files. I might revisit this and do exactly that later on if it turns out that this pattern doesn't optimize very well / direct import optimizes better.

lib/intrinsicclass.ts
```diff
@@ -1,9 +1,33 @@
...
+};
+type TemporalIntrinsicPrototypeRegistrations = {
+  [key in keyof TemporalIntrinsics as `Temporal.${key}.prototype`]: TemporalIntrinsics[key]['prototype'];
+};
+type TemporalIntrinsicRegisteredKeys = {
+  [key in keyof TemporalIntrinsicRegistrations as `%${key}%`]: TemporalIntrinsicRegistrations[key];
+};
+
+interface StandaloneIntrinsics {
+  'Temporal.Calendar.from': typeof Temporal.Calendar.from;
```
ptomato, line 17: It's possible that Temporal.Calendar.from no longer needs to be an intrinsic. We could maybe just remove this, it's worth checking?

Temporal.TimeZone.prototype.getOffsetNanosecondsFor is still needed but FWIW https://github.com/tc39/proposal-temporal/issues/1724 is under discussion which would make it no longer necessary.

ptomato (commented):
:+1:

lib/intrinsicclass.ts
justingrant: Dumb question: what's an "intrinsic" and why do we care about them?

12wrigja: .... I actually had the exact same question. Jordan has a proposal here that I think aligns with my thinking: https://github.com/ljharb/proposal-get-intrinsic

The TL;DR seems to be that an Intrinsic is something that is "built-in" to the environment and "user-land" code can't adjust it. For example, I suppose it's theoretically possible for users of this polyfill to overwrite `Temporal.Calendar.from` to do whatever they like, and that would break the assumptions of the rest of the polyfill.

ptomato: That's exactly right.

The intrinsics defined in this file correspond to the ones that are defined by the Temporal specification using the _%Temporal.x.y%_ notation, and used somewhere within the proposal. There are also other intrinsics that we capture at the top of each file like `const ReflectApply = Reflect.apply;`

PR #56, opened 2021-09-24
Remove es-abstract dependency.

12wrigja:


justingrant (approved):
🎉

justingrant:
Is the next step to break up the `ES` object in ecmascript.ts into separate exported functions and then to import those functions (not `ES` overall) in all other files?

12wrigja:
I'm not sure it's strictly needed, but it seems like a relatively easy thing to do (and it's considered idiomatic Google TypeScript to not have "container objects" like this) so I was going to send a PR for that today.

justingrant:
> I'm not sure it's strictly needed, but it seems like a relatively easy thing to do (and it's considered idiomatic Google TypeScript to not have "container objects" like this) so I was going to send a PR for that today.

Cool. I assume the main reason to do that would be to help with tree-shaking?

12wrigja:
My understanding is that tree-shaking is easier when the things to be shaken are module exports, but that this might only apply to some build tools and not others. E.g. I believe Closure Compiler can handle this and DCE the unused functions.

justingrant:
Yep. Not sure if webpack's tree-shaking can handle it.  FWIW, I could help you test it by trying to build one of my apps (which is bundled with webpack) before and after your upcoming PR. I'll let you know if there's a bundle size diff.

PR #57, opened 2021-09-24
Collapse ecmascript.ts container.

12wrigja:
This drastically improves the ability of TypeScript to infer types of the functions that were previously within the container.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -937,7 +937,7 @@ export namespace Temporal {
       millisecond?: number,
       microsecond?: number,
       nanosecond?: number,
-      calendar?: CalendarProtocol
+      calendar?: CalendarProtocol | string
     );
     readonly era: string | undefined;
     readonly eraYear: number | undefined;
--- a/index.d.ts
+++ b/index.d.ts
@@ -814,7 +814,7 @@ export namespace Temporal {
       one: Temporal.PlainDate | PlainDateLike | string,
       two: Temporal.PlainDate | PlainDateLike | string
     ): ComparisonResult;
-    constructor(isoYear: number, isoMonth: number, isoDay: number, calendar?: CalendarProtocol);
+    constructor(isoYear: number, isoMonth: number, isoDay: number, calendar?: CalendarProtocol | string);
     readonly era: string | undefined;
     readonly eraYear: number | undefined;
     readonly year: number;
--- a/lib/calendar.ts
+++ b/lib/calendar.ts
@@ -1,5 +1,5 @@
 import { DEBUG } from './debug';
-import { ES } from './ecmascript';
+import * as ES from './ecmascript';
 import { GetIntrinsic, MakeIntrinsicClass, DefineIntrinsic } from './intrinsicclass';
 import {
   CALENDAR_ID,
--- a/lib/ecmascript.ts
+++ b/lib/ecmascript.ts
@@ -3861,7 +3861,7 @@ export function RoundTime(
       quantity = quantity.multiply(1000).plus(nanosecond);
   }
   const nsPerUnit = unit === 'day' ? dayLengthNs : nsPerTimeUnit[unit];
-  const rounded = RoundNumberToIncrement(quantity, bigInt(nsPerUnit).multiply(increment), roundingMode);
+  const rounded = RoundNumberToIncrement(quantity, nsPerUnit * increment, roundingMode);
   const result = rounded.divide(nsPerUnit).toJSNumber();
   switch (unit) {
     case 'day':
--- a/lib/ecmascript.ts
+++ b/lib/ecmascript.ts
@@ -10,7 +10,6 @@ const NumberIsNaN = Number.isNaN;
 const NumberIsFinite = Number.isFinite;
 const NumberCtor = Number;
 const NumberMaxSafeInteger = Number.MAX_SAFE_INTEGER;
-const ObjectAssign = Object.assign;
 const ObjectCreate = Object.create;
 const ObjectDefineProperty = Object.defineProperty;
 const ObjectIs = Object.is;
@@ -63,13 +62,13 @@ const YEAR_MIN = -271821;
 const YEAR_MAX = 275760;
 const BEFORE_FIRST_DST = bigInt(-388152).multiply(1e13); // 1847-01-01T00:00:00Z
 
-function IsInteger(value: unknown): value is number {
+export function IsInteger(value: unknown): value is number {
   if (typeof value !== 'number' || !NumberIsFinite(value)) return false;
   const abs = MathAbs(value);
   return MathFloor(abs) === abs;
 }
 
-function Type(value: unknown): string {
+export function Type(value: unknown): string {
   if (value === null) return 'Null';
   switch (typeof value) {
     case 'symbol':
@@ -90,12 +89,12 @@ function Type(value: unknown): string {
   }
 }
 
-function ToNumber(value: unknown): number {
+export function ToNumber(value: unknown): number {
   if (typeof value === 'bigint') throw new TypeError('Cannot convert BigInt to number');
   return NumberCtor(value);
 }
 
-function ToInteger(value: unknown): number {
+export function ToInteger(value: unknown): number {
   const num = ToNumber(value);
   if (NumberIsNaN(num)) return 0;
   const integer = MathTrunc(num);
@@ -103,22 +102,22 @@ function ToInteger(value: unknown): number {
   return integer;
 }
 
-function ToString(value: unknown): string {
+export function ToString(value: unknown): string {
   if (typeof value === 'symbol') {
     throw new TypeError('Cannot convert a Symbol value to a String');
   }
   return String(value);
 }
 
-const ToIntegerThrowOnInfinity = (value) => {
-  const integer = ES.ToInteger(value);
+export function ToIntegerThrowOnInfinity(value) {
+  const integer = ToInteger(value);
   if (!NumberIsFinite(integer)) {
     throw new RangeError('infinity is out of range');
   }
   return integer;
-};
+}
 
-const ToPositiveInteger = (value, property?: string) => {
+export function ToPositiveInteger(value, property?: string) {
   value = ToInteger(value);
   if (!NumberIsFinite(value)) {
     throw new RangeError('infinity is out of range');
@@ -130,14 +129,15 @@ const ToPositiveInteger = (value, property?: string) => {
     throw new RangeError('Cannot convert a number less than one to a positive integer');
   }
   return value;
-};
-const ToIntegerNoFraction = (value) => {
-  value = ES.ToNumber(value);
-  if (!ES.IsInteger(value)) {
+}
+
+export function ToIntegerNoFraction(value) {
+  value = ToNumber(value);
+  if (!IsInteger(value)) {
     throw new RangeError(`unsupported fractional value ${value}`);
   }
   return value;
-};
+}
 
 const BUILTIN_CASTS = new Map([
   ['year', ToIntegerThrowOnInfinity],
@@ -192,14 +192,6 @@ const SINGULAR_PLURAL_UNITS = [
 
 import * as PARSE from './regex';
 
-const ES2020 = {
-  IsInteger,
-  ToInteger,
-  ToNumber,
-  ToString,
-  Type
-};
-
 const IntlDateTimeFormatEnUsCache = new Map();
 
 function getIntlDateTimeFormatEnUsForTimeZone(timeZoneIdentifier) {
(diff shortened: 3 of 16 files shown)


lib/ecmascript.ts
```diff
@@ -221,594 +213,299 @@ function getIntlDateTimeFormatEnUsForTimeZone(timeZoneIdentifier) {
...
+    const ns = ParseTemporalInstant(stringIdent);
+    const offsetNs = GetIANATimeZoneOffsetNanoseconds(ns, result);
+    if (FormatTimeZoneOffsetString(offsetNs) !== offset) {
+      throw new RangeError(`invalid offset ${offset}[${ianaName}]`);
     }
+  }
+  return result;
+}
+
+export function FormatCalendarAnnotation(id, showCalendar) {
```
justingrant, line 277:
```suggestion
export function FormatCalendarAnnotation(id: string, showCalendar: 'auto' | 'never') {
```

At some point the arguments to all exported ecmascript.js methods should probably get type annotations. Is your plan to do this in a follow-up PR?  Or would it make sense to do it in this one? I don't have an opinion either way, just wondering.

lib/ecmascript.ts
```diff
@@ -836,17 +883,350 @@ export const ES = ObjectAssign({}, ES2020, {
...
+  if (!any) {
+    throw new TypeError('no supported properties found');
+  }
+  if ((result['era'] === undefined) !== (result['eraYear'] === undefined)) {
+    throw new RangeError("properties 'era' and 'eraYear' must be provided together");
+  }
+  return result;
+}
+
+export function // field access in the following operations is intentionally alphabetical
```
justingrant, line 1007: Should move the comment before `export function`

lib/ecmascript.ts
```diff
@@ -221,594 +213,299 @@ function getIntlDateTimeFormatEnUsForTimeZone(timeZoneIdentifier) {
...
-        canonicalIdent = canonicalIdent.toString();
-        if (ES.ParseOffsetString(canonicalIdent) !== null) return { offset: canonicalIdent };
-        return { ianaName: canonicalIdent };
-      }
-    } catch {
-      // fall through
+    !HasSlot(item, ISO_YEAR, ISO_MONTH, ISO_DAY)
+  );
+}
+export function IsTemporalDateTime(item) {
```
justingrant, line 238: I found it frustrating when reviewing that the old code and new code doesn't line up in the diff, even though almost all the code is identical. Could this be fixed by restoring the old indenting (indent all the new stuff two spaces) during the review of this PR?

Then right before merging you could add another commit to remove the extra indent. 

I'm not sure if this is enough to solve the diff problem, but seems worth a try to make it easier to review.

justingrant (commented):
I didn't see anything needed to change (other than one comment) but the diff made it really hard to review so I'm not 100% comfortable giving this a thumbs up yet. See my comment below re: the hard-to-review issue and an idea to fix it.

lib/ecmascript.ts
ptomato: Undocumented GitHub trick that someone once told me: you can append `?w=1` to the URL to print a diff that ignores whitespace and indentation changes. (It's still a large diff, but this cuts it down to about a third of the size, and the blocks match up better.)

lib/ecmascript.ts
```diff
@@ -836,17 +883,350 @@ export const ES = ObjectAssign({}, ES2020, {
...
+        any[property] = callerCast(value);
+      } else {
+        any[property] = value;
+      }
+    }
+  }
+  return any ? any : undefined;
+}
+
+interface TemporalFields {
```
justingrant, line 964: This is very similar to the publicly-exposed type `Temporal.ZonedDateTimeLike`. The only difference is lack of `calendar` and the type of the `timeZone` property.  Should we derive this type from the public type or leave it as-is here?  I don't have an opinion either way, but seems like we should either be consistently re-using public types where we can, or should never be re-using them.

lib/ecmascript.ts
justingrant: Wow that's a great trick!  It totally works!  @12wrigja, never mind my suggestion above: @ptomato's trick does the same thing without messing with commits. 

BTW, looks like there's a UI option for getting this behavior:

lib/calendar.ts
```diff
@@ -1,5 +1,5 @@
 import { DEBUG } from './debug';
-import { ES } from './ecmascript';
+import * as ES from './ecmascript';
```
justingrant, line 2: At first I was concerned that using `import *` would break tree-shaking, but according to https://stackoverflow.com/a/64569600 this is fine. 

But... I'd still like us to verify that TS's transpilation doesn't break webpack's ability to tree-shake this. Any idea how to test this without doing lots of probably-unnecessary work to import each function separately in each non-ecmascript.ts file?  I guess we could just look at how TS emits JS for this case, specifically whether the underlying JS is still `import *` or if it's transpiled into something that will break tree shaking.

lib/ecmascript.ts
```diff
@@ -2095,1136 +1976,1397 @@ export const ES = ObjectAssign({}, ES2020, {
...
-        nanoseconds
-      );
-      const startNs = GetSlot(relativeTo, EPOCHNANOSECONDS);
-      nanoseconds = endNs.subtract(startNs);
+      return epochNanoseconds;
+    })
+    .filter((x) => x !== undefined);
+}
+
+export function LeapYear(year) {
```
justingrant, line 2515: Is there an easy way to determine whether all of the exported functions in the old ES object are called in any other files?  This function is called (twice in calendar.ts) but I suspect that at least one ecmasript.ts function probably snuck into the ES object while only being used inside ecmascript.ts itself.

lib/ecmascript.ts
```diff
@@ -3233,444 +3375,468 @@ export const ES = ObjectAssign({}, ES2020, {
...
-          microseconds,
-          nanoseconds,
-          largestUnit
-        ));
-      } else {
-        ({ years, months, weeks, days, hours, minutes, seconds, milliseconds, microseconds, nanoseconds } =
-          ES.DifferenceZonedDateTime(GetSlot(relativeTo, EPOCHNANOSECONDS), endNs, timeZone, calendar, largestUnit));
-      }
-    }
+export function DifferenceZonedDateTime(ns1, ns2, timeZone, calendar, largestUnit, options?: any) {
```
justingrant, line 3399: if we're going to give options a type annotation, seems like we should use the correct one.

lib/ecmascript.ts
```diff
@@ -3233,444 +3375,468 @@ export const ES = ObjectAssign({}, ES2020, {
...
+  months,
+  weeks,
+  days,
+  h,
+  min,
+  s,
+  ms,
+  µs,
+  ns,
+  options?: any
```
justingrant, line 3751: same as above

lib/ecmascript.ts
```diff
@@ -3233,444 +3375,468 @@ export const ES = ObjectAssign({}, ES2020, {
...
+  hour,
+  minute,
+  second,
+  millisecond,
+  microsecond,
+  nanosecond,
+  increment,
+  unit,
+  roundingMode,
+  dayLengthNs: bigInt.BigInteger | number = 86400e9
```
justingrant, line 3836: Is it valid to assign a non-`n` literal to a bigint?  Something seems fishy here. Seems like it should either be a number or a bigint, but mixing them seems to invite bugs.  Not saying this needs to be fixed in this PR-- probably better somewhere else--but at least wanted to flag it here.

lib/ecmascript.ts
```diff
@@ -3680,151 +3846,204 @@ export const ES = ObjectAssign({}, ES2020, {
...
```
justingrant, line 3865: same as above

lib/ecmascript.ts
```diff
@@ -3850,397 +4063,366 @@ export const ES = ObjectAssign({}, ES2020, {
...
+  hours,
+  minutes,
+  seconds,
+  milliseconds,
+  microseconds,
+  nanoseconds,
+  increment,
+  unit,
+  roundingMode,
+  relativeTo = undefined
```
justingrant, line 4098: Should this be `relativeTo?`? Is there any difference between marking as optional in TS and giving it a default value of `undefined`?

lib/ecmascript.ts
```diff
@@ -3850,397 +4063,366 @@ export const ES = ObjectAssign({}, ES2020, {
...
-  GetNumberOption: (options, property, minimum, maximum, fallback) => {
-    let value = options[property];
-    if (value === undefined) return fallback;
-    value = ES.ToNumber(value);
-    if (NumberIsNaN(value) || value < minimum || value > maximum) {
-      throw new RangeError(`${property} must be between ${minimum} and ${maximum}, not ${value}`);
+// Note: This method returns values with bogus nanoseconds based on the previous iteration's
+// milliseconds. That way there is a guarantee that the full nanoseconds are always going to be
+// increasing at least and that the microsecond and nanosecond fields are likely to be non-zero.
+export const SystemUTCEpochNanoSeconds = (() => {
```
justingrant, line 4380: This code is weird enough (took me a few times reading it before I understood what it was doing and why it was a const instead of an exported function) that it probably deserves an explicit type annotation on `SystemUTCEpochNanoSeconds`.

lib/ecmascript.ts
```diff
@@ -221,594 +213,299 @@ function getIntlDateTimeFormatEnUsForTimeZone(timeZoneIdentifier) {
...
+  month: number;
+  day: number;
+  hour: number;
+  minute: number;
+  second: number;
+  millisecond: number;
+  microsecond: number;
+  nanosecond: number;
+  ianaName: string;
+  offset: any;
```
justingrant, line 422: Can we use a real type here?  There are only a few places where `any` is used, so seems like a good oppty to replace those with real types.

lib/ecmascript.ts
```diff
@@ -1294,798 +1484,489 @@ export const ES = ObjectAssign({}, ES2020, {
...
+  const calMergeFields = calendar.mergeFields;
+  if (!calMergeFields) {
+    return { ...fields, ...additionalFields };
+  }
+  const result = Reflect.apply(calMergeFields, calendar, [fields, additionalFields]);
+  if (Type(result) !== 'Object') throw new TypeError('bad return from calendar.mergeFields()');
+  return result;
+}
+
+export function CalendarDateAdd(calendar, date, duration, options, dateAdd?: any) {
```
justingrant, line 1675: possible to replace `any` with a real type?

lib/ecmascript.ts
```diff
@@ -1294,798 +1484,489 @@ export const ES = ObjectAssign({}, ES2020, {
...
+  if (dateAdd === undefined) {
+    dateAdd = calendar.dateAdd;
+  }
+  const result = ReflectApply(dateAdd, calendar, [date, duration, options]);
+  if (!IsTemporalDate(result)) throw new TypeError('invalid result');
+  return result;
+}
+
+export function CalendarDateUntil(calendar, date, otherDate, options, dateUntil?: any) {
```
justingrant, line 1684: same as above

lib/ecmascript.ts
```diff
@@ -2095,1136 +1976,1397 @@ export const ES = ObjectAssign({}, ES2020, {
...
+    -GetSlot(duration, MILLISECONDS),
+    -GetSlot(duration, MICROSECONDS),
+    -GetSlot(duration, NANOSECONDS)
+  );
+}
+
+export function ConstrainToRange(value, min, max) {
+  return MathMin(max, MathMax(min, value));
+}
+export function ConstrainISODate(year, month, day?: any) {
```
justingrant, line 3078: possible to replace `any` with a real type?

lib/ecmascript.ts
```diff
@@ -1294,798 +1484,489 @@ export const ES = ObjectAssign({}, ES2020, {
...
+  if (sOne === sTwo || sOne === 'iso8601') {
+    return two;
+  } else if (sTwo === 'iso8601') {
+    return one;
+  } else {
+    throw new RangeError('irreconcilable calendars');
+  }
+}
+
+export function DateFromFields(calendar, fields, options?: any) {
```
justingrant, line 1826: Should the options type be `Temporal.AssignmentOptions` ?  That's what's used in index.d.ts, although if that's not accurate then we should fix it.

lib/ecmascript.ts
```diff
@@ -1294,798 +1484,489 @@ export const ES = ObjectAssign({}, ES2020, {
...
+  const result = calendar.dateFromFields(fields, options);
+  if (!IsTemporalDate(result)) throw new TypeError('invalid result');
+  return result;
+}
+
+export function YearMonthFromFields(calendar, fields, options?: any) {
```
justingrant, line 1832: same as above

lib/ecmascript.ts
```diff
@@ -1294,798 +1484,489 @@ export const ES = ObjectAssign({}, ES2020, {
...
+  const result = calendar.yearMonthFromFields(fields, options);
+  if (!IsTemporalYearMonth(result)) throw new TypeError('invalid result');
+  return result;
+}
+
+export function MonthDayFromFields(calendar, fields, options?: any) {
```
justingrant, line 1838: same as above

justingrant (approved):
This was a lot of work, thanks for doing it!  Overall looks great. A few suggestions in comments, otherwise all good. 

A good follow-up for this PR would be to add type annotations to the parameters of all exported methods in ecmascript.ts.

lib/ecmascript.ts
12wrigja: Ah yeah - I was playing around with that setting yesterday with the hope that this would make things a bit easier. There's at least a couple places where the diffs still doesn't line up much, but it does reduce the diffs down significantly.

That being said, I still plan to split out as many changes as possible and land those separately (e.g. most new type annotations, runtime changes etc) before sending this for review.

lib/ecmascript.ts
```diff
@@ -836,17 +883,350 @@ export const ES = ObjectAssign({}, ES2020, {
...
+  }
+  return 'nanosecond';
+}
+
+export function LargerOfTwoTemporalUnits(unit1, unit2) {
+  if (ALLOWED_UNITS.indexOf(unit1) > ALLOWED_UNITS.indexOf(unit2)) return unit2;
+  return unit1;
+}
+
+export function CastIfDefined(value, cast) {
```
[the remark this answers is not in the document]
12wrigja: I'm unsure, but the nice part about moving these to exported functions is that tsc should be able to point this out to us now :) If so, I'll send a PR to remove it before flattening.

lib/ecmascript.ts
justingrant: > That being said, I still plan to split out as many changes as possible and land those separately (e.g. most new type annotations, runtime changes etc) before sending this for review.

Would it be easier to land this PR before putting type annotations on every exported function parameter?  I ask because it seems like TS seems to struggle with typing the ES object, perhaps because it has so many methods. On my fork, ES resolved to just `any`.  So I suspect it will be easier to find type errors if you first convert from `export const ES` => separate exports. But you're closer to this than me, so if you have a better plan feel free to do that instead!

I agree with fixing obvious type errors that show up simply as a result of doing the  `export const ES` => separate exports. Mostly thinking of the next-level-down typing that's not required to make this PR go green. (Except for the `any` ones that IMHO make sense to fix here.)

lib/ecmascript.ts
```diff
@@ -817,16 +514,366 @@ export const ES = ObjectAssign({}, ES2020, {
...
+    second: 60,
+    millisecond: 1000,
+    microsecond: 1000,
+    nanosecond: 1000
+  };
+  return ToTemporalRoundingIncrement(options, maximumIncrements[smallestUnit], false);
+}
+
+export function ToSecondsStringPrecision(options): {
+  precision: 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 'auto' | 'minute';
```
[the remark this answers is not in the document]
ptomato: IIUC, the public API in index.d.ts has `fractionalSecondDigits` and `smallestUnit` options, this internal structure is the consolidated version that combines the information provided by both of those public options.

lib/ecmascript.ts
ptomato: This pre-existing fishiness is my fault for being in a hurry to get ZonedDateTime working... I never had time to track down all the places where this was inconsistent. Using tsc it should be easier to fix, hopefully.

lib/ecmascript.ts
```diff
@@ -817,16 +514,366 @@ export const ES = ObjectAssign({}, ES2020, {
...
+    case 7:
+    case 8:
+    case 9:
+      return { precision, unit: 'nanosecond', increment: 10 ** (9 - precision) };
+    default:
+      throw new RangeError(`fractionalSecondDigits must be 'auto' or 0 through 9, not ${digits}`);
+  }
+}
+
+export function ToLargestTemporalUnit(options, fallback, disallowedStrings = [], autoValue?: any) {
```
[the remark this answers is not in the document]
12wrigja: Any is the current type, but it's definitely not the actual type :) As you mentioned elsewhere, I'll send a separate PR to specify these as actual types.

The only reason these changes are needed is so we can call this function without passing either the `disallowedStrings` or `autoValue` parameters - otherwise, tsc complains that we are missing parameters in the various calls.

lib/ecmascript.ts
12wrigja: As you mention, I plan to land this and then change the types. In many places, the type inference that tsc will now be able to do should help me greatly in deciding how to type things (though, especially for e.g. options bags if you know that there's a pre-defined type from e.g. the public API that is passed all the way through, then it would be awesome if you wanted to send a PR to use that once this lands).

lib/ecmascript.ts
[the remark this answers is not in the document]
12wrigja: Gotcha. For the moment, I'm going to leave this as-is, and I'll send a separate PR to create the right type for this once I've had a chance to look at it more closely. I think there's some type-magic we can do to automatically infer the right type based off the public types - this type is just generated based on a quick look over the code and what values it specifically checks for.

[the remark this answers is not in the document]
justingrant: > IIUC, the public API in index.d.ts has `fractionalSecondDigits` and `smallestUnit` options, this internal structure is the consolidated version that combines the information provided by both of those public options.

Makes sense, thanks for clarifying.

lib/calendar.ts
12wrigja: The underlying JS that tsc emits is:

```
import * as ES from './ecmascript';
```

(from `./tsc-out/calendar.js` after running `npm run build`).

At that point, so long as Webpack says they support tree-shaking of this pattern we should be good?

lib/plaindatetime.ts
```diff
@@ -31,7 +31,7 @@ export class PlainDateTime implements Temporal.PlainDateTime {
     millisecond = 0,
     microsecond = 0,
     nanosecond = 0,
-    calendar = ES.GetISO8601Calendar()
+    calendar: Temporal.CalendarProtocol | string = ES.GetISO8601Calendar()
```
12wrigja, line 34: Similarly, should this be reflected back into the types? `Temporal.PlainDateTime` doesn't accept a string here, just a `CalendarProtocol`, but there's definitely places it's passed a string in the implementation.

lib/ecmascript.ts
12wrigja: Should this be treated as a string?

lib/ecmascript.ts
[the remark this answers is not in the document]
12wrigja: I'm going to send this along with the PR to improve types instead.

lib/ecmascript.ts
12wrigja: I think there might be a way to make this use the TS Mapped Types functionality to be able to type it based on the fields you give it. Will investigate in a follow-up PR.

I'm slightly hesitant to re-use the public API types given that those might change and while that might be right for the external API it could easily become incorrect for part of the internal API.

lib/ecmascript.ts
12wrigja: Will do in a follow-up - this is just to get tsc to allow calling this without that particular parameter.

lib/ecmascript.ts
12wrigja: Will do in a follow-up - this is just to get tsc to allow calling this without that particular parameter.

lib/ecmascript.ts
12wrigja: Done.

index.d.ts
```diff
@@ -814,7 +814,7 @@ export namespace Temporal {
       one: Temporal.PlainDate | PlainDateLike | string,
       two: Temporal.PlainDate | PlainDateLike | string
     ): ComparisonResult;
-    constructor(isoYear: number, isoMonth: number, isoDay: number, calendar?: CalendarProtocol);
+    constructor(isoYear: number, isoMonth: number, isoDay: number, calendar?: CalendarProtocol | string);
```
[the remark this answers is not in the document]
ptomato: Yes, this is intended to be part of the API: https://tc39.es/proposal-temporal/#sec-temporal-totemporalcalendar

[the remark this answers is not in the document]
justingrant: Yep, looks like the spec accepts this too, so this change looks fine.

lib/plaindatetime.ts
justingrant: Yes, it looks like the spec accepts strings in the constructor for PlainXXX (except PlainTime) and ZonedDateTime. So it should be OK to update the public types of those constructors.

lib/ecmascript.ts
justingrant: SGTM. Try this: `npx ts-unused-exports ./tsconfig.json lib/ecmascript.ts`. 

If that doesn't work, there's always manual brute force + muscle memory: 

1. In VSCode, open all the other /lib/*.ts files in the editor, so that TS will find errors in them without having to run anything.
2. Search for `export ` in ecmascript.mjs
3. Hit the Delete key.
4. Are there errors showing up in VSCode's "problems" pane?  If not, use Find All References to verify that the function is unused, and if so then delete it!
5. If not, then Cmd+Z to undo
6. Cmd+G to advance to the next match
7. GOTO 3

This is manual but after the first few iterations it will sink into muscle memory movements so probably could get it done in 10-20 (boring) minutes. 😄

lib/ecmascript.ts
justingrant: > Is it ever possible for the length of a day to be longer than number-size nanoseconds?

No. MAX_SAFE_INTEGER nanoseconds is 104 days. 

I'm not sure about using bigint internally. Others may have opinions about this. 

I assume a reasonable rule should be this: if the value is only used in calculations with bigints, then make it a bigint like you did with DAYNANOS. Otherwise, then it's probably a case-by-case decision.

lib/calendar.ts
justingrant: Yeah, I think we're good. If we find out otherwise, can always fix in a later PR.

lib/ecmascript.ts
```diff
@@ -3232,444 +3360,468 @@ export const ES = ObjectAssign({}, ES2020, {
...
+  );
+
+  // Note that 'compatible' is used below because this disambiguation behavior
+  // is required by RFC 5545.
+  const instantIntermediate = BuiltinTimeZoneGetInstantFor(timeZone, dtIntermediate, 'compatible');
+  return AddInstant(GetSlot(instantIntermediate, EPOCHNANOSECONDS), h, min, s, ms, µs, ns);
+}
+
+export function RoundNumberToIncrement(quantity, increment, mode) {
+  if (increment === 1) return quantity;
```
justingrant, line 3777: From a quick look at usage of this function, it seems like increment is usually a Number, not a bigint. Also, `1n === 1` is `false`, so if the inputs are bigint then this test will never succeed. So I think it's good to type `increment` as a Number and `quantity` is a BigInt.

lib/ecmascript.ts
```diff
@@ -3679,151 +3831,204 @@ export const ES = ObjectAssign({}, ES2020, {
...
+      quantity = quantity.multiply(1000).plus(millisecond);
+    // fall through
+    case 'microsecond':
+      quantity = quantity.multiply(1000).plus(microsecond);
+    // fall through
+    case 'nanosecond':
+      quantity = quantity.multiply(1000).plus(nanosecond);
+  }
+  const nsPerUnit = unit === 'day' ? dayLengthNs : bigInt(nsPerTimeUnit[unit]);
+  const rounded = RoundNumberToIncrement(quantity, nsPerUnit.multiply(increment), roundingMode);
```
justingrant, line 3874: It seems that `RoundNumberToIncrement` method expects increment to be a Number, not a bigint.

lib/ecmascript.ts
justingrant: OK sounds good. I know that in the public API, relativeTo can be a PlainDateTime or a ZonedDateTime (or undefined) but not sure how that shakes out down here in the implementation.

justingrant (approved):
Overall this all looks good so far, modulo a few outstanding comments and a few new comments. OK by me to merge when they're resolved.

lib/ecmascript.ts
```diff
@@ -3232,444 +3360,468 @@ export const ES = ObjectAssign({}, ES2020, {
...
+  hour,
+  minute,
+  second,
+  millisecond,
+  microsecond,
+  nanosecond,
+  increment,
+  unit,
+  roundingMode,
+  dayLengthNs = DAY_NANOS
```
justingrant, line 3821: Per comments below, I think this should probably be a Number, not a bigint

lib/ecmascript.ts
```diff
@@ -3679,151 +3831,204 @@ export const ES = ObjectAssign({}, ES2020, {
...
```
justingrant, line 3850: Same here: I think this should be a Number not a bigint

lib/ecmascript.ts
justingrant: > usually a Number

FWIW, I think "is always a Number except for bugs" is probably more accurate. @ptomato may have more insight.

ptomato: That may be the case, but it probably _ought_ to be a BigInt. The big-integer wrapper library we use allows sloppy mixing of BigInt and Number types, where real BigInts wouldn't.

12wrigja: FWIW, JSBI's typings doesn't allow this (at least, for some of the functions), so deciding on one or the other and using it consistently will be better enforced as that migration + better typings are landed throughout.

It seems that they should both be bigints?

lib/ecmascript.ts
justingrant: Looking upstream from this code, I think the root-cause bug in the current `main` (before this PR) is that the implementation of `round()` in zoneddatetime.ts passes a bigint for `dayLengthNs`. AFAICT, in current main this is the only case that leads to a bigint being passed as `increment`. 

For the purposes of this PR, I suspect that the easiest path would be to change `ZonedDateTime.p.round` to convert `dayLengthNs` to Number before passing to RoundISODateTime. Then (assuming there isn't other use of bigint in these values) all downstream code (including the lines above this comment) could use Number for `dayLengthNs` and `increment` parameters, and we'd only have to JIT-coerce them to bigint before combining them in arithmetic operations with bigint values.

I think it's reasonable to discuss whether we'd want to convert these and similar parameters to bigint, but that seems out of scope to this PR, and might have a perf or bundle-size impact so IMHO seems worth considering separately after this PR is merged.

lib/ecmascript.ts
```diff
@@ -3232,444 +3360,468 @@ export const ES = ObjectAssign({}, ES2020, {
...
+      break;
+    case 'floor':
+      if (sign < 0) quotient = quotient.add(sign);
+      break;
+    case 'trunc':
+      // no change needed, because divmod is a truncation
+      break;
+    case 'halfExpand':
+      // "half up away from zero"
+      if (remainder.multiply(2).abs() >= increment) quotient = quotient.add(sign);
```
justingrant, line 3793: In addition to the zoneddatetime.ts bug described in https://github.com/js-temporal/temporal-polyfill/pull/57#discussion_r718042170, this line also looks like a bug if `increment` is a Number.

lib/ecmascript.ts
justingrant: > It seems that they should both be bigints?

I dug into where `increment` is used. It's clear that `increment` (which often comes from user-entered Number-typed options) is always a Number. I found only one exception which looks like it's caused by a bug in the ZonedDateTime implementation. See  https://github.com/js-temporal/temporal-polyfill/pull/57#discussion_r718042170. @12wrigja are there others?

Changing all usage of `increment` (~60 lines) to `bigint` seems like a pretty significant refactor with non-trivial impact on the types of other variables and function parameters. I'd be hesitant to do it before this PR is merged and before we have non-`any` types on parameters and return types of ES methods, because having everything typed would make it much easier to understand downstream impact and avoid mistakes.  I'd also like to understand the bundle-size impact of a big refactor like that, esp. in the case of `increment` where AFAICT always smaller than MAX_SAFE_INTEGER. 

For this PR (or a separate PR before merging this one), my suggestion is that we leave usage as Number and do the minimal required to fix known type bugs, e.g. `ZonedDateTime.p.round()` should coerce `dayLengthNs` to Number before passing to `RoundISODateTime`. 

Then we can discuss as a follow-up whether (and if so, when) we want to change `increment` (and other similar parameters) to bigint globally.

Would that work?

> > The big-integer wrapper library we use allows sloppy mixing of BigInt and Number types, where real BigInts wouldn't.
>
> JSBI's typings doesn't allow this (at least, for some of the functions), so deciding on one or the other and using it consistently will be better enforced as that migration + better typings are landed throughout.

Stricter types sounds really helpful. Even if our bigint library allows passing `number` params to some methods, we could always wrap it with stricter types ourselves. But I really like the idea of not doing any library math between bigint and integer values if the native operators wouldn't allow it. 

Not sure if making that change would blow up the complexity of this PR though. I'd leave it up to @12wrigja which order is best.

12wrigja: > I found only one exception which looks like it's caused by a bug in the ZonedDateTime implementation. See #57 (comment). @12wrigja are there others?
That was the only one I found as well.

> For this PR (or a separate PR before merging this one), my suggestion is that we leave usage as Number and do the minimal required to fix known type bugs, e.g. ZonedDateTime.p.round() should coerce dayLengthNs to Number before passing to RoundISODateTime.

SGTM, I'll definitely be following up on this and other similar issues once I return to integrating JSBI (hopefully today / later this week).

lib/ecmascript.ts
12wrigja: I've reverted this back to what it is as head, which does type this as a number.

lib/ecmascript.ts
12wrigja: Same.

lib/ecmascript.ts
12wrigja: https://github.com/peterolson/BigInteger.js/blob/4c1f2067768bfdd923b3d5e64392e90f1b748502/BigInteger.js#L1342-L1344

So long as that value isn't larger than Number.MAX_SAFE_INTEGER, this should be fine I think. This code will have to change once we use JSBI, as it doesn't do this (potentially lossy) conversion as all (yay!).

I'm going to return to this once we've migrated to JSBI, as we will likely have to deal with a bunch of these sorts of issues all at once.

lib/plaindatetime.ts
12wrigja: Thanks! I changed this both for PlainDateTime and PlainDate.

12wrigja:
Phew! What a slog - thank you both for bearing with me on this.

I've reverted some of the changes I made to treat `dayLengthNs` as a BigInt - it's now treated as a number (though I suspect this will need to be revisited as I'm integrating JSBI... along with other locations where the implicit conversion of `big-integer`'s `BigInteger` to Number is happening via `valueOf` (as this doesn't happen with JSBI, fortunately).

What's left should mostly be mechanical changes (removal of `ES.`, de-indentation, line-wrapping), and some small type adjustments to allow methods to be called without their optional params (implicit `any` isn't the same as `?: any`).

I'll be following up with PRs to swap to using JSBI and to add types to ecmascript.ts, and I'll definitely be circling back to the conversations here for hints - I simply didn't want to end up "hiding" those changes in such a larger otherwise mechanical PR like this one.

lib/ecmascript.ts
```diff
@@ -3679,151 +3821,204 @@ export const ES = ObjectAssign({}, ES2020, {
...
+      quantity = quantity.multiply(1000).plus(millisecond);
+    // fall through
+    case 'microsecond':
+      quantity = quantity.multiply(1000).plus(microsecond);
+    // fall through
+    case 'nanosecond':
+      quantity = quantity.multiply(1000).plus(nanosecond);
+  }
+  const nsPerUnit = unit === 'day' ? dayLengthNs : nsPerTimeUnit[unit];
+  const rounded = RoundNumberToIncrement(quantity, bigInt(nsPerUnit).multiply(increment), roundingMode);
```
justingrant, line 3864: Per our discussion in https://github.com/js-temporal/temporal-polyfill/pull/57/files?diff=split&w=1#r718741733, I think this change should be reverted.

lib/ecmascript.ts
```diff
@@ -3849,397 +4038,366 @@ export const ES = ObjectAssign({}, ES2020, {
...
+      nanoseconds = bigInt(microseconds).multiply(1e3).plus(nanoseconds);
+      total = nanoseconds.toJSNumber() / divisor;
+      const rounded = RoundNumberToIncrement(nanoseconds, divisor * increment, roundingMode);
+      microseconds = rounded.divide(divisor).toJSNumber();
+      nanoseconds = 0;
+      break;
+    }
+    case 'nanosecond': {
+      total = nanoseconds;
+      nanoseconds = RoundNumberToIncrement(bigInt(nanoseconds), increment, roundingMode);
```
justingrant, line 4282: BTW, interesting that the first parameter of this function does seem to sometimes be a number and sometimes be a bigint, unlike the second (`increment`) param that's always a Number. Good to investigate in a follow-up.

justingrant (approved):
Looks good!  I made one substantive comment at ecmascript.js line 3864 that should probably be fixed, otherwise just some minor suggestions and/or follow-up hints.  Good to merge once that one thing is fixed.

index.d.ts
```diff
@@ -937,7 +937,7 @@ export namespace Temporal {
       millisecond?: number,
       microsecond?: number,
       nanosecond?: number,
-      calendar?: CalendarProtocol
+      calendar?: CalendarProtocol | string
```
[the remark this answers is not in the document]
12wrigja: Will change this in a followup. I'll also send a corresponding PR to the proposal repo to keep this in sync.

lib/ecmascript.ts
```diff
@@ -2095,1135 +1952,1396 @@ export const ES = ObjectAssign({}, ES2020, {
...
+    precision
+  );
+  let timeZoneString = 'Z';
+  if (timeZone !== undefined) timeZoneString = BuiltinTimeZoneGetOffsetStringFor(outputTimeZone, instant);
+  return `${year}-${month}-${day}T${hour}:${minute}${seconds}${timeZoneString}`;
+}
+
+export function TemporalDurationToString(
+  duration,
+  precision: 'auto' | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 = 'auto',
```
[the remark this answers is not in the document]
12wrigja: Will integrate into a followup :) Thanks!

lib/ecmascript.ts
12wrigja: Reverted.

lib/ecmascript.ts
12wrigja: Definitely - I think the eventual signature will be `RoundNumberToIncrement(quantity: BigInt, increment: number, roundingMode: Temporal.RoundingMode)`.

Will fix in a followup.

ptomato (approved):
I agree with merging whatever passes TSC and the tests quickly, and fixing types afterwards.

lib/ecmascript.ts
justingrant: Turns out that `npx ts-prune` works perfectly. See #62 and #63.

justingrant: And we can even make ts-prune into a CI test to detect future unused functions or unused exports. I put this CI test into #62.

PR #58, opened 2021-09-25
Stop treating the System time since Epoch as a Number, and instead treat

12wrigja:
it as a bigint.BigInteger.

tsc found this when I flattened the ES container into exported functions.


justingrant (approved):

lib/ecmascript.ts
```diff
@@ -2266,7 +2266,7 @@ export const ES = ObjectAssign({}, ES2020, {
     return ES.BalanceISODateTime(year, month, day, hour, minute, second, millisecond, microsecond, nanosecond);
   },
   GetIANATimeZoneNextTransition: (epochNanoseconds, id) => {
-    const uppercap = ES.SystemUTCEpochNanoSeconds() + 366 * DAYMILLIS * 1e6;
+    const uppercap = ES.SystemUTCEpochNanoSeconds().plus(366).multiply(DAYMILLIS).multiply(1e6);
```
justingrant, line 2269: Oops, approved too soon. This new code isn't accurate because of order of operations and precedence. 

The original code adds 366 days' worth of nanoseconds. 

The new code adds 366 nanoseconds, then multiplies the new number of nanoseconds by 86400000 and then multiplies again by 10^6. That's a totally different operation. 

AFAIK, 366 days is too large a number of nanoseconds to fit inside MAX_SAFE_INTEGER. (~104 days is the limit, if I remember correctly).  So the easy alternative of `plus(366 * DAYMILLIS * 1e6)` won't work. 

One  suggestion would be to define a bigint constant for 366 days of nanoseconds, and then add that constant in the method above.

justingrant (changes requested):
Oops, approved too soon!

lib/ecmascript.ts
justingrant: Also, I'm curious about test coverage. Is there a test that calls this function? If so, how long does it take to run?  I'd assume that running through the `while` loop below this code an additional 8.64 trillion times might have taken a loooong time, but I didn't see the tests taking much longer. Which made me wonder whether this was being tested properly. (Where "properly" implies using a few dates that are far enough from epoch that the test should be really slow.)

12wrigja: Looking over the implementation of that while loop, I don't think that accidentally creating a very large upper cap makes a huge difference for the current tests as it seems to break out when it detects the transition point, and jumps forward in 2 week increments before breaking out of that loop and bisecting that 2 week chunk to get the actual transition point. That sounds relatively quick to me assuming that the upper cap really is ~1 year - that's only 26 iterations of the loop.

I suppose to test this there needs to be a timezone without any transition points within the upper cap of 366 days after the start point. Is that a thing?

12wrigja: Early morning coffee brain thought: UTC doesn't have transitions, right?

12wrigja: I've looked at this code some more, and now I just have more questions than I started with:
 - why is the upper cap defined based on the current system time + 266 days, instead of the passed in timestamp + 366 days?
  - why does the while loop compare leftNanos, which is (mostly) static and doesn't change per-loop invocation, instead of rightNanos, which is the time being advanced forward by 2 weeks per iteration?

12wrigja: Never mind - one of the testcases explains the first bullet point. Presumably, it checks up to ~the next year from the current time as that's likely how far in advance the transitions are defined in system/browser data, and we do want to support finding transitions that were newly defined more than a year from the given start point in the future.

12wrigja: ... and now the rest of the code makes sense as well. The left and right nanos values are the sliding window which later gets bisected. Comparing leftNanos against the cap makes sense given that it is also moving.

Clearly I need to go get another coffee... Sorry for all the notifications!

justingrant: Not sure it's related to this order-of-operations bug, but the cookbook tests break for the equivalent PR over in https://github.com/tc39/proposal-temporal/pull/1848, so we may want to see why and turn that "why" into a Demitasse test in both places.

> Looking over the implementation of that while loop, I don't think that accidentally creating a very large upper cap makes a huge difference for the current tests

Yep, because we must only be testing only zones with transitions (which won't show the problem) or offset time zones (e.g. `-08:00`) or `UTC` which are special cased and won't run into the bug either. 

> I suppose to test this there needs to be a timezone without any transition points within the upper cap of 366 days after the start point. Is that a thing?

Correct. `Etc/GMT+10` or `Asia/Calcutta` are good candidates to try.

You'll know you've found the right zone when the current code in this PR runs for ~1.17 × 10<sup>17</sup> iterations. 😄  With current computers that's only about a million years. Better get on the AWS Quantum&trade; beta!

> Early morning coffee brain thought: UTC doesn't have transitions, right?

It has no transitions, but UTC is special-cased in the implementation. UTC is still a good test (of the special-casing) but we also need to test another TZ without transitions that isn't special-cased.

justingrant: BTW, I opened #60 to deal with problems discussed above. We probably don't want the GetNextTransition / GetPreviousTransition to be such an easy DoS attack.

justingrant: Also I opened #59 to see if we can run cookbook tests in this repo just like we run Test262.

test/timezone.mjs
```diff
@@ -417,6 +417,10 @@ describe('TimeZone', () => {
   });
   describe('getNextTransition works', () => {
     const nyc = Temporal.TimeZone.from('America/New_York');
+    // At the time of writing, the Kolkata TZ is not scheduled to have any
+    // transitions.
+    const kolkataTz = Temporal.TimeZone.from('Asia/Kolkata');
```
justingrant, line 422: May want to use `Etc/GMT+10` which is guaranteed to *never* have transitions so this test will never break due to India's govt decisions.

lib/ecmascript.ts
```diff
@@ -2266,7 +2266,7 @@ export const ES = ObjectAssign({}, ES2020, {
     return ES.BalanceISODateTime(year, month, day, hour, minute, second, millisecond, microsecond, nanosecond);
   },
   GetIANATimeZoneNextTransition: (epochNanoseconds, id) => {
-    const uppercap = ES.SystemUTCEpochNanoSeconds() + 366 * DAYMILLIS * 1e6;
+    const uppercap = ES.SystemUTCEpochNanoSeconds().plus(bigInt(366).multiply(DAYMILLIS).multiply(1e6));
```
justingrant, line 2269: Minor nit: we don't need bigint math until it goes above 104 days, so no need for two `multiply` calls here.  Only one is needed.

That said, everywhere DAYMILLIS is used, it's multiplied by 1e6 and used as a bigint. So it may make sense as part of this PR to replace DAYMILLIS with `const DAYNANOS = bigInt(86400).multiply(1e9);`.  Could also put our use of 366 days and 14 days into bigint constants instead of calculating on the fly.  Like this:

```suggestion
    const uppercap = ES.SystemUTCEpochNanoSeconds().plus(YEARNANOS);
```

test/timezone.mjs
12wrigja: Done.

lib/ecmascript.ts
12wrigja: Done, at least for defining DAY_NANOS and using that elsewhere. I'm sure there's a tradeoff between initializing all these bigint values once at the beginning vs using them per-call-site, and I suppose in an ideal world the bundlers / optimizers would be able to streamline this as much as possible (e.g. Closure Compiler can optimize some operations on numbers so the resulting bundle is just the value already initialized, but unsure if without native bigint usage that is feasible here / whether that is a thing other OSS tools do).

justingrant: Cool, sounds good.

PR #66, opened 2021-09-30
Update dependencies (except Demitasse) and make it easier to update deps in the future

justingrant:
This PR brings all dependencies (except Demitasse) up to latest. I mainly did this now to get TS 4.4 which includes some significant improvements to type inference based on upstream code.

It will also hopefully quiet the annoying security warning at the top of this page. 😄 

Also adds a new package.json script to make updating deps easier in the future. Now you can run `npm run update` which will update all non-Demitasse deps in package.json to `latest`. Just remember to run `npm i` afterwards to actually install the new deps!  

Note that I didn't install `npm-check-updates` locally because it would have brought in ~300 dependencies. Given that this is run only rarely, a locale install wasn't needed; using `npx` was fine.

Also re-ran prettier on all files, which resulted in one minor change.

### DIFF
--- a/lib/intrinsicclass.ts
+++ b/lib/intrinsicclass.ts
@@ -25,9 +25,9 @@ type customFormatFunction<T> = (
   depth: number,
   options: { stylize: (value: unknown, type: 'number' | 'special') => string }
 ) => string;
-const customUtilInspectFormatters: Partial<
-  { [key in keyof TemporalIntrinsicRegistrations]: customFormatFunction<TemporalIntrinsicRegistrations[key]> }
-> = {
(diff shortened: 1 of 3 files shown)


12wrigja (approved):

ptomato (commented):
Seems good, but I suspect you accidentally updated Demitasse and that's why the tests are failing...

justingrant:
> Seems good, but I suspect you accidentally updated Demitasse and that's why the tests are failing...

Oops, you beat me to it.  Was just fixing this. Original checkin had a typo when excluding demitasse from updates.

PR #72, opened 2021-10-02
Fix some param reassignments, and enable lint. 

12wrigja:
Fixed for PlainDate, PlainDateTime, PlainMonthDay, PlainTime, and  PlainYearMonth so as not to clobber @justingrant's work on fixing up types for the other files (not sure how cleanly the PRs would merge).


.eslintrc.yml
```diff
@@ -88,3 +91,16 @@ overrides:
       - test/validStrings.mjs
     rules:
       no-console: off
+  - files:
+      - test/*.mjs
+      - test/**/*
+      # TODO Fix these files!
+      - lib/calendar.ts
```
justingrant, line 98: I've only been working on ecmascript.ts (with a few minor changes to calendar.ts that would be easy to rebase on top of any work you do), so all other files are fair game for this PR. No need to limit to just Plain types.

lib/plaindate.ts
```diff
@@ -23,10 +23,10 @@ const DISALLOWED_UNITS = ['hour', 'minute', 'second', 'millisecond', 'microsecon
 
 export class PlainDate implements Temporal.PlainDate {
   constructor(isoYear, isoMonth, isoDay, calendar: Temporal.CalendarProtocol | string = ES.GetISO8601Calendar()) {
-    isoYear = ES.ToIntegerThrowOnInfinity(isoYear);
-    isoMonth = ES.ToIntegerThrowOnInfinity(isoMonth);
-    isoDay = ES.ToIntegerThrowOnInfinity(isoDay);
-    calendar = ES.ToTemporalCalendar(calendar);
+    const isoYearConverted = ES.ToIntegerThrowOnInfinity(isoYear);
```
justingrant, line 26: How'd you feel abut reversing the renaming, so instead of creating new identifiers inside the function, instead we'd rename the parameters?  Like this:

``` js
constructor(isoYearParam, isoMonthParam, isoDayParam, calendarParam: Temporal.CalendarProtocol | string = ES.GetISO8601Calendar()) {
  const isoYear = ES.ToIntegerThrowOnInfinity(isoYearParam);
```

A few advantages:
* Dramatically reduces divergence from proposal-temporal code, which will make future porting of PRs easier.
* No risk of forgetting to replace some uses of the old identifier lower down in functions
* Coalesces TS-related changes to the top of most methods
* Allows using a consistent naming convention for all renamed params (e.g. xxxParam or whatever) instead of different naming conventions depending on usage of the value.  I think this will make the code easier to read later because it's always clear when you're dealing with the "original" or "unsafe" value. 
* Might make it easier to use regex search-n-replace to make the changes (although might not-- haven't thought this through to know for sure)

What do you think?

BTW, one downside of doing this would be that if we ever switched to auto-generating index.d.ts from the source code, then the parameter names would be wrong. But IMHO, given that Temporal types are defined by the spec anyway (not our code), I'm not sure we'd actually want to auto-generate index.d.ts anyway. And I guess if we used a consistent naming convention, then it'd be trivial to fix up the names with a post-build step.

justingrant (changes requested):
This is great!  Such a great idea to do this. Every file except ecmascript.ts should be fair game for this change, and I'll plan to do the same for ecmascript.ts.

Do you think we should make the same change to the proposal-temporal code to keep them in sync? Or best to make it only here? @ptomato do you have an opinion?

lib/plaindate.ts
ptomato: This change would show up in the toString output of functions, e.g. in `Temporal.PlainDate.prototype.add.toString()`. I don't know if IDEs and such use the toString output to get the parameter names to display in popups or whatnot, but if they do, that might be another downside.

ptomato:
I think an advantage of porting this to proposal-temporal would be reduced divergence for porting future changes from proposal-temporal to here, but on the other hand it would increase the divergence between the proposal-temporal code and the specification. So I feel neutral-to-negative about that.

lib/plaindate.ts
justingrant: > I don't know if IDEs and such use the toString output to get the parameter names to display in popups or whatnot, but if they do, that might be another downside.

Nope, AFAIK the TS types in index.d.ts is where IDE help UI comes from. 

> This change would show up in the toString output of functions, e.g. in `Temporal.PlainDate.prototype.add.toString()`.

Hmm, interesting. Thanks for pointing this out. If the names are recognizable (just with `param` at the end) is this a problem?

ptomato: Likely not a problem, then.

12wrigja:
(Not a Temporal Champion but) I'm in agreement with Philip here - the spec isn't written this way (though, frankly that seems a bit odd to me too, though I suppose for JS itself it doesn't really care about reassignment in the way that developers do for e.g. readability of the code), but given that it is I think it makes sense to leave the proposal repo as-is.

lib/plaindate.ts
12wrigja: So long as the names that appear in the toString aren't being expected anywhere (Test262 tests would be the only place I'd maybe expect this), then I'm ok with swapping the order.

> given that Temporal types are defined by the spec anyway (not our code), I'm not sure we'd actually want to auto-generate index.d.ts

presumably at some point these types might end up in the TS standard library, so we'd use those?

Something I've been considering: it would be nice to have the type system enforcing that these coercion functions are used. One way this is sometimes done is by typing the params as `unknown`, but I don't think that would work very well in our case because it would just make calling those functions more error-prone (no compiler support on the calling side for "you've clearly passed the wrong type of thing"). Another possibility might be to add nominal brands to these coercion functions (and the corresponding internal types), something like:

```
declare const temporalInternalConverted: unique symbol;
type TemporalInternalCalendar: Temporal.CalendarLike & {
   [temporalInternalConverted]: never;
}

export function ToTemporalCalendar(calendarLike: Temporal.CalendarLike | Temporal.Calendar):  TemporalInternalCalendar {
  // Have to force this to be the right type, unfortunately.
  return something as unknown as TemporalInternalCalendar;
}

export function SomeInternalEcmascriptFunction(calendar: TemporalInternalCalendar) ...
```

In theory, tsc should now complain if you accidentally directly use the function parameter and should be passing it through `ToTemporalCalendar` first.

.eslintrc.yml
12wrigja: Done.

lib/plaindate.ts
12wrigja: Seemingly 262 tests don't mind this at all, so I adopted this convention. I'll look more into the branding stuff in another PR, after we land Justin's changes to ecmascript.

PR #73, opened 2021-10-02
Make runtime type checks TS-friendly

justingrant:
This PR makes minor refactors to runtime type checks to make that validation more TS-friendly:
* Replace use of `ES.Type` for validating `Object` with a new type guard `IsObject` function. The function is overloaded so unknown values are typed as `Record`, while known types like `DurationLike | string` are simply stripped of primitive types while leaving the object types. As a nice side effect, this should reduce bundle size a little bit.
* For the other uses of ES.Type (there were only <5 of them), converts to simple `typeof` checks which provide the same type-narrowing benefit as type guard functions.
* Removes the ES.Type function.
* Adds type annotations to make all IsTemporalXxx into TS type guards.

@ptomato what do you think about whether I should port the ES.Type=>{IsObject, typeof} changes over to proposal-temporal to reduce divergence in the code?

### DIFF
--- a/lib/calendar.ts
+++ b/lib/calendar.ts
@@ -62,19 +62,19 @@ export class Calendar implements Temporal.Calendar {
   }
   dateFromFields(fields, options = undefined) {
     if (!ES.IsTemporalCalendar(this)) throw new TypeError('invalid receiver');
-    if (ES.Type(fields) !== 'Object') throw new TypeError('invalid fields');
+    if (!ES.IsObject(fields)) throw new TypeError('invalid fields');
     options = ES.GetOptionsObject(options);
     return impl[GetSlot(this, CALENDAR_ID)].dateFromFields(fields, options, this);
   }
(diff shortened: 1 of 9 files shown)


justingrant:
Heads up @12wrigja - This is the only non-trivial change I've made so far outside ecmascript.ts, so I figured it made sense to split it out. Also, these changes may reduce TS error noise for you regardless of what file you're working on, so may make sense to rebase your changes over this one. It's also OK to rebase the other direction later; whichever is easiest for you.

ptomato:
This sounds like a good change to make. But I don't think it should be ported to proposal-temporal, as we already intentionally changed it to the current state. (Partly to make the polyfill code closer to the specification, and partly to avoid inadvertent bugs where "Type(_v_) is Object" in the specification was implemented as `typeof v === 'object'` instead of `typeof v === 'object' || typeof v === 'function'`.)

12wrigja (approved):

PR #74, opened 2021-10-04
DO NOT MERGE: Add types to ecmascript.ts

justingrant:
Interim commit to do code-sharing with @12wrigja . Will push later with real commit messages.

Quick summary:
* Fancier types for slots and intrinsics
* ecmascript.ts now has types throughout and (almost) compiles with `noImplicitAny`
* Quick pass at fixing BigInt errors in a few places to enable compiles to not fail. Will be replace by James's work on JSBI.
* Found a few index.d.ts bugs in the public types.
* Some initial work in calendar.ts, but I paused that while James does work on that file.
* A new internal types file for non-public shared types.

I'll probably split out some of above into separate PRs to make them easier to review.


justingrant:
Not sure why Test262 shows as failed given that the output doesn't report any failures. Trying to re-run.

lib/ecmascript.ts
```diff
@@ -1080,8 +1235,20 @@ export function InterpretTemporalDateTimeFields(calendar, fields, options) {
   return { year, month, day, hour, minute, second, millisecond, microsecond, nanosecond };
 }
 
-export function ToTemporalDateTime(item, options = ObjectCreate(null)) {
-  let year, month, day, hour, minute, second, millisecond, microsecond, nanosecond, calendar;
+export function ToTemporalDateTime(
+  item: Parameters<typeof Temporal.PlainDateTime['from']>[0],
+  options = ObjectCreate(null)
```
12wrigja, line 1240: Does this need explicitly typing as `object` or something else? Last I recall, ObjectCreate returned `any`.

lib/ecmascript.ts
```diff
@@ -1197,9 +1367,11 @@ export function ToTemporalMonthDay(item, options = ObjectCreate(null)) {
   }
 
   ToTemporalOverflow(options); // validate and ignore
-  let { month, day, referenceISOYear, calendar } = ParseTemporalMonthDayString(ToString(item));
-  if (calendar === undefined) calendar = GetISO8601Calendar();
-  calendar = ToTemporalCalendar(calendar);
+  let { month, day, referenceISOYear, calendar: maybeStringCalendar } = ParseTemporalMonthDayString(ToString(item));
+  // TODO: should this be a ternary?
```
12wrigja, line 1371: as

`const calendar = maybeStringCalendar ? ToTemporalCalendar(maybeStringCalendar) : GetISO8601Calendar();`

Seems reasonable to me, though should empty string be treated the same as undefined? Or should it be an explicit undefined check?

lib/ecmascript.ts
```diff
@@ -1259,19 +1435,24 @@ export function ToTemporalTime(item, overflow = 'constrain') {
   return new TemporalPlainTime(hour, minute, second, millisecond, microsecond, nanosecond);
 }
 
-export function ToTemporalYearMonth(item, options = ObjectCreate(null)) {
+export function ToTemporalYearMonth(
+  item: Temporal.PlainYearMonth | Temporal.PlainYearMonthLike | string,
+  options = ObjectCreate(null)
```
12wrigja, line 1440: Same question here r.e. the type of options.

lib/ecmascript.ts
```diff
@@ -1573,19 +1824,26 @@ export function GetISO8601Calendar() {
   return new TemporalCalendar('iso8601');
 }
 
-export function CalendarFields(calendar, fieldNames) {
+// TODO: should (can?) we make this generic so the field names are checked
```
12wrigja, line 1827: This seems interesting :) We should try this in another PR.

lib/ecmascript.ts
```diff
@@ -1595,7 +1853,14 @@ export function CalendarMergeFields(calendar, fields, additionalFields) {
   return result;
 }
 
-export function CalendarDateAdd(calendar, date, duration, options, dateAdd?: any) {
+export function CalendarDateAdd(
```
12wrigja, line 1856: Nit: Mind adding a comment here + elsewhere that using the `Parameters` pattern here for typing arguments is especially important given that user-defined calendars are a thing and we (presumably) partially document our expectations within the .d.ts.

lib/ecmascript.ts
```diff
@@ -1713,21 +2028,23 @@ export function ToTemporalCalendar(calendarLike) {
   return new TemporalCalendar(calendar);
 }
 
-function GetTemporalCalendarWithISODefault(item) {
+function GetTemporalCalendarWithISODefault(
+  item: Temporal.CalendarProtocol | { calendar?: Temporal.PlainDateLike['calendar'] } | undefined
```
12wrigja, line 2032: Why isn't the second thing here just `Temporal.Calendar` ? Can this also be a string identifier?

lib/internaltypes.d.ts
```diff
@@ -0,0 +1,89 @@
+import { Temporal } from '..';
+export type BuiltinCalendarId =
```
12wrigja, line 2: This is now the source of truth for what the registered calendars are, right? It looks like the `impl` within calendar.ts will now (correctly) reject assigning any key to it that isn't listed here?

lib/intrinsicclass.ts
```diff
@@ -1,8 +1,24 @@
...
+import bigInt from 'big-integer';
+import { Temporal } from '..';
 
 import { DEBUG } from './debug';
 
-type TemporalIntrinsics = Omit<typeof Temporal, 'Now'>;
+type OmitConstructor<T> = { [P in keyof T as T[P] extends new (...args: any[]) => any ? P : never]: T[P] };
+
+type TemporalIntrinsics = Omit<typeof Temporal, 'Now' | 'Instant' | 'ZonedDateTime'> & {
+  Instant: OmitConstructor<Temporal.Instant> &
```
12wrigja, line 9: Why is this needed?

lib/intrinsicclass.ts
```diff
@@ -96,15 +120,18 @@ type IntrinsicDefinitionKeys =
...
-);
-export function DefineIntrinsic<KeyT extends keyof StandaloneIntrinsics>(name: KeyT, value: StandaloneIntrinsics[KeyT]);
-export function DefineIntrinsic<KeyT>(name: KeyT, value: never);
+): void;
+export function DefineIntrinsic<KeyT extends keyof StandaloneIntrinsics>(
+  name: KeyT,
+  value: StandaloneIntrinsics[KeyT]
+): void;
+export function DefineIntrinsic<KeyT>(name: KeyT, value: never): void;
 export function DefineIntrinsic<KeyT extends IntrinsicDefinitionKeys>(name: KeyT, value: unknown) {
```
12wrigja, line 133: Nit: Add the void return type here too.

lib/slots.ts
```diff
@@ -39,21 +43,249 @@ export const NANOSECONDS = 'slot-nanoseconds';
...
+export function CreateSlots(container: AnyTemporalType): void {
   slots.set(container, Object.create(null));
 }
-function GetSlots(container) {
+
+function GetSlots<T extends AnyTemporalType>(container: T) {
   return slots.get(container);
 }
-export function HasSlot(container, ...ids) {
+// TODO: is there a better way than 9 overloads to make HasSlot into a type
```
12wrigja, line 161: My understanding is that this can be done by:
  - wrapping each of the elements in the variadic args array type into a tuple (`[keyType1, keyType2]` becomes `[[keyType1], [keyType2]]`)
  - convert this into an intersection type (following https://fettblog.eu/typescript-union-to-intersection/) (`[[keyType1], [keyType2]]` becomes `[keyType1]|[keyType2]`)
  - unwrap the tuples  (`[keyType1]|[keyType2]` becomes `keyType1|keyType2`)
  
  Though I can try and sort this out in a separate PR if you don't want to take a shot at that.

lib/ecmascript.ts
```diff
@@ -249,7 +251,8 @@ export function IsTemporalMonthDay(item: unknown): item is Temporal.PlainMonthDa
 export function IsTemporalZonedDateTime(item: unknown): item is Temporal.ZonedDateTime {
   return HasSlot(item, EPOCHNANOSECONDS, TIME_ZONE, CALENDAR);
 }
-function TemporalTimeZoneFromString(stringIdent) {
+function TemporalTimeZoneFromString(stringIdent: string) {
+  // TODO: why aren't these three variables destructured to include `undefined` as possible types?
```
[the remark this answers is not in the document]
justingrant: Oh, yeah that probably explains it. My main TS project has strictNullChecks=true so I've never run into this. Your explanation makes sense. I tried turning on `strictNullChecks` and got 400 compiler errors so maybe we'll save that for a later PR. 😄  

I'll leave this comment here as a reminder for then.

lib/ecmascript.ts
```diff
@@ -797,9 +911,9 @@ export function ToRelativeTemporalObject(options) {
       fields,
       dateOptions
     ));
-    offset = relativeTo.offset;
+    offset = (relativeTo as Temporal.ZonedDateTimeLike).offset;
```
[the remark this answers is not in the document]
justingrant: This is a different issue. Only ZonedDateTime has an `offset` property. So this assertion is needed to check for the presence of this property. I could use a type guard but IMHO there's a fairly high bar to diverge the runtime code from proposal-temporal and IMHO this doesn't meet the bar.

lib/ecmascript.ts
justingrant: All the To* functions get explicit parameter types in a later commit I'm working on, so this problem will go away. I'll resolve for now.

lib/ecmascript.ts
justingrant: For now I'm generally leaving these cases as-is to align with proposal-temporal. I'd be open to cleaning them up when there's less change coming from the other repo, maybe in a few months? Sound OK?

lib/ecmascript.ts
justingrant: Same answer. :-)

lib/ecmascript.ts
justingrant: 👍

lib/ecmascript.ts
justingrant: These are getting cleaned up in a later commit. The new code will look like this: 
```ts
export function CalendarDateAdd(
  calendar: Temporal.CalendarProtocol,
  date: CalendarProtocolParams['dateAdd'][0],
  duration: CalendarProtocolParams['dateAdd'][1],
  options: CalendarProtocolParams['dateAdd'][2],
  dateAddParam?: Temporal.CalendarProtocol['dateAdd']
) {
```

Better?

I don't think it makes sense to add a comment everywhere these are used because we'd end up with 100s of comments. Seems better to just document the *Params types in one place. Let's discuss when we review the later commit where this changes.

lib/ecmascript.ts
justingrant: This can be a property bag (one of the Temporal.XxxLike types) or (maybe, not sure) a Temporal object that has a calendar. This `{ calendar?: Temporal.PlainDateLike['calendar'] }` type covers all those cases. Look at references to this function and should be clearer how it's used. LMK if still unclear.

lib/internaltypes.d.ts
justingrant: Yes.

lib/intrinsicclass.ts
justingrant: The constructor in the public types only accepts `bigint`. The implementation constructor accepts `bigInt.BigInteger | bigint`. This hacky type was one way to support this dichotomy. My suggestion would be to not spend a lot of time worrying about it now-- let's solve it "the right way" (whatever that is!) with your JSBI PR.

lib/intrinsicclass.ts
justingrant: oops!  fixed.

lib/slots.ts
justingrant: Yeah, I'd like to land this big PR quickly. If you want to take a shot in a later PR that'd be awesome!

lib/calendar.ts
```diff
@@ -60,27 +132,36 @@ export class Calendar implements Temporal.Calendar {
   get id() {
     return ES.ToString(this);
   }
-  dateFromFields(fields, optionsParam = undefined) {
+  dateFromFields(
+    fields: { year: number | undefined; month: number | undefined; monthCode: string | undefined; day: number },
```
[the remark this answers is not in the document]
justingrant: I'll clean this up in a later PR

lib/ecmascript.ts
```diff
@@ -513,7 +545,23 @@ export function RegulateISOYearMonth(year, month, overflow) {
   return { year, month };
 }
 
-function DurationHandleFractions(fHours, minutes, fMinutes, seconds, milliseconds, microseconds, nanoseconds) {
+function DurationHandleFractions(
+  fHoursParam: number,
```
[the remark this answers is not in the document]
12wrigja: Ah - that's fine by me then.

lib/calendar.ts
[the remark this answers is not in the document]
12wrigja: Sg!

lib/ecmascript.ts
12wrigja: Ah gotcha.

12wrigja:
Seems like the GH UI is doing something weird - I can't seem to respond to any of the open conversations unless I click "into" them, and even then the comments don't seem to propagate back.

I have no objections to merging this as-is and following up on my comments later, especially given some of them might be fixed by compiler flag changes (e.g. `strictNullChecks`).

Issue #76, opened 2021-10-06
Unable to run test262 test suite locally


12wrigja:
This could be some weird issue with my local setup, or some issue with Git submodules, but I can no longer run the test262 tests locally.

Steps followed:
```
cd ~/Documents && git clone git@github.com:js-temporal/temporal-polyfill.git temporal-polyfill-tmp
cd temporal-polyfill-tmp
npm install
npm run test262
```
yields
```
jameswr:~/Documents/temporal-polyfill-tmp$ npm run test262

> @js-temporal/polyfill@0.2.0 test262
> TEST262=1 npm run build && ./test/test262.sh


> @js-temporal/polyfill@0.2.0 build
> rm -rf dist/* && rollup -c rollup.config.js


lib/init.ts → dist/script.js...
(!) Circular dependency
lib/ecmascript.ts -> lib/calendar.ts -> lib/ecmascript.ts
created dist/script.js in 4.2s
./test/test262.sh: line 14: cd: ./test/../test262/test/: No such file or directory
jameswr:~/Documents/temporal-polyfill-tmp$ 
```

@ptomato any ideas why this might happen? The tests seem to work fine on CI, and as far as I can tell that runs basically all the same commands as I did.

ptomato:
When you clone a fresh copy of the repo, you have to do `git submodule init && git submodule update` to clone the submodule as well, or add `--recurse-submodules=test262` to the `git clone` command. (I never understood why Git didn't make this the default.)

12wrigja:
Ah gotcha. One more question: do you have the temporal repo's node_modules as part of your path? Now when I run the test262 script, it complains that it can't find the test262-harness binary. If I use `npx test262-harness` that works, and the CI is clearly fine, so I'm wondering what the difference here could be.

I'll send a PR that adds a warning for the submodule initialization shortly.

ptomato:
I guess I was only running the script via the `npm run test262` command.  `npm run` puts the `node_modules/.bin` in the path temporarily.

12wrigja:
I just came to the same conclusion myself :) Oops!

12wrigja:
To fix it you should rebase your commit(s) and explicitly remove the submodule from the staging area before committing again, I think. If you scroll all the way to the bottom of the files changes in your PR you can see it was added as a submodule.

PR #77, opened 2021-10-06
Use JSBI instead of big-integer

12wrigja:
JSBI has the following advantages over big-integer:
 - it has better type-safety and mirrors native BigInt behavior (cannot mix native BigInt and number)
 - has a babel plugin (not configured in this PR) that can transpile away JSBI entirely to use native BigInt instead, for code-size improvements. We should consider shipping a separate package (maybe this should be the default) using this that provides all the Temporal features and uses native BigInt, as I'd guess this will be more common than users wanting to use this on platforms without native BigInt.

JSBI has some disadvantages:
 - the JSBI.BigInt factory function does NOT support automatically converting native BigInt to a JSBI instance at runtime. Instead, in places where the type might be a native BigInt, we use `ES.ToBigInt` to convert to a JSBI instance.
 - does not feature-detect platform availability of BigInts and return those instead of JSBI instances, in the event that a user doesn't use the Babel plugin. We do this ourselves instead with `ES.ToBigIntExternal`, which guarantees it returns a native BigInt when available, and something with a `ToString` method that returns the base 10 value of the bigint as a string, which I think is likely the most straightforward way for interop between various bigint compatibility libraries. It would be great if there was a way to properly encode this in typings, and this might be possible with multiple entry points and [`typesVersions`](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#version-selection-with-typesversions).

### DIFF
--- a/lib/ecmascript.ts
+++ b/lib/ecmascript.ts
@@ -17,7 +17,7 @@ const ObjectIs = Object.is;
 const ReflectApply = Reflect.apply;
 
 import { DEBUG } from './debug';
-import bigInt from 'big-integer';
+import JSBI from 'jsbi';
 
 import type { Temporal } from '..';
 import type {
@@ -74,13 +74,21 @@ import {
   NANOSECONDS
 } from './slots';
 import { IsBuiltinCalendar } from './calendar';
+
+export const ZERO = JSBI.BigInt(0);
+const ONE = JSBI.BigInt(1);
+const SIXTY = JSBI.BigInt(60);
+export const THOUSAND = JSBI.BigInt(1e3);
+export const MILLION = JSBI.BigInt(1e6);
+export const BILLION = JSBI.BigInt(1e9);
+const NEGATIVE_ONE = JSBI.BigInt(-1);
 const DAY_SECONDS = 86400;
-const DAY_NANOS = bigInt(DAY_SECONDS).multiply(1e9);
-const NS_MIN = bigInt(-DAY_SECONDS).multiply(1e17);
-const NS_MAX = bigInt(DAY_SECONDS).multiply(1e17);
+const DAY_NANOS = JSBI.multiply(JSBI.BigInt(DAY_SECONDS), BILLION);
+const NS_MIN = JSBI.multiply(JSBI.BigInt(-86400), JSBI.BigInt(1e17));
+const NS_MAX = JSBI.multiply(JSBI.BigInt(86400), JSBI.BigInt(1e17));
 const YEAR_MIN = -271821;
 const YEAR_MAX = 275760;
-const BEFORE_FIRST_DST = bigInt(-388152).multiply(1e13); // 1847-01-01T00:00:00Z
+const BEFORE_FIRST_DST = JSBI.multiply(JSBI.BigInt(-388152), JSBI.BigInt(1e13)); // 1847-01-01T00:00:00Z
 
 function IsInteger(value: unknown): value is number {
   if (typeof value !== 'number' || !NumberIsFinite(value)) return false;
@@ -150,6 +158,16 @@ export function ToIntegerWithoutRounding(valueParam: unknown): number {
     throw new RangeError(`unsupported fractional value ${value}`);
   }
   return ToInteger(value); // ℝ(value) in spec text; converts -0 to 0
+}
+function divmod(x: JSBI, y: JSBI): { quotient: JSBI; remainder: JSBI } {
+  const quotient = JSBI.divide(x, y);
+  const remainder = JSBI.remainder(x, y);
+  return { quotient, remainder };
+}
+
+function abs(x: JSBI): JSBI {
+  if (JSBI.lessThan(x, ZERO)) return JSBI.multiply(x, NEGATIVE_ONE);
+  return x;
 }
 
 const BUILTIN_CASTS = new Map<PrimitivePropertyNames, (v: unknown) => string | number>([
@@ -498,7 +516,7 @@ function ParseTemporalInstant(isoString: string) {
   if (epochNs === null) throw new RangeError('DateTime outside of supported range');
   if (!z && !offset) throw new RangeError('Temporal.Instant requires a time zone offset');
   const offsetNs = z ? 0 : ParseOffsetString(offset);
-  return epochNs.subtract(offsetNs);
+  return JSBI.subtract(epochNs, JSBI.BigInt(offsetNs));
 }
 
 export function RegulateISODate(
@@ -1550,14 +1568,16 @@ export function InterpretISODateTimeOffset(
     // Calculate the instant for the input's date/time and offset
     const epochNs = GetEpochFromISOParts(year, month, day, hour, minute, second, millisecond, microsecond, nanosecond);
     if (epochNs === null) throw new RangeError('ZonedDateTime outside of supported range');
-    return epochNs.minus(offsetNs);
+    return JSBI.subtract(epochNs, JSBI.BigInt(offsetNs));
   }
 
   // "prefer" or "reject"
   const possibleInstants = GetPossibleInstantsFor(timeZone, dt);
   for (const candidate of possibleInstants) {
     const candidateOffset = GetOffsetNanosecondsFor(timeZone, candidate);
-    const roundedCandidateOffset = RoundNumberToIncrement(bigInt(candidateOffset), 60e9, 'halfExpand').toJSNumber();
+    const roundedCandidateOffset = JSBI.toNumber(
+      RoundNumberToIncrement(JSBI.BigInt(candidateOffset), 60e9, 'halfExpand')
+    );
     if (candidateOffset === offsetNs || (matchMinute && roundedCandidateOffset === offsetNs)) {
       return GetSlot(candidate, EPOCHNANOSECONDS);
     }
@@ -1841,7 +1861,7 @@ export function CreateTemporalYearMonth(
 
 export function CreateTemporalZonedDateTimeSlots(
   result: Temporal.ZonedDateTime,
-  epochNanoseconds: bigInt.BigInteger,
+  epochNanoseconds: JSBI,
   timeZone: Temporal.TimeZoneProtocol,
   calendar: Temporal.CalendarProtocol
 ) {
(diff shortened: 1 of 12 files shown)


12wrigja:
I'm only drafting this PR so that I can get the test262 tests run against it (as I can't seem to run those locally - see #76). I'm down to 2 failing Demitasse tests, both to do with rounding triggering overflow, which I'll hopefully fix tomorrow.

lib/ecmascript.ts
```diff
@@ -147,6 +148,19 @@ function ToIntegerNoFraction(valueParam: unknown): number {
   }
   return value;
 }
+function divmod(x: JSBI, y: JSBI): { quotient: JSBI; remainder: JSBI } {
```
ptomato, line 151: I don't have a strong opinion one way or the other, but it may be good to just inline this where it's used? I'm not sure if the quotient and remainder are both used in every instance.

lib/ecmascript.ts
```diff
@@ -4826,91 +4858,111 @@ export function RoundDuration(
...
-        .plus(bigInt(minutes).multiply(60e9))
-        .plus(bigInt(seconds).multiply(1e9))
-        .plus(bigInt(milliseconds).multiply(1e6))
-        .plus(bigInt(microseconds).multiply(1e3))
-        .plus(nanoseconds);
-      total = nanoseconds.toJSNumber() / divisor;
-      const rounded = RoundNumberToIncrement(nanoseconds, divisor * increment, roundingMode);
-      hours = rounded.divide(divisor).toJSNumber();
-      nanoseconds = bigInt.zero;
+      const hoursAsNanos = JSBI.multiply(JSBI.BigInt(hours), JSBI.BigInt(3600e9));
```
justingrant, line 4892: Personally, I find this rewrite harder to read than the code it's replacing, and also pretty divergent from proposal-temporal.  What about making `allNanoseconds` a mutable value instead?  

Then you could approximate the current chained flow, e.g. 

```js
let allNanoseconds = JSBI.multiply(JSBI.BigInt(hours), JSBI.BigInt(3600e9));
allNanoseconds = JSBI.add(allNanoseconds, JSBI.multiply(JSBI.BigInt(minutes), JSBI.BigInt(60e9)));
. . .
allNanoseconds = JSBI.add(allNanoseconds, nanoseconds);
```

lib/ecmascript.ts
```diff
@@ -4826,91 +4858,111 @@ export function RoundDuration(
...
     case 'minute': {
       const divisor = 60e9;
-      nanoseconds = bigInt(minutes)
-        .multiply(60e9)
-        .plus(bigInt(seconds).multiply(1e9))
-        .plus(bigInt(milliseconds).multiply(1e6))
-        .plus(bigInt(microseconds).multiply(1e3))
-        .plus(nanoseconds);
-      total = nanoseconds.toJSNumber() / divisor;
+      const minutesAsNanos = JSBI.multiply(JSBI.BigInt(minutes), JSBI.BigInt(60e9));
```
justingrant, line 4913: Same comment as above.

lib/ecmascript.ts
```diff
@@ -4826,91 +4858,111 @@ export function RoundDuration(
...
     }
     case 'second': {
       const divisor = 1e9;
-      nanoseconds = bigInt(seconds)
-        .multiply(1e9)
-        .plus(bigInt(milliseconds).multiply(1e6))
-        .plus(bigInt(microseconds).multiply(1e3))
-        .plus(nanoseconds);
-      total = nanoseconds.toJSNumber() / divisor;
+      const secondsAsNanos = JSBI.multiply(JSBI.BigInt(seconds), JSBI.BigInt(1e9));
```
justingrant, line 4930: same

lib/ecmascript.ts
```diff
@@ -4826,91 +4858,111 @@ export function RoundDuration(
...
+      seconds = JSBI.toNumber(JSBI.divide(rounded, JSBI.BigInt(divisor)));
+      nanoseconds = ZERO;
       milliseconds = microseconds = 0;
       break;
     }
     case 'millisecond': {
       const divisor = 1e6;
-      nanoseconds = bigInt(milliseconds).multiply(1e6).plus(bigInt(microseconds).multiply(1e3)).plus(nanoseconds);
-      total = nanoseconds.toJSNumber() / divisor;
+      const millisecondsAsNanos = JSBI.multiply(JSBI.BigInt(milliseconds), JSBI.BigInt(1e6));
```
justingrant, line 4943: same

lib/ecmascript.ts
```diff
@@ -4826,91 +4858,111 @@ export function RoundDuration(
...
+      milliseconds = JSBI.toNumber(JSBI.divide(rounded, JSBI.BigInt(divisor)));
+      nanoseconds = ZERO;
       microseconds = 0;
       break;
     }
     case 'microsecond': {
       const divisor = 1e3;
-      nanoseconds = bigInt(microseconds).multiply(1e3).plus(nanoseconds);
-      total = nanoseconds.toJSNumber() / divisor;
+      const microSecondsAsNanos = JSBI.multiply(JSBI.BigInt(microseconds), JSBI.BigInt(1e3));
```
justingrant, line 4955: same

lib/ecmascript.ts
12wrigja: I avoided inlining this because it makes the diffs much easier to read in some cases where there are several divmod operations happening sequentially.

lib/ecmascript.ts
```diff
@@ -2704,15 +2715,15 @@ export function GetEpochFromISOParts(
...
+  ns = JSBI.add(ns, JSBI.BigInt(nanosecond));
+  if (JSBI.lessThan(ns, NS_MIN) || JSBI.greaterThan(ns, NS_MAX)) return null;
   return ns;
 }
 
-function GetISOPartsFromEpoch(epochNanoseconds: bigInt.BigInteger) {
-  const { quotient, remainder } = bigInt(epochNanoseconds).divmod(1e6);
+function GetISOPartsFromEpoch(epochNanoseconds: JSBI) {
+  const { quotient, remainder } = divmod(epochNanoseconds, JSBI.BigInt(1e6));
   let epochMilliseconds = +quotient;
```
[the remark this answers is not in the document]
justingrant: Also, if + doesn't work for JSBI, then why didn't it cause tests to fail?  Is that Test262 being out of date, or something else?

lib/zoneddatetime.ts
```diff
@@ -606,8 +607,8 @@ export class ZonedDateTime implements Temporal.ZonedDateTime {
     const dtStart = new DateTime(GetSlot(dt, ISO_YEAR), GetSlot(dt, ISO_MONTH), GetSlot(dt, ISO_DAY), 0, 0, 0, 0, 0, 0);
     const instantStart = ES.BuiltinTimeZoneGetInstantFor(timeZone, dtStart, 'compatible');
     const endNs = ES.AddZonedDateTime(instantStart, timeZone, calendar, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0);
-    const dayLengthNs = endNs.subtract(GetSlot(instantStart, EPOCHNANOSECONDS));
-    if (dayLengthNs.isZero()) {
+    const dayLengthNs = JSBI.subtract(endNs, JSBI.BigInt(GetSlot(instantStart, EPOCHNANOSECONDS)));
+    if (JSBI.equal(dayLengthNs, ZERO)) {
```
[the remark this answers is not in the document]
12wrigja: I was treating zero specially simply because bigInteger has the `isZero` method, and so I started by quickly find/replacing this with a reference to the constant.

I agree we should dedupe the others, but I don't know how best to name the constants. THOUSAND, MILLION, <whatever 1e9 is called> ? Open to suggestions.

lib/ecmascript.ts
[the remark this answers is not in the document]
12wrigja: I need to correct myself above. `+JSBI.BigInt(10)` "does" work - it returns `10` (a `Number`). However, that isn't the same as the equivalent native BigInt operation: `+globalThis.BigInt(10)` throws `TypeError: VM112:1 Uncaught TypeError: Cannot convert a BigInt value to a number`. Technically, my PR to JSBI will fix that, though at the cost of diverging JSBI behavior from native BigInt (you can call `globalThis.BigInt(10).valueOf()` and that doesn't throw, but I don't know of any other way to intercept just the implicit conversion done by `+` and friends - maybe `Symbol.toPrimitive` could help here?). It's unclear whether that is desirable behavior or too large of a breaking change at this time for JSBI.

Where I got `NaN` from: I was testing this in DevTools and attempted to emulate it by running `+({})`, which does return NaN, but this is wrong given that JSBI instances do define toString, so the equivalent repro would be something like `+({toString(){return "10";}})`, which does return `10`.

Regardless, I used the patch to JSBI to track down where this was happening in this polyfill: here, and in `ToBigIntExternal` (from passing a JSBI instance to the native BigInt constructor). I've changed the implicit conversion here to use ToNumber for clarity, and adjusted the implementation of `ToBigIntExternal` to ensure that we don't rely on this implicit conversion behavior, instead preferring to explicitly convert to a JSBI and then to a native BigInt by using `toString(10)` as the intermediate value.

lib/ecmascript.ts
```diff
@@ -4826,91 +4858,111 @@ export function RoundDuration(
...
-        .plus(nanoseconds);
-      total = nanoseconds.toJSNumber() / divisor;
-      const rounded = RoundNumberToIncrement(nanoseconds, divisor * increment, roundingMode);
-      hours = rounded.divide(divisor).toJSNumber();
-      nanoseconds = bigInt.zero;
+      const hoursAsNanos = JSBI.multiply(JSBI.BigInt(hours), JSBI.BigInt(3600e9));
+      const minutesAsNanos = JSBI.multiply(JSBI.BigInt(minutes), JSBI.BigInt(60e9));
+      const secondsAsNanos = JSBI.multiply(JSBI.BigInt(seconds), JSBI.BigInt(1e9));
+      const millisecondsAsNanos = JSBI.multiply(JSBI.BigInt(milliseconds), JSBI.BigInt(1e6));
+      const microSecondsAsNanos = JSBI.multiply(JSBI.BigInt(microseconds), JSBI.BigInt(1e3));
```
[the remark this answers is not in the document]
12wrigja: This is now obsolete as I'm using your earlier suggestion instead.

lib/ecmascript.ts
12wrigja: I agree your suggestion looks a lot nicer. Done, both here and for the other cases.

12wrigja:
Now that we've released, I plan to merge this unless there's any other concerns?

ptomato:
Assuming nothing substantial changed in the meantime, no concerns from me!

justingrant:
Not from me either!

12wrigja:
This has actually been surprisingly smooth - I haven't had many problems rebasing it over the past several weeks.

PR #79, opened 2021-10-06
Fix local debugging of test262

12wrigja:
The (existing) local debugging setup for VSCode uses environment variables, not arguments.

### DIFF
--- a/test/test262.sh
+++ b/test/test262.sh
@@ -3,7 +3,7 @@
 # being manually executed.
 set -e
 
-TESTS=${@:-"**/*.js"}
+TESTS=${TESTS:-"*/Temporal/**/*.js"}
 TIMEOUT=${TIMEOUT:-10000}
 
 if [ "$(uname)" = 'Darwin' ]; then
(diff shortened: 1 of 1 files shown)


justingrant (approved):

test/test262.sh
```diff
@@ -3,7 +3,7 @@
 # being manually executed.
 set -e
 
-TESTS=${@:-"**/*.js"}
+TESTS=${TESTS:-"*/Temporal/**/*.js"}
```
ptomato, line 6: This makes it slightly more annoying to run a subset of tests at the command line - previously we could do something like `npm run test262 "Instant/**/*.js"`. After this change, we'd have to do `TESTS="*/Temporal/Instant/**/*.js" npm run test262` which is certainly not insurmountable, but it would be nice to be able to preserve the previous behaviour.

(IMO it'd be most important to preserve the ability to omit `*/Temporal` because otherwise it's easy to forget that there are tests in both `built-ins/Temporal` and `intl402/Temporal`.)

12wrigja: I will send another PR shortly to do this.

PR #80, opened 2021-10-07
Save String constructor in ecmascript.ts at module load time

justingrant:
This PR aligns with the pattern we use for the Number constructor and other ES built-in objects, where the object is saved when the ecmascript.ts module is imported and that saved reference is used in place of `String()` elsewhere.

My only concern with this PR is whether it will impact perf in the case where the literal text `"String"` is special-cased into a fast path by JS engines. I'll check the test runtime before and after this change to see if perf looks like a problem.

UPDATE: I didn't see any huge perf impact.

### DIFF
--- a/lib/ecmascript.ts
+++ b/lib/ecmascript.ts
@@ -9,6 +9,7 @@ const MathTrunc = Math.trunc;
 const NumberIsNaN = Number.isNaN;
 const NumberIsFinite = Number.isFinite;
 const NumberCtor = Number;
+const StringCtor = String;
 const NumberMaxSafeInteger = Number.MAX_SAFE_INTEGER;
 const ObjectCreate = Object.create;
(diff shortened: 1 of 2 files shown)


12wrigja:
If I checkout your branch, and run `git reset --soft justin/main` to see what your working directory looked like before you committed, I see this:
```
HEAD detached from justin/string-ctor
Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
        modified:   lib/ecmascript.ts
        modified:   lib/timezone.ts
        new file:   test/tc39
```

You need to make sure that `test/tc39` doesn't show up. I removed it from the staging area by running `git reset test/tc39`, and then to be sure I deleted it from the filesystem entirely with `rm -rf test/tc39/`.

ptomato:
Yes, I think that would be correct.

PR #81, opened 2021-10-11
Add types to all files and enable `noImplicitAny`

justingrant:
This PR continues where #74 left off by adding parameter and return types to the rest of (non-ecmascript.ts) Temporal source files.

Also, it turns on `noImplicitAny`.  Except for calendar.ts, this PR removes `any` (implicit or otherwise) from the polyfill, with only a few exceptions for truly generic code or testing propeties on `globalThis`.

For calendar.ts, `any` is removed from the top level `Calendar` implementation and some of the `impl[]` code, but fully declaring all types in that file was more work than I wanted to take on in this PR. So I did an initial pass on obvious types and assigned a lot of `any` for the rest. Will definitely need a followup PR to improve calendar.ts. @12wrigja has discussed the possibility of replacing the old-school inheritance using Object.apply with a modern `class`-based implementation. If we're going to do that, then it may make sense to fix the types then instead of on top of the current solution. 

Aside from calendar.ts, there's still lots of TS work remaining (e.g. `strictNullChecks`, `strictPropertyInitialization`, and cleaning up calendar.ts), but at least we'll now have valid types on all functions.

I didn't find any runtime errors as a result of this work (good!) but I did find quite a few problems with the public types in index.d.ts, and this PR fixes those too. Once the dust settles and the changes are reviewed, I can port a change back to proposal-temporal to keep all public types in sync.


lib/intl.ts
```diff
@@ -45,36 +55,79 @@ const ObjectAssign = Object.assign;
...
 
-export function DateTimeFormat(
-  this: Intl.DateTimeFormat,
-  locale = undefined,
-  optionsParam: Partial<Intl.DateTimeFormatOptions> = {}
-): void {
-  if (!(this instanceof DateTimeFormat)) return new DateTimeFormat(locale, optionsParam);
+type DateTimeFormatImpl = Intl.DateTimeFormat & PrivateProps;
+
+function DateTimeFormatImpl(
```
justingrant, line 115: Here's the story behind renaming this function: I wanted to only export this function as an Intl.DateTimeFormat (from the public types) so that other modules wouldn't know about its implementation details. This is what we do, for example, with other Temporal types where callers can `import { Temporal } from '..'`.  (BTW, this works in tests but we should validate that it will also work in real apps!)

But the module resolver used in Demitasse tests (resolve.source.mjs) was special-cased for importing Temporal only. `import { Temporal, Intl } from '..'` caused Demitasse tests to fail. 

But I still wanted other-module callers to see this function as a plain `Intl.DateTimeFormat` (from public types) without seeing its private properties. I couldn't figure out how to do this when it was named `DateTimeFormat` without getting a TS error when attempting to change its type to the public type. So I renamed the function, which required a runtime code change to update the `name` property.

Feel free to suggest a better way to do it!

lib/intl.ts
```diff
@@ -31,7 +41,7 @@ const CAL_ID = Symbol('calendar-id');
 const LOCALE = Symbol('locale');
 const OPTIONS = Symbol('options');
 
-const descriptor = (value) => {
+const descriptor = <T extends (...args: any[]) => any>(value: T) => {
```
12wrigja, line 44: Could these be unknown instead of any?

lib/intl.ts
```diff
@@ -113,10 +166,21 @@ export function DateTimeFormat(
   this[DATETIME] = datetimeAmend;
   this[ZONED] = zonedDateTimeAmend;
   this[INST] = instantAmend;
+  return undefined; // TODO: I couldn't satisfy TS without adding this. Is there another way?
```
12wrigja, line 169: This is a bit worrying... is the intent that DateTimeFormatImpl be a private impl of the DateTimeFormat ctor?

lib/intl.ts
```diff
@@ -133,56 +197,95 @@ if ('formatRangeToParts' in IntlDateTimeFormat.prototype) {
...
+  // Existing Intl isn't typed to accept Temporal-specific options, but will not
+  // break at runtime if we pass them. Also, the lib types for resolved options
+  // are less restrictive than the types for options. For example, `weekday` is
+  // `'long' | 'short' | 'narrow'` in options but `string` in resolved options.
+  // TODO: investigate why, and file an issue against TS if it's a bug.
+  return new IntlDateTimeFormat(options.locale, { ...(options as globalThis.Intl.DateTimeFormatOptions), timeZone });
 }
 
-function format(this: typeof DateTimeFormat, datetime, ...rest) {
+// TODO: investigate why there's a rest parameter here. Does this function really need to accept extra params?
```
12wrigja, line 221: Maybe so that in the future extra params can be passed to format without us having to specify them? Though it clearly isn't wired up into formatRange where it probably should... For another PR I think :)

lib/intl.ts
```diff
@@ -214,14 +324,16 @@ function amend(optionsParam = {}, amended = {}) {
     'timeZoneName',
     'dateStyle',
     'timeStyle'
-  ]) {
-    options[opt] = opt in amended ? amended[opt] : options[opt];
-    if (options[opt] === false || options[opt] === undefined) delete options[opt];
+  ] as const) {
+    // TODO: can this be typed more cleanly?
+    type OptionMaybeFalse = typeof options[typeof opt] | false;
```
12wrigja, line 329: Hopefully there aren't any boolean-typed options added to Intl.DateTimeFormat that default to true... :|

lib/intl.ts
```diff
@@ -341,17 +453,17 @@ function instantAmend(optionsParam) {
...
 
-function hasTimeOptions(options) {
+function hasTimeOptions(options: Parameters<TypesWithToLocaleString['toLocaleString']>[1]) {
   return (
     'hour' in options || 'minute' in options || 'second' in options || 'timeStyle' in options || 'dayPeriod' in options
   );
 }
 
-function isTemporalObject(obj) {
+function isTemporalObject(obj: unknown) {
```
12wrigja, line 466: Nit: Is the returned type being inferred correctly here? Maybe it would be better to explicitly type is as `value is AnyTemporalType` (or whatever the right combination actually is)?

index.d.ts
```diff
@@ -990,14 +992,8 @@ export namespace Temporal {
...
     with(timeLike: Temporal.PlainTime | PlainTimeLike, options?: AssignmentOptions): Temporal.PlainTime;
-    add(
-      durationLike: Temporal.PlainTime | Temporal.Duration | DurationLike | string,
-      options?: ArithmeticOptions
-    ): Temporal.PlainTime;
-    subtract(
-      durationLike: Temporal.PlainTime | Temporal.Duration | DurationLike | string,
-      options?: ArithmeticOptions
-    ): Temporal.PlainTime;
+    add(durationLike: Temporal.Duration | DurationLike | string, options?: ArithmeticOptions): Temporal.PlainTime;
```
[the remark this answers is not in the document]
justingrant: Yep. Not sure how this bug got in here, but TS found it!

lib/intl.ts
justingrant: Yes. Typing this damn function was a huge hassle. See https://github.com/js-temporal/temporal-polyfill/pull/81#discussion_r726540190. LMK if you can think of a better solution.

lib/intl.ts
12wrigja: If it doens't "just work" then let's just leave it and I'll take a look some other time. Unknown is the 'better' "this could be anything" because it requires actually type-narrowing unknown before using it (instead of turning type-checking off entirely, which is what `any` does), but it can certainly be harder to use as a result.

PR #85, opened 2021-10-18
Port latest Normative changes from tc39/proposal-temporal.

12wrigja:
This also bumps up the Test262 test cases to commit  3dbf940c9bff08b05a87e34e98864887550bda8d to match the current state of the tc39/proposal-temporal repo.


12wrigja:
Fine by me - thanks for the heads up.

justingrant:
> Do you mean [tc39/proposal-temporal#1878](https://github.com/tc39/proposal-temporal/issues/1878) ?

That's the root cause spec bug, but #1861 has links to Test262 changes that need to be reverted.

12wrigja:
I'm re-using this PR to port over the various changes that have landed recently to the non-prod polyfill.

lib/zoneddatetime.ts
```diff
@@ -212,9 +203,29 @@ export class ZonedDateTime implements Temporal.ZonedDateTime {
...
+      ['millisecond', 0],
+      ['minute', 0],
+      ['month', undefined],
+      ['monthCode', undefined],
+      ['nanosecond', 0],
+      ['second', 0],
+      ['year', undefined],
+      ['offset'],
+      ['timeZone']
+    ];
```
justingrant, line 219: In other types, the `entries` variable is typed (via a type annotation) as something like this:

```js
const entries: [keyof Temporal.ZonedDateTimeLike, 0 | undefined][]
```

This removes the need for the `as any` below.

usually these `entries` values are typed `as const` or with a type annotation on

justingrant (commented):
Looks good so far. One minor type annotation suggestion. 

Probably should change the PR title to match the wider scope. 

Not sure we want to delete PlainMonthDay demitasse tests even if they exist in Test262. Let's discuss pros/cons of this.

ptomato:
Agreed, I would not remove the PlainMonthDay tests.

12wrigja:
Done - I've brought back and fixed up the single failing testcase.

ptomato (approved):

PR #88, opened 2021-11-10
Allow string param for `round` and `total`

justingrant:
Enable strings to be used as the param of `round` and `total` methods. Port of https://github.com/tc39/proposal-temporal/pull/1875.

Note that this will fail two Test262 tests because passing invalid strings to this method used to throw a TypeError but now will throw a RangeError. 

After #1875 is merged, those tests should start passing. For that reason, we should probably wait to merge this PR until #1875 has landed over in proposal-temporal.

### DIFF
--- a/lib/duration.ts
+++ b/lib/duration.ts
@@ -292,7 +292,10 @@ export class Duration implements Temporal.Duration {
       microseconds,
       nanoseconds
     );
-    const options = ES.GetOptionsObject(optionsParam);
+    const options =
+      typeof optionsParam === 'string'
+        ? (ES.CreateOnePropObject('smallestUnit', optionsParam) as Exclude<typeof optionsParam, string>)
+        : ES.GetOptionsObject(optionsParam);
     let smallestUnit = ES.ToSmallestTemporalUnit(options, undefined);
     let smallestUnitPresent = true;
     if (!smallestUnit) {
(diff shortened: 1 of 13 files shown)


lib/duration.ts
```diff
@@ -292,7 +292,8 @@ export class Duration implements Temporal.Duration {
       microseconds,
       nanoseconds
     );
-    const options = ES.GetOptionsObject(optionsParam);
+    const options =
+      typeof optionsParam === 'string' ? { smallestUnit: optionsParam } : ES.GetOptionsObject(optionsParam);
```
ptomato, line 296: I think we need to change this, I left a comment on the proposal-temporal PR: https://github.com/tc39/proposal-temporal/pull/1875/files#r748524136

justingrant: Fixed!  To reduce bundle size and to make the TS part easier I implemented the change slightly differently here than in proposal-temporal, but result is the same.

lib/ecmascript.ts
```diff
@@ -5016,6 +5016,12 @@ export function GetOptionsObject<T>(options: T) {
   throw new TypeError(`Options parameter must be an object, not ${options === null ? 'null' : `${typeof options}`}`);
 }
 
+export function CreateOnePropObject<K extends string, V extends unknown>(propName: K, propValue: V) {
```
12wrigja, line 5019: Nit: Can we explicitly write out the return value for this instead of letting it be inferred? Probably something like `{[K]: V}` ?

12wrigja (approved):

lib/ecmascript.ts
justingrant: Done in latest commit. The only syntax I could get to work was this: `{ [k in K]: V }`.

justingrant:
Merging this now, but will need a follow-up PR to update Test262 submodule after https://github.com/tc39/test262/pull/3304 lands.

In the meantime, I verified this PR by running tests locally from https://github.com/tc39/test262/pull/3304 that were related to this PR, and they all passed. Then I added the two failing tests in current Test626 to the ignored file.

PR #90, opened 2021-11-18
Port https://github.com/tc39/proposal-temporal/pull/1872.

12wrigja:


lib/ecmascript.ts
```diff
@@ -140,12 +140,16 @@ export function ToPositiveInteger(valueParam: unknown, property?: string): numbe
   return value;
 }
 
-function ToIntegerNoFraction(valueParam: unknown): number {
+export function ToIntegerWithoutRounding(valueParam: unknown): number {
   const value = ToNumber(valueParam);
+  if (NumberIsNaN(value)) return 0;
```
justingrant, line 145: @ptomato, why is `NaN` treated as zero in this context?  I'm not questioning the decision just curious.

justingrant (approved):

lib/ecmascript.ts
ptomato: The only reason is because [ToIntegerOrInfinity](https://tc39.es/ecma262/#sec-tointegerorinfinity) also does that.

ptomato (approved):

PR #95, opened 2021-11-30
Port tc39/proposal-temporal/pull/1873

12wrigja:


12wrigja:
(the right PR this time... :) ) 

@justingrant Maybe you can enlighten me: I believe that this PR should fix the issue mentioned [here](https://github.com/js-temporal/temporal-polyfill/blob/3008a670b3758abe1b2341d54da4217b251bc234/lib/ecmascript.ts#L4786-L4799) and that code can be removed in the 2 places that it exists in ecmascript.ts (as that is being done in a single more centralized place). However, if I do remove those lines there's tests that fail. As far as I can tell, those same tests that fail are still being run in the proposal repo and those corresponding lines don't (and maybe didn't ever?) exist in that (proposal) repo, so how are those tests passing over there?

If you checkout f36200f25db543c7f6dee2dcdefc9be2a122d48e and npm run test that will show you what I'm seeing.

justingrant:
I looked into this for a while. The tests are correct and https://github.com/js-temporal/temporal-polyfill/commit/f36200f25db543c7f6dee2dcdefc9be2a122d48e is wrong. But I'm not sure where it's wrong. So far I'm stumped. 

Here's a simplified repro case to illustrate the problem:

```js
Temporal.PlainDate.from('2019-02-01').until('2019-02-15', { roundingMode: 'halfExpand', smallestUnit: 'month' })
// => P1M (correct, 14 days rounds up to one 28-day month)
Temporal.PlainDate.from('2019-01-01').until('2019-02-15', { roundingMode: 'halfExpand', smallestUnit: 'month' })
// => P1M (wrong, because rounding should happen at the last month, so result should be P2M)
```

@12wrigja, do you want to try stepping through both the buggy code and proposal-temporal in a debugger to see if you can spot where they diverge?

FYI @ptomato in case any of this looks familiar.

lib/ecmascript.ts
```diff
@@ -3428,9 +3415,9 @@ export function BalanceDurationRelative(
   if (sign === 0) return { years, months, weeks, days };
 
   let calendar;
-  let relativeTo: Temporal.PlainDateTime;
+  let relativeTo: Temporal.PlainDate;
```
ptomato, line 3418: Could the discrepancy be that this should be typed as `Temporal.PlainDate | undefined`?

justingrant: TS type differences shouldn't cause different runtime behavior, though.

12wrigja:
I'll take a look in a debugger at some point this week - just wanted to make sure my intuition was right and that we should be able to remove the chunks I added back in 2c1fbb58d471a8ee041a1e6dee92285981849139.

lib/ecmascript.ts
```diff
@@ -4830,16 +4793,10 @@ export function RoundDuration(
       const weeksInDays = DaysUntil(yearsMonthsLater, yearsMonthsWeeksLater);
       // This conversion is needed because of an issue in the spec that's likely to change.
       // See https://github.com/tc39/proposal-temporal/issues/1821.
-      relativeTo = CreateTemporalDateTime(
+      relativeTo = CreateTemporalDate(
```
justingrant, line 4796: I found the cause of some of the test failures: in place of this line, proposal-temporal has this code:
```
      relativeTo = yearsMonthsLater;
```

Iin https://github.com/js-temporal/temporal-polyfill/commit/2c1fbb58d471a8ee041a1e6dee92285981849139, `relativeTo` is never reassigned here, but (at least according to proposal-temporal) it should be reassigned here.

There's still two failing tests related to `'year'` units. I'll take a look at those next.

lib/ecmascript.ts
```diff
@@ -4785,16 +4754,10 @@ export function RoundDuration(
...
         GetSlot(relativeToDateOnly, ISO_DAY),
-        0,
-        0,
-        0,
-        0,
-        0,
-        0,
         GetSlot(relativeTo, CALENDAR)
       );
       let { days: oneYearDays } = MoveRelativeDate(calendar, relativeTo, oneYear);
```
justingrant, line 4763: If you want to remove the conversion above, then this line needs be changed too:

```suggestion
      let { days: oneYearDays } = MoveRelativeDate(calendar, relativeToDateOnly, oneYear);
```

This is a good lesson for the future to be extra-careful when copying code from proposal-temporal!  😄

justingrant:
> just wanted to make sure my intuition was right and that we should be able to remove the chunks I added back in [2c1fbb5](https://github.com/js-temporal/temporal-polyfill/commit/2c1fbb58d471a8ee041a1e6dee92285981849139).

Yep. Once the two lines I commented about are fixed. 😄

lib/ecmascript.ts
12wrigja: Thanks Justin! This really helped - I diffed the content of the RoundDuration method for both implementations and the reassignment here turned out to be critical - I needed to change all usages of `relativeToDateOnly` back to `relativeTo` along with this reassignment and then all the Demitasse tests pass.

Really appreciate you diving in to take a look at this for me.

lib/ecmascript.ts
12wrigja: Definitely - though, as I mentioned above, this is actually the right content for this line if the reassignment was there.

lib/ecmascript.ts
justingrant: BTW, the reason why lack of `undefined` in that declaration doesn't cause a TS error is because `strictNullChecks` is still `false`. I filed #98 to track the (considerable) work to be able to turn it on.

justingrant (approved):
Cool!  Thanks @12wrigja!

lib/ecmascript.ts
justingrant: Yep. Originally, `relativeToDateOnly` was there because TS emitted a compiler error when passing `relativeTo` to some other method (forget which) that required a PlainDate.  Looks like that error doesn't show up anymore, perhaps because that method was changed to also accept a PlainDateLike too?  Regardless, glad this helped.

lib/ecmascript.ts
```diff
@@ -4717,17 +4685,18 @@ export function RoundDuration(
   const TemporalDuration = GetIntrinsic('%Temporal.Duration%');
   let calendar, zdtRelative;
   // if it's a ZDT, it will be reassigned to a PDT below.
```
justingrant, line 4687: This comment is inaccurate now

justingrant: Oh, you merged it already!  I'll submit a new PR to fix this up.

12wrigja: Oops - the GH UI didn't show this to me :| That would be great.

PR #96, opened 2021-11-30
Port tc39/proposal-temporal/pull/1874

12wrigja:


12wrigja:
@justingrant Maybe you can enlighten me: I believe that this PR should fix the issue mentioned [here](https://github.com/js-temporal/temporal-polyfill/commit/ac78fd9ddff96dd792703a4948d11196d52fbbed#diff-e14772ac556c06a8e66dd62385ca318959717ac8ddbc367420e3dcc9527ae4e2R4778-R4791) and that code can be removed in the 2 places that it exists in ecmascript.ts (as that is being done elsewhere). However, if I do remove those lines there's tests that fail. As far as I can tell, those same tests that fail are still being run in the proposal repo and those corresponding lines don't (and maybe didn't ever?) exist in the repo, so how are those tests passing?

If you checkout 61ed7884eeb393f9ae53f273359d563042775512 and `npm run test` that will show you what I'm seeing.

12wrigja:
(disregard that comment above - that was meant for a different PR! 😆 )

12wrigja:
A reminder for myself tomorrow: the plainmonthday Deimtasse tests don't exist in the proposal repo anymore, but instead of deleting the corresponding assertions outright in this repo it should be split into a 'Z not supported' testcase like was done for plaintime.

PR #97, opened 2021-12-03
Fix GetFormatterParts for Firefox Nightly

### LINKED ISSUE #92: GetFormatterParts failing on latest Firefox Nightly
From https://bugzilla.mozilla.org/show_bug.cgi?id=1742572:
> Standalone test case:
> ```js
> function GetFormatterParts(timeZone, epochMilliseconds){
>   const formatter = new Intl.DateTimeFormat("en-us", {
>     timeZone,
>     hour12: false,
>     era: "short",
>     year: "numeric",
>     month: "numeric",
>     day: "numeric",
>     hour: "numeric",
>     minute: "numeric",
>     second: "numeric"
>   });
>   const datetime = formatter.format(new Date(epochMilliseconds));
>   const [date, fullYear, time] = datetime.split(/,\s+/);
>   const [month, day] = date.split(" ");
>   const [year, era] = fullYear.split(" ");
>   const [hour, minute, second] = time.split(":");
>   return {
>     year: era === "BC" ? -year + 1 : +year,
>     month: +month,
>     day: +day,
>     hour: hour === "24" ? 0 : +hour,
>     minute: +minute,
>     second: +second
>   };
> }
> 
> GetFormatterParts("America/Los_Angeles", Date.now());
> ```
> 
> `GetFormatterParts` expects that the formatted date-time string can be split into three parts through the regular expression `/,\s+/`. This worked before the update, because the formatted string looked like `"11 23, 2021 AD, 05:00:00"`. But after the update to ICU 70, which includes an update to CLDR 40, the string now looks like `"11/23/2021 A, 05:00:00"`, i.e. can no longer be separated into three parts when searching for a comma.

bananer:
This fixes #92 with the suggestions by @justingrant 

-   Accept 1+ non-alphanumeric characters as separators. No need to restrict to spaces.
-   Only look at the first character of the era, to match both "B" and "BC". Might as well make the comparison case-insensitive while we're at it.

Tested on Firefox Nightly 96.0a1, Firefox Developer 95.0b12 (with the previous behavior) and Chrome 96.0.4664.45

### DIFF
--- a/lib/ecmascript.ts
+++ b/lib/ecmascript.ts
@@ -2845,7 +2845,7 @@ export function GetIANATimeZonePreviousTransition(epochNanoseconds: bigInt.BigIn
 // ts-prune-ignore-next TODO: remove this after tests are converted to TS
 export function GetFormatterParts(timeZone: string, epochMilliseconds: number) {
   const formatter = getIntlDateTimeFormatEnUsForTimeZone(timeZone);
-  // FIXME: can this use formatToParts instead?
+  // Using `format` instead of `formatToParts` for compatibility with older clients
   const datetime = formatter.format(new Date(epochMilliseconds));
   const [month, day, year, era, hour, minute, second] = datetime.split(/[^\w]+/);
   return {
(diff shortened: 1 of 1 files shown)


lib/ecmascript.ts
```diff
@@ -2847,12 +2847,9 @@ export function GetFormatterParts(timeZone: string, epochMilliseconds: number) {
   const formatter = getIntlDateTimeFormatEnUsForTimeZone(timeZone);
   // FIXME: can this use formatToParts instead?
```
justingrant, line 2848:
```suggestion
  // Using `format` instead of `formatToParts` for compatibility with older clients
```

justingrant (approved):
Thanks for the PR!  Looks good. I'll make a small edit to a comment and merge. Thanks again.

PR #100, opened 2021-12-06
Remove getOffsetNanosecondsFor fallback from polyfill

12wrigja:
When porting over [063dd4d7a33cef728db6744d68cb4a21a720290b](https://github.com/tc39/proposal-temporal/commit/063dd4d7a33cef728db6744d68cb4a21a720290b) I found when retrieving the `getOffsetNanosecondsFor` method we don't actually check the type is callable. Fixing this then lead me to remove the registration of the intrinsic, as it's now unused (`undefined` is not callable, and will throw a TypeError). The [corresponding function](https://github.com/tc39/proposal-temporal/blob/f2405e18f5f9fdf2d80250b1075f16f40547c911/polyfill/lib/ecmascript.mjs#L1759-L1769) in the proposal repo doesn't have this fallback either.

### DIFF
--- a/lib/ecmascript.ts
+++ b/lib/ecmascript.ts
@@ -2179,8 +2179,8 @@ export function GetOffsetNanosecondsFor(
   instant: TimeZoneProtocolParams['getOffsetNanosecondsFor'][0]
 ) {
   let getOffsetNanosecondsFor = timeZone.getOffsetNanosecondsFor;
-  if (getOffsetNanosecondsFor === undefined) {
-    getOffsetNanosecondsFor = GetIntrinsic('%Temporal.TimeZone.prototype.getOffsetNanosecondsFor%');
+  if (typeof getOffsetNanosecondsFor !== 'function') {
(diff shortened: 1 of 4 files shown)


ptomato:
This looks correct at first glance, though I'm not sure where the two polyfills diverged!

12wrigja:
Nor I - I'm about to take a look at that. I know part of the discrepancy was due to my removal of the es-abstract dependency (ES.GetMethod performs a more complex check than this), but I do want to understand what happened to the fallback logic...

12wrigja:
https://github.com/tc39/proposal-temporal/pull/1929

Seems the fallback was removed and we never ported that over here?

Better late than never I suppose.

ptomato:
So is this PR now effectively a port of that proposal-temporal#1929 PR?

12wrigja:
I guess so, yeah.

ptomato (approved):
:+1:

PR #101, opened 2021-12-06
Update manual ISO string tests to account for disallowed Z

12wrigja:
Since #1874, strings with Z UTC designator are disallowed for Plain types.
This was not yet reflected in the manual ISO string generator test.


ptomato (approved):
Yep, this looks like what I wrote the other day :-)
I still hope to update this to be reflected in the grammar directly, rather than in this extra step, but that's a project for another day.

12wrigja:
> Yep, this looks like what I wrote the other day :-)

I would hope so... it's a direct `git cherry-pick` :D

PR #102, opened 2021-12-06
Update dependencies

justingrant:
- Update dependencies (except demitasse which requires test changes)
- Fixes eslint & prettier issues
- Add James to the contributors list
- Append `npm install` step in the package.json `update` script
- Add `fix` script to package.json that auto-fixes prettier and eslint
- Add `prettier` script to package.json which includes non-code files (e.g. json) that eslint ignores. Also include this in `npm run lint` so that PRs will catch these prettier issues too.


package.json
```diff
@@ -22,12 +22,15 @@
...
     "playground": "TEMPORAL_PLAYGROUND=1 npm run build && node --experimental-modules --no-warnings --icu-data-dir node_modules/full-icu -r ./dist/playground.cjs",
-    "lint": "eslint . --ext ts,js,mjs,.d.ts --max-warnings 0 --cache \"$@\"",
+    "lint": "eslint . --ext ts,js,mjs,.d.ts --max-warnings 0 --cache \"$@\" && npm run prettier",
     "postlint": "npm run tscheck",
     "prune": "ts-prune -e -i test/tc39 -i \"(lib/index|lib/init|index.d).ts\"",
-    "pretty": "eslint . --ext ts,js,mjs,.d.ts --fix",
+    "prettier": "prettier lib/* ./test/*.mjs ./*.md ./*.json ./*.d.ts ./*.js ./.*.yml --check",
+    "prettierfix": "prettier lib/* ./test/*.mjs ./*.md ./*.json ./*.d.ts ./*.js ./.*.yml  --check --write",
+    "eslintfix": "eslint . --ext ts,js,mjs,.d.ts --fix",
+    "fix": "npm run eslintfix && npm run prettierfix",
```
12wrigja, line 33: 🙏 thank you!

I've been jumping back and forth between this repo and others lately, and having to remember that it's `npm run pretty` was a bit confusing.

12wrigja (approved):

test/duration.mjs
```diff
@@ -371,7 +371,9 @@ describe('Duration', () => {
     });
     it('unrepresentable number is not allowed', () => {
       units.forEach((unit, ix) => {
+        // eslint-disable-next-line @typescript-eslint/no-loss-of-precision,no-loss-of-precision
```
justingrant, line 374: After some experimentation, I found that I needed to disable both variations of this rule because the prettier command-line tool would complain if one of them wasn't disabled, while the in-IDE prettier would complain if the other one wasn't disabled.  There's probably a better way to handle this (maybe exclude code files from running using the prettier command line and rely on ESLint only for code files, using prettier only for non-code files) but I was lazy and didn't want to spend the time to figure that out given that the workaround above is only needed in a few spots.

Issue #98, opened 2021-12-03
Enable `strictNullChecks` and `strictPropertyInitialization` in TS compilation


justingrant:
`strictNullChecks` and `strictPropertyInitialization` are currently commented out in tsconfig.json. At some point we should fix the 300+ TS errors that show up with these flags enabled.

jens-ox:
Hi Justin! If you want I can look into it. I wanted to start working with the new Temporal APIs, and this seems like a nice starting point. 😊

justingrant:
Fixed in #109

PR #103, opened 2021-12-07
RFC: Fix null-check stuff

jens-ox:
Hi all!

As mentioned in #98, there are currently quite a few alerts when enabling strict null checks etc in the tsconfig. If it makes sense I'd really like to help, but I'm not sure what the preferred way of handling undefined stuff is.

This PR fixes all TS issues in `plainDate.ts` when turning on the all the currently commented-out flags in the tsconfig. In most cases, the only change is to allow methods that are called with possibly undefined values to handle that gracefully. Or do you prefer to throw more often if something is undefined?

If this looks ok to you I can start fixing the remaining TS issues.

### DIFF
--- a/lib/ecmascript.ts
+++ b/lib/ecmascript.ts
@@ -38,7 +38,6 @@ import type {
   PlainYearMonthParams,
   FieldRecord
 } from './internaltypes';
-import { GetIntrinsic } from './intrinsicclass';
 import {
   CreateSlots,
   GetSlot,
@@ -73,7 +72,7 @@ import {
   MICROSECONDS,
   NANOSECONDS
 } from './slots';
-import { IsBuiltinCalendar } from './calendar';
+import { Calendar, IsBuiltinCalendar } from './calendar';
 const DAY_SECONDS = 86400;
 const DAY_NANOS = bigInt(DAY_SECONDS).multiply(1e9);
 const NS_MIN = bigInt(-DAY_SECONDS).multiply(1e17);
@@ -205,6 +204,15 @@ const SINGULAR_PLURAL_UNITS: readonly PluralAndSingularUnitTuple<Temporal.DateTi
 ] as const;
 
 import * as PARSE from './regex';
+import { PlainDate } from './plaindate';
+import { Duration } from './duration';
+import { Instant } from './instant';
+import { PlainMonthDay } from './plainmonthday';
+import { PlainTime } from './plaintime';
+import { PlainDateTime } from './plaindatetime';
+import { TimeZone } from './timezone';
+import { PlainYearMonth } from './plainyearmonth';
+import { ZonedDateTime } from './zoneddatetime';
 
 const IntlDateTimeFormatEnUsCache = new Map<string, Intl.DateTimeFormat>();
 
@@ -1257,8 +1265,7 @@ export function ToTemporalDate(
   ToTemporalOverflow(options); // validate and ignore
   const { year, month, day, calendar, z } = ParseTemporalDateString(ToString(item));
   if (z) throw new RangeError('Z designator not supported for PlainDate');
-  const TemporalPlainDate = GetIntrinsic('%Temporal.PlainDate%');
-  return new TemporalPlainDate(year, month, day, calendar); // include validation
+  return new PlainDate(year, month, day, calendar); // include validation
 }
 
 export function InterpretTemporalDateTimeFields(
@@ -1364,30 +1371,16 @@ export function ToTemporalDuration(item: DurationParams['from'][0]) {
     ({ years, months, weeks, days, hours, minutes, seconds, milliseconds, microseconds, nanoseconds } =
       ParseTemporalDurationString(ToString(item)));
   }
-  const TemporalDuration = GetIntrinsic('%Temporal.Duration%');
-  return new TemporalDuration(
-    years,
-    months,
-    weeks,
-    days,
-    hours,
-    minutes,
-    seconds,
-    milliseconds,
-    microseconds,
-    nanoseconds
-  );
+  return new Duration(years, months, weeks, days, hours, minutes, seconds, milliseconds, microseconds, nanoseconds);
 }
 
 export function ToTemporalInstant(item: InstantParams['from'][0]) {
   if (IsTemporalInstant(item)) return item;
   if (IsTemporalZonedDateTime(item)) {
-    const TemporalInstant = GetIntrinsic('%Temporal.Instant%');
-    return new TemporalInstant(GetSlot(item, EPOCHNANOSECONDS));
+    return new Instant(GetSlot(item, EPOCHNANOSECONDS));
   }
   const ns = ParseTemporalInstant(ToString(item));
-  const TemporalInstant = GetIntrinsic('%Temporal.Instant%');
-  return new TemporalInstant(ns);
+  return new Instant(ns);
 }
 
 export function ToTemporalMonthDay(
@@ -1449,8 +1442,7 @@ export function ToTemporalTime(
       );
     }
     if (IsTemporalDateTime(item)) {
-      const TemporalPlainTime = GetIntrinsic('%Temporal.PlainTime%');
-      return new TemporalPlainTime(
+      return new PlainTime(
         GetSlot(item, ISO_HOUR),
         GetSlot(item, ISO_MINUTE),
         GetSlot(item, ISO_SECOND),
@@ -1482,8 +1474,7 @@ export function ToTemporalTime(
       throw new RangeError('PlainTime can only have iso8601 calendar');
     }
   }
-  const TemporalPlainTime = GetIntrinsic('%Temporal.PlainTime%');
-  return new TemporalPlainTime(hour, minute, second, millisecond, microsecond, nanosecond);
+  return new PlainTime(hour, minute, second, millisecond, microsecond, nanosecond);
 }
 
 export function ToTemporalYearMonth(
@@ -1533,8 +1524,7 @@ export function InterpretISODateTimeOffset(
   offsetOpt: Temporal.OffsetDisambiguationOptions['offset'],
   matchMinute: boolean
 ) {
-  const DateTime = GetIntrinsic('%Temporal.PlainDateTime%');
-  const dt = new DateTime(year, month, day, hour, minute, second, millisecond, microsecond, nanosecond);
+  const dt = new PlainDateTime(year, month, day, hour, minute, second, millisecond, microsecond, nanosecond);
 
   if (offsetBehaviour === 'wall' || offsetOpt === 'ignore') {
     // Simple case: ISO string without a TZ offset (or caller wants to ignore
@@ -1636,8 +1626,7 @@ export function ToTemporalZonedDateTime(
     } else if (!offset) {
       offsetBehaviour = 'wall';
     }
-    const TemporalTimeZone = GetIntrinsic('%Temporal.TimeZone%');
-    timeZone = new TemporalTimeZone(ianaName);
+    timeZone = new TimeZone(ianaName);
     if (!calendar) calendar = GetISO8601Calendar();
     calendar = ToTemporalCalendar(calendar);
     matchMinute = true; // ISO strings may specify offset with less precision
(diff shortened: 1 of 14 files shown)


lib/plaindate.ts
```diff
@@ -188,8 +188,8 @@ export class PlainDate implements Temporal.PlainDate {
       this
     ));
 
-    const Duration = GetIntrinsic('%Temporal.Duration%');
-    return new Duration(years, months, weeks, days, 0, 0, 0, 0, 0, 0);
+    // const Duration = GetIntrinsic('%Temporal.Duration%')!;
+    return new Temporal.Duration(years, months, weeks, days, 0, 0, 0, 0, 0, 0);
```
ptomato, line 192: This we can't do, because it would break if user code monkey-patched `globalThis.Temporal.Duration`. (That's the reason for the GetIntrinsic mechanism appearing all over the place, so that we can continue to use the original values internally even if they are replaced in user code.) Sorry, this should've been documented somewhere more clearly! Probably in `intrinsic.ts`.

I think the best solution here (and elsewhere in the file) would be to type GetIntrinsic such that TypeScript knows that it returns Temporal.Duration for an argument of `%Temporal.Duration%`. Unfortunately I don't immediately have an idea of how to do that properly :smile:

lib/slots.ts
```diff
@@ -275,9 +275,10 @@ export function HasSlot(container: unknown, ...ids: (keyof Slots)[]): boolean {
   return !!myslots && ids.reduce((all: boolean, id) => all && id in myslots, true);
 }
 export function GetSlot<KeyT extends keyof Slots>(
-  container: Slots[typeof id]['usedBy'],
+  container: Slots[typeof id]['usedBy'] | undefined,
```
ptomato, line 278: I think it signifies a bug in the polyfill if we end up calling `GetSlot(undefined, ...)` so I'd be happy if TS would complain at compile time rather than resorting to a runtime error. I wonder if there's any way to do that?

ptomato (commented):
How exciting to receive a PR from someone new to the project and eager to help, thanks very much!

I gave it a quick look and commented on two things that stood out to me. I can take a closer look again later, but it seems like adjusting the GetIntrinsic thing might change a lot of other stuff.

lib/plaindate.ts
jens-ox: Could this be avoided by just doing `import { Duration } from './duration'`? Or am I missing something? 😅 
Makes sense otherwise, if importing is not sensible here I'll try to teach TypeScript how `GetIntrinsic` works 😊

12wrigja: I don't think we can import as it causes circular value dependencies between the different files? If that would work that could be one potential fix.

If we need to stick with `GetIntrinsic` it should already be typed such that a key of `%Temporal.Duration%` returns `typeof Temporal.Duration`, but what probably needs to change is that the return type for GetIntrinsic needs to be `NonNullable<typeof INTRINSICS[key]>` [here](https://github.com/js-temporal/temporal-polyfill/blob/08346dc5bc809e7575eacde3200f9775fe19c378/lib/intrinsicclass.ts#L137-L139).

12wrigja:
A quick drive-by comment: it is probably easiest to fix these flags by starting at the "root-most" files and working outwards. In this project, that would be in slots.ts and ecmascript.ts. Even small changes here will "ripple" through the rest of the files and might mean other fixes are necessary or obsolete.

jens-ox:
> A quick drive-by comment: it is probably easiest to fix these flags by starting at the "root-most" files and working outwards. In this project, that would be in slots.ts and ecmascript.ts. Even small changes here will "ripple" through the rest of the files and might mean other fixes are necessary or obsolete.

Yeah, that makes sense. `plaindate` just seemed like the easiest to understand for a newcomer like me 😊 
Once it's clear how all of this can best be done I'll close this PR and work on a "proper" one.

lib/slots.ts
jens-ox: In this case it can be `undefined` if an incorrect disambiguation is given to `BuiltinTimeZoneGetInstantFor`, so maybe it would be best to just throw?

12wrigja:
On second thought, I think probably the two hardest files to change (but definitely most useful...) would be slots.ts and intrinsicclass.ts, as that is where a good chunk of the "type magic" is needed (places where TS can't easily infer the right typing). I'd suggest ecmascript.ts as third.

lib/plaindate.ts
jens-ox: Circular dependencies should be fine as there's already ecmascript -> calendar -> ecmascript :D
If just importing is an option that might be cleaner, right?

jens-ox:
I tried removing `GetIntrinsic` - this doesn't seem to break anything and also has the nice side-effect to clear up some internal naming inconsistencies.

lib/slots.ts
ptomato: Maybe we can tighten the types in BuiltinTimeZoneGetInstantFor to ensure at compile time that no incorrect disambiguation can be given there? Or otherwise throw earlier if that happens?

ptomato:
If GetIntrinsic works, that might indeed be cleaner! But I'd be concerned, does it still work when bundling for a platform that doesn't have ES6 modules? I'd also suggest verifying that it really is robust in the face of monkey-patching.

jens-ox:
> does it still work when bundling for a platform that doesn't have ES6 modules?

I haven't looked that deeply into the Rollup config used here, but as long as Rollup is properly configured, yes (and there seems to be a CommonJS export generated here, so we should be good to go).

> I'd also suggest verifying that it really is robust in the face of monkey-patching

To be honest I never heard of monkey-patching before, so I might have to do a bit of reading first 🙈 
But as far as I can see monkey-patching is not possible here.

lib/intrinsicclass.ts
```diff
@@ -134,6 +134,3 @@ export function DefineIntrinsic<KeyT extends IntrinsicDefinitionKeys>(name: KeyT
   if (INTRINSICS[key] !== undefined) throw new Error(`intrinsic ${name} already exists`);
   INTRINSICS[key] = value;
 }
-export function GetIntrinsic<KeyT extends keyof typeof INTRINSICS>(intrinsic: KeyT): typeof INTRINSICS[KeyT] {
```
justingrant, line 137: Not sure this can be completely removed, even if the imports work, because there are some objects and methods (e.g. `from` methods, maybe intrinsic prototypes) that still need to be accessed.

lib/plaindate.ts
```diff
@@ -1,5 +1,5 @@
 import * as ES from './ecmascript';
-import { GetIntrinsic, MakeIntrinsicClass } from './intrinsicclass';
+import { MakeIntrinsicClass } from './intrinsicclass';
```
justingrant, line 2: Where is `MakeIntrinsicClass` used in this and other files where it's imported?

lib/plaindate.ts
```diff
@@ -351,6 +350,7 @@ export class PlainDate implements Temporal.PlainDate {
       calendar
     );
     const instant = ES.BuiltinTimeZoneGetInstantFor(timeZone, dt, 'compatible');
+    if (!instant) throw new TypeError(`Could not create instant for timezone "${timeZone}" and dateTime "${dt}"`);
```
justingrant, line 353: Can `instant` actually be undefined here?  I don't think it can be. The type problem is  upstream. For example, `DisambiguatePossibleInstants` needs an assertion at the end to prevent `undefined` from being returned, e.g. 

```js
    throw new Error(`assertion failed: invalid disambiguation value ${disambiguation}`);
```

lib/slots.ts
justingrant: Seems like there may be a legit issue with `DisambiguatePossibleInstants` because the spec says there should be an assertion that the `disambiguation` value is valid but the polyfill lacks this assertion. We could add a line at the end of that method e.g. 
```js
    throw new Error(`assertion failed: invalid disambiguation value ${disambiguation}`);
```

justingrant:
> > I'd also suggest verifying that it really is robust in the face of monkey-patching
> 
> To be honest I never heard of monkey-patching before, so I might have to do a bit of reading first 🙈 But as far as I can see monkey-patching is not possible here.

Yep, there are a few possible issues. I'm probably missing some, but here's a few things to look at: 
* Does TS allow the `prototype` of TS-created classes to be changed?  Currently, the polyfill allows calling `setPrototypeOf` on Temporal types, and if that's the correct behavior then the TS polyfill can't behave differently. 
* When we create new instances internally in ecmascript.ts by manually using the prototype object, do we use the intrinsic prototype or the monkeypatchable one?
* What about static methods like `Calendar.from` which AFAIK are used internally?  Those need to be still accessible internally even if the user monkeypatches them. 

The first two bullet points above are covered here: https://github.com/tc39/proposal-temporal/issues/1965

@jens-ox thanks so much for your work on this PR and in uncovering questions like this!

justingrant (changes requested):
Thanks @jens-ox! Lots of notes are inline. You've already helped us find a few issues which is great. Some general feedback: 
* We'll need to figure out the monkeypatching thing before we know if we can really get rid of GetIntrinsic or not.
* I agree with @12wrigja that it'd be best to start upstream with ecmascript.ts, slots, and intrinsics.  In general, before you make a change to a method, consider whether the right change to make is upstream from it.
* @12wrigja and I can help!  Especially with the upstream parts, there's lots of confusing parts of this polyfill. So feel free to tag us with questions if you get stuck.
* In general, you should try really hard to avoid runtime behavior changes (e.g. setting default parameters) to work around type issues. There's usually a way to solve those upstream.
* Also, adding nullability to parameters usually is incorrect; the problem is usually upstream.
* We'll certainly find problems or questions about the proposal-temporal polyfill and/or spec and/or TS types, etc. Examples that I've already found: https://github.com/tc39/proposal-temporal/issues/1963, https://github.com/tc39/proposal-temporal/issues/1965, https://github.com/tc39/proposal-temporal/pull/1964.  So this PR has already been helpful!

lib/ecmascript.ts
```diff
@@ -1699,8 +1688,7 @@ export function CreateTemporalDate(
   isoDay: number,
   calendar: Temporal.CalendarProtocol = GetISO8601Calendar()
 ) {
-  const TemporalPlainDate = GetIntrinsic('%Temporal.PlainDate%');
-  const result = ObjectCreate(TemporalPlainDate.prototype);
+  const result = ObjectCreate(PlainDate.prototype);
```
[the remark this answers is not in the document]
jens-ox: I'm having a difficult time understanding what is happening here - why is some `PlainDate` constructed here in this convoluted way? What exactly should `CreateTemporalDate` do?

What's the difference between
```js
const date = CreateTemporalDate(2021, 1, 1, 'iso8601')
```
and
```js
const date = new PlainDate(2021, 1, 1, 'iso8601')
```
?

jens-ox:
Hi @justingrant! Thanks for your comments. I opened #105 for the `GetIntrinsic` things. I agree that all the `undefined` things should be handled upstream (and then probably throw as early as possible).
I'll open a new PR and will start working from the outermost files, avoiding method signature changes.

lib/ecmascript.ts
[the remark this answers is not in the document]
ptomato: It's a really subtle difference — CreateTemporalDate is an internal operation that creates the object without doing any operations that could be observed from user code. `new PlainDate()` supports the case where the calendar argument is a property bag of the form of any other Temporal type that has a calendar, and accessing the `calendar` property of that property bag is potentially observable. For example, you could do
```js
const loggingDatePropertyBag = {
  year: 2021,
  month: 1,
  day: 1,
  get calendar() {
    console.log('calendar access observed!');
    return 'iso8601';
  },
};
const date = new Temporal.PlainDate(2021, 1, 1, loggingDatePropertyBag);
```
and this would log the "observed!" message.
CreateTemporalDate does not support this  usage. Whenever we call it it's always assumed that we have either already performed the observable operation, or ensured that it isn't necessary.

More info in the commit message of https://github.com/tc39/proposal-temporal/pull/1484/commits/12dab2e53f16a20da824408b0e922713c159c257

lib/plaindate.ts
ptomato: Way down at the bottom :smile:

lib/plaindate.ts
ptomato: How does TypeScript usually handle circular module dependencies?

lib/plaindate.ts
justingrant: Oh whoops, I saw the green and thought "new line" but it's just removing GetIntrinsic; the `MakeIntrinsicClass` was already there.

justingrant:
I think this PR has been replaced by others, so I'm going to close it now. If that's a mistake, feel free to re-open.

PR #104, opened 2021-12-08
Fix TS types for required CalendarProtocol methods

justingrant:
Port of https://github.com/tc39/proposal-temporal/pull/1964


ptomato (approved):
:+1:

PR #108, opened 2021-12-13
Ensure that Temporal prototypes aren't writable

justingrant:
Port of https://github.com/tc39/proposal-temporal/pull/1974

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -8,4 +8,29 @@
 import * as Temporal from './temporal';
 import * as Intl from './intl';
 import { toTemporalInstant } from './legacydate';
+
+// Work around https://github.com/babel/babel/issues/2025.
+const types = [
+  Temporal.Instant,
+  Temporal.Calendar,
+  Temporal.PlainDate,
+  Temporal.PlainDateTime,
+  Temporal.Duration,
+  Temporal.PlainMonthDay,
+  // Temporal.Now, // plain object (not a constructor), so no `prototype`
+  Temporal.PlainTime,
+  Temporal.TimeZone,
+  Temporal.PlainYearMonth,
+  Temporal.ZonedDateTime
+];
+for (const type of types) {
+  const descriptor = Object.getOwnPropertyDescriptor(type, 'prototype') as PropertyDescriptor;
+  if (descriptor.configurable || descriptor.enumerable || descriptor.writable) {
+    descriptor.configurable = false;
+    descriptor.enumerable = false;
+    descriptor.writable = false;
+    Object.defineProperty(type, 'prototype', descriptor);
+  }
+}
+
 export { Temporal, Intl, toTemporalInstant };
--- a/lib/intl.ts
+++ b/lib/intl.ts
@@ -174,8 +174,6 @@ Object.defineProperty(DateTimeFormatImpl, 'name', {
   value: 'DateTimeFormat'
 });
 
-export const DateTimeFormat = DateTimeFormatImpl as unknown as typeof Intl.DateTimeFormat;
-
 DateTimeFormatImpl.supportedLocalesOf = function (
   locales: Params['supportedLocalesOf'][0],
   options: Params['supportedLocalesOf'][1]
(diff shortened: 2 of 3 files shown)


lib/index.ts
```diff
@@ -8,4 +8,29 @@
 import * as Temporal from './temporal';
 import * as Intl from './intl';
 import { toTemporalInstant } from './legacydate';
+
+// Work around https://github.com/babel/babel/issues/2025.
```
12wrigja, line 12: This has already been fixed in Babel - maybe we should just update our build once they release instead?

12wrigja (approved):

lib/intl.ts
```diff
@@ -170,12 +170,12 @@ function DateTimeFormatImpl(
 }
 
 Object.defineProperty(DateTimeFormatImpl, 'name', {
-  writable: true,
+  writable: false,
```
ptomato, line 173: We don't run our polyfill against the whole test262 suite (and that's probably not a good idea), but from browsing it looks like this property is actually intended to be configurable: https://github.com/tc39/test262/blob/main/test/intl402/DateTimeFormat/name.js

lib/index.ts
ptomato: Do we know their release schedule? 7.16.4 was almost a month ago according to GitHub, so if it's monthly and we only have to wait a few days then that seems like a good idea.

lib/intl.ts
```diff
@@ -199,6 +199,15 @@ if ('formatRangeToParts' in IntlDateTimeFormat.prototype) {
 
 DateTimeFormatImpl.prototype = Object.create(IntlDateTimeFormat.prototype, properties);
 
+// Ensure that the prototype isn't writeable.
+Object.defineProperty(DateTimeFormatImpl, 'prototype', {
+  writable: false,
+  enumerable: false,
+  configurable: false
```
ptomato, line 206: We could consider merging this with line 200 and adding `value: Object.create(IntlDateTimeFormat.prototype, properties)` but I don't know if TypeScript is smart enough to see through that...

ptomato (approved):
:+1:

lib/intl.ts
justingrant: Oh!  I'll revert.  I assumed this was a mistake. Serves me right for not checking the spec first.

lib/intl.ts
justingrant: I tried that earlier and got errors saying that a read-only prototype couldn't be overwritten. This code was the fallback solution. :-(

lib/index.ts
justingrant: I actually think it's somewhat unrelated to Babel's schedule, because (unless I'm missing something) anyone who uses babel while bundling their app will be vulnerable. So even if *our* bundling is OK with the new Babel, we'll still want to protect users who are using an older Babel version from being able to build code that replaces the prototype (which will break when they upgrade to the latest Babel).

BTW, we still need to merge the changes to Intl.DateTimeFormat prototype regardless of Babel.

ptomato: If we required at least 7.16.5 wouldn't users' babel update when they updated @js-temporal/polyfill to 0.3.0, as long as the versions were compatible?

12wrigja: https://stackoverflow.com/questions/18875674/whats-the-difference-between-dependencies-devdependencies-and-peerdependencies

Maybe I'm mis-reading this, but it seems that devDependencies (which is what our dep on babel is) aren't installed transitively, so we wouldn't be forcing downstream users to update their babel versions.

That being said, I wonder if there's a way to "suggest" that update to users to encourage they update.... maybe an (ab)use of [peerDependencies](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#peerdependencies) can do this?

justingrant: Yep, my understanding of build-time dependencies is that they're not enforced on the main app. 

Regardless, I'm not sure it's worth the bother. We understand the problem and have a simple workaround that seems to work (and is a no-op at runtime if it's been transpiled with a fixed Babel). So I'm inclined to just declare success and move on.  What do you guys think?

12wrigja: That sounds reasonable - let's just merge this and move on :)

lib/index.ts
```diff
@@ -8,4 +8,29 @@
...
+const types = [
+  Temporal.Instant,
+  Temporal.Calendar,
+  Temporal.PlainDate,
+  Temporal.PlainDateTime,
+  Temporal.Duration,
+  Temporal.PlainMonthDay,
+  // Temporal.Now,
```
12wrigja, line 20: Either remove, or add a comment explaining why Now isn't also not-writeable?

ptomato: I think it's because Now doesn't have its own prototype, it's a plain object, but agreed a comment would be good

ptomato (approved):
:+1:

lib/index.ts
justingrant: Yep, that's why it's omitted. I'll add a comment and merge.

PR #109, opened 2021-12-13
Compile with `strictNullChecks`/`strictPropertyInitialization` (focus on calendar.ts and ecmascript.ts)

justingrant:
This PR focuses on unblocking `strictNullChecks` and `strictPropertyInitialization` in ecmascript.ts and calendar.ts. It's intended as a complement to @jens-ox's work in #105 to remove `GetIntrinsic`.

This PR:
* Enables `strictNullChecks: true` and `strictPropertyInitialization: true` in tsconfig.json.
* Improves calendar.ts TS types (this was most of the work in this PR)
* Removes all unnecessary use of `any` throughout the polyfill
* Updates types of ecmascript.ts and fixes a number of type bugs
* Refactors a few types in intl.ts
* A handful of trivial type changes in other files (ecmascript.ts and calendar.ts were 95%+ of the work).

There is a one-line runtime change (by itself in commit https://github.com/js-temporal/temporal-polyfill/pull/109/commits/fac8f4a2d839be420f4938184099050d79641fcb) but otherwise there should be no runtime-visible changes in this PR, only TS type changes. This means that, aside from that one line in calendar.ts, the transpiled output should be the same before and after this PR, except for parentheses used  for type casts. 

Other runtime changes were split out into other PRs, all of which have already been merged except #112 which will hopefully land soon. This PR is rebased on #112.


lib/ecmascript.ts
```diff
@@ -3412,7 +3434,7 @@ export function BalanceDurationRelative(
   if (sign === 0) return { years, months, weeks, days };
 
   let calendar;
-  let relativeTo: Temporal.PlainDate;
+  let relativeTo = undefined as Temporal.PlainDate | undefined;
```
ptomato, line 3437: I'm not very familiar with TS best practices, but should this be something like `let relativeTo: Temporal.PlainDate? = undefined;`?

ptomato (commented):
I didn't go through calendar.mjs and ecmascript.mjs too carefully, as this is a draft, but let me know what kind of reviews would be helpful! FWIW, it would be easier to review if you could split it into a series of commits that each fix one type of bug (although I realize that may not be possible; when you fix the types on one thing, maybe it "cascades" into another thing?)

justingrant (commented):
> I didn't go through calendar.mjs and ecmascript.mjs too carefully, as this is a draft, but let me know what kind of reviews would be helpful! 

@ptomato Lol, I actually didn't expect any reviews at all until this was further along, so your initial notes are more than I expected! I'll ping when it's ready for a more thorough review.

The best assistance at this point would be to help review and land the proposal-temporal PRs listed in the OP of this PR, because those all came from issues I discovered here while making these TS changes.  The sooner those PRs are merged, the sooner this PR can get wrapped up.

> FWIW, it would be easier to review if you could split it into a series of commits that each fix one type of bug (although I realize that may not be possible; when you fix the types on one thing, maybe it "cascades" into another thing?)

My current plan is to merge all the related proposal-temporal PRs, then port those PRs here, and then rebase this PR on top which should reduce its scope a lot. Furthermore, be removing those PRs it should limit this PR to almost entirely type changes: either none or only a few runtime-visible changes and zero changes that affect tests. It should be both easier and safer to review a big PR with only type changes vs. a mix of type a runtime changes. At least that's the goal. 😄 

But you're right-- it's hard to make these kinds of type changes piecemeal, because once you pull on a thread of types it's hard to split things out.  Luckily, once the runtime changes are removed it should be a lot clearer what's different. 

Sound like an OK plan? Holler with any suggestions to make it easier.

lib/calendar.ts
```diff
@@ -599,83 +601,103 @@ function toUtcIsoDateString({ isoYear, isoMonth, isoDay }: { isoYear: number; is
...
+type CalendarYM = { year: number; month: number };
+type CalendarYearOnly = { year: number };
+type IsoYMD = {
+  year: number;
+  month: number;
+  day: number;
+};
+type Overflow = Temporal.AssignmentOptions['overflow'];
+
+interface NonIsoHelperBaseBase {
```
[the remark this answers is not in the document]
justingrant: Update: here's a comment from the commit I just pushed that includes the new names of the types and an explanation for each: 

```
// Some ES implementations don't include ECMA 402. For this reason, it's helpful
// to ensure a clean separation between the ISO calendar implementation which is
// a part of ECMA 262 and the non-ISO calendar implementation which requires
// ECMA 402.
//
// To ensure this separation, the implementation is split. The `NonIsoImpl`
// interface is the top-level implementation for all non-ISO calendars. This
// type has the same shape as the ECMA 262-only ISO calendar implementation so
// can use the same callers, tests, etc.
//
// A derived interface `NonIsoImplWithHelper` adds a `helper` property that
// includes the remaining non-ISO implementation properties and methods beyond
// the ISO implementation above. The `helper` property's shape is a base
// singleton object common to all calendars (`HelperSharedImpl`) that's extended
// (interface `HelperPerCalendarImpl`) with implementation that varies for each
// calendar.
```

lib/ecmascript.ts
```diff
@@ -448,10 +450,16 @@ function ParseTemporalTimeZoneString(stringIdent: string): Partial<{
   }
   try {
     // Try parsing ISO string instead
-    return ParseISODateTime(stringIdent, { zoneRequired: true });
+    const result = ParseISODateTime(stringIdent, { zoneRequired: true });
+    // TODO: maybe just return result depending on the outcome of
+    // https://github.com/tc39/proposal-temporal/issues/1973
```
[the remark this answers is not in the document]
ptomato: Sounds good, I think that's an equally valid solution to the problem.

lib/ecmascript.ts
```diff
@@ -1045,35 +1061,42 @@ export function LargerOfTwoTemporalUnits<T1 extends Temporal.DateTimeUnit, T2 ex
...
 
-export function ToPartialRecord<B extends AnyTemporalLikeType>(
-  bag: B,
-  fields: readonly (keyof B)[],
-  callerCast?: (value: unknown) => unknown
-) {
-  if (!IsObject(bag)) return false;
-  let any: B;
+export function ToPartialRecord<B extends AnyTemporalLikeType>(bagParam: B, fieldsParam: ReadonlyArray<keyof B>) {
+  if (!IsObject(bagParam)) return false;
```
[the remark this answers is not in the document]
justingrant: Update: we removed it from proposal-temporal too. https://github.com/tc39/proposal-temporal/pull/1980

justingrant:
I split this PR up into a bunch of commits to make reviewing easier, esp. for calendar.ts.

It's currently rebased on top of #112 which will hopefully land soon.

justingrant:
Oh, and it's ready for review now. Reviews may not happen until after the new year.

lib/calendar.ts
```diff
@@ -1691,8 +1691,11 @@ const makeHelperGregorian = (id: BuiltinCalendarId, originalEras: InputEra[]) =>
     eras,
     anchorEra,
     calendarType: 'solar',
-    inLeapYear(calendarDate /*, cache */) {
-      const { year } = this.estimateIsoDate(calendarDate);
+    inLeapYear(calendarDate /*, cache: OneObjectCache */) {
```
ptomato: Do any of the overriding calendars use the month and day to determine `inLeapYear`? If not, maybe we could just pass `year: number` to this method instead of what is effectively `calendarDate: { year: number }`?

lib/calendar.ts
```diff
@@ -455,9 +455,213 @@ impl['iso8601'] = {
...
+    calendar: Temporal.Calendar
+  ): Temporal.PlainYearMonth;
+  monthDayFromFields(
+    this: NonIsoImplWithHelper,
+    fieldsParam: Params['monthDayFromFields'][0],
+    options: NonNullable<Params['monthDayFromFields'][1]>,
+    calendar: Temporal.Calendar
+  ): Temporal.PlainMonthDay;
+  fields(fieldsParam: string[]): Return['fields'];
+  mergeFields(fields: Params['mergeFields'][0], additionalFields: Params['mergeFields'][1]): Return['mergeFields'];
```
ptomato, line 510: No idea if this would work, but is it possible to do something like `mergeFields: typeof Temporal.Calendar.prototype.mergeFields`? Just out of curiosity.

lib/calendar.ts
```diff
@@ -1958,6 +2197,13 @@ const helperJapanese: NonIsoHelperBase = ObjectAssign(
   } as Partial<NonIsoHelperBase>
 );
 
+interface ChineseMonthInfo {
+  [key: string]: { monthIndex: number; daysInMonth: number };
+}
+interface ChineseDraftMonthInfo {
+  [key: string]: { monthIndex: number; daysInMonth?: number };
+}
```
ptomato, line 2139: Also not sure if this is valid TypeScript but can we do something like `interface ChineseMonthInfo = Required<ChineseDraftMonthInfo>`?

lib/calendar.ts
```diff
@@ -1509,21 +1424,28 @@ const helperHebrew: HelperSharedImpl = ObjectAssign({}, helperSharedImpl, {
...
+    cache?: OneObjectCache,
+    overflow: Overflow = 'constrain',
     fromLegacyDate = false
   ) {
-    let { year, eraYear, month, monthCode, day, monthExtra } = calendarDate;
+    // The incoming type is actually CalendarDate (same as args to
+    // Calendar.dateFromParams) but TS isn't smart enough to follow all the
+    // reassignments below, so as an alternative to 10+ type casts, we'll lie
+    // here and claim that the type has `day` and `year` filled in already.
+    let { year, eraYear, month, monthCode, day, monthExtra } = calendarDate as Omit<
```
ptomato, line 1445: Wouldn't this make TypeScript think that the check below (`if (year === undefined ...)`) was redundant? Should the type here be unioned with `{ year?: number, day?: number }` instead?

lib/calendar.ts
```diff
@@ -1537,17 +1459,18 @@ const helperHebrew: HelperSharedImpl = ObjectAssign({}, helperSharedImpl, {
       if (monthExtra) {
         const monthInfo = this.months[monthExtra];
         if (!monthInfo) throw new RangeError(`Unrecognized month from formatToParts: ${monthExtra}`);
-        month = this.inLeapYear({ year }) ? monthInfo.leap : monthInfo.regular;
+        month = (this.inLeapYear({ year }) ? monthInfo.leap : monthInfo.regular) as number;
```
ptomato, line 1462: Does the `as number` work because the `inLeapYear` result makes sure that the `undefined` variant is never selected?

lib/calendar.ts
```diff
@@ -1613,39 +1536,42 @@ const helperIslamic: HelperSharedImpl = ObjectAssign({}, helperSharedImpl, {
   DAYS_PER_ISLAMIC_YEAR: 354 + 11 / 30,
   DAYS_PER_ISO_YEAR: 365.2425,
   constantEra: 'ah',
-  estimateIsoDate(this: typeof helperIslamic & HelperSharedImpl, calendarDate) {
+  estimateIsoDate(
+    this: HelperPerCalendarImpl & { DAYS_PER_ISLAMIC_YEAR: number; DAYS_PER_ISO_YEAR: number },
```
ptomato, line 1540: Is it worth defining a type alias for this, or is it just going to go away anyway when we use ES6 classes here?

lib/ecmascript.ts
```diff
@@ -2160,7 +2164,7 @@ export function ToTemporalTimeZone(temporalTimeZoneLikeParam: TimeZoneParams['fr
   if (IsObject(temporalTimeZoneLike)) {
     if (IsTemporalZonedDateTime(temporalTimeZoneLike)) return GetSlot(temporalTimeZoneLike, TIME_ZONE);
     if (!('timeZone' in temporalTimeZoneLike)) return temporalTimeZoneLike;
-    temporalTimeZoneLike = (temporalTimeZoneLike as { timeZone: typeof temporalTimeZoneLike }).timeZone;
+    temporalTimeZoneLike = (temporalTimeZoneLike as unknown as { timeZone: typeof temporalTimeZoneLike }).timeZone;
```
ptomato, line 2167: I don't quite get this one, why do we need `as unknown` now if it worked before? (Does this mean we should actually be typing `temporalTimeZoneLike` as something like `Temporal.TimeZoneProtocol | { timeZone: Temporal.TimeZoneProtocol }`?)

lib/ecmascript.ts
```diff
@@ -4733,7 +4755,9 @@ export function RoundDuration(
 
   // First convert time units up to days, if rounding to days or higher units.
   // If rounding relative to a ZonedDateTime, then some days may not be 24h.
-  let dayLengthNs: JSBI;
+  // TS doesn't know that `dayLengthNs` is only used if the unit is day or
+  // larger. This makes the cast below acceptable.
+  let dayLengthNs: JSBI = undefined as unknown as JSBI;
```
ptomato, line 4760: I'm not a good judge of what's idiomatic, but might it be better to type this as `JSBI | undefined` and use `!` in the branch where we know it is not undefined?

lib/ecmascript.ts
```diff
@@ -5085,11 +5133,31 @@ export function CreateOnePropObject<K extends string, V>(propName: K, propValue:
 }
 
 function GetOption<P extends string, O extends Partial<Record<P, unknown>>>(
```
ptomato, line 5135: I'm not familiar enough with TypeScript to know what these declarations do if they don't have function bodies. Do they just serve as alternate type signatures for the function below that does have a body?

lib/intl.ts
```diff
@@ -16,16 +16,7 @@ import {
...
-  DateTimeFormatParams as Params,
-  DateTimeFormatReturn as Return,
-  InstantParams,
-  PlainDateParams,
-  PlainDateTimeParams,
-  PlainMonthDayParams,
-  PlainTimeParams,
-  PlainYearMonthParams
-} from './internaltypes';
+import { DateTimeFormatParams as Params, DateTimeFormatReturn as Return } from './internaltypes';
```
ptomato, line 19: Should this be `import type`?

lib/intrinsicclass.ts
```diff
@@ -96,13 +96,13 @@ export function MakeIntrinsicClass(
     });
   }
   for (const prop of Object.getOwnPropertyNames(Class)) {
-    const desc = Object.getOwnPropertyDescriptor(Class, prop);
+    const desc = Object.getOwnPropertyDescriptor(Class, prop) as PropertyDescriptor;
```
ptomato, line 99: Out of curiosity, why doesn't `Object.getOwnPropertyDescriptor` have type `PropertyDescriptor` already?

lib/plaintime.ts
```diff
@@ -43,7 +43,7 @@ type TemporalTimeToStringOptions = {
 function TemporalTimeToString(
   time: Temporal.PlainTime,
   precision: ReturnType<typeof ES.ToSecondsStringPrecision>['precision'],
-  options: TemporalTimeToStringOptions = undefined
+  options: TemporalTimeToStringOptions | undefined = undefined
```
ptomato, line 46: Could be `options?: TemporalTimeToStringOptions`?

ptomato (approved):
Great stuff, I think you'll find that my comments reflect my ignorance of TypeScript idioms rather than problems with the code. Thanks, by the way, for the separate commits and the explanations in the commit messages, that really made reviewing this a breeze.

justingrant (commented):
Thanks so much @ptomato for the thorough review!  There were two things you found that should result in code changes, but both of them are wider in scope than this PR, so I split them into separate issues. 

For your other comments, I don't think changes are needed, but if you disagree please let me know!  Otherwise, I think the PR code is still good to merge unless I hear differently from @12wrigja. I'll wait a few days for James to catch up before merging.

lib/ecmascript.ts
12wrigja: I think I'd type this as `dayLengthNs: JSBI|undefined` and then cast using `!` along with a comment explaining why we know for certain it's been assigned a value there. This might force us to write this out multiple times, but I think that it's best to document these assumptions.

lib/intrinsicclass.ts
12wrigja: This is probably somewhere that I'd use `!` instead of `as`, but it sounds like there's an ESLint rule against that?

lib/timezone.ts
```diff
@@ -125,7 +125,7 @@ export class TimeZone implements Temporal.TimeZone {
       return null;
     }
 
-    let epochNanoseconds = GetSlot(startingPoint, EPOCHNANOSECONDS);
+    let epochNanoseconds: JSBI | null = GetSlot(startingPoint, EPOCHNANOSECONDS);
```
12wrigja, line 128: I'm guessing GetSlot is defined to return `JSBI` for this key, so why doesn't this slot always contain this value? Should this be changed in slots to return `JSBI|undefined`?

12wrigja (approved):

lib/calendar.ts
ptomato: Is this something that will go away when we use ES6 classes here? I ask because I find these types a bit difficult to read...

lib/plaintime.ts
ptomato: That playground link won't load for me, unfortunately (it gets stuck on "Downloading TypeScript"...) but are you referring to the function length?

justingrant: Oh, weird. Never heard of that TS Playground symptom. Anyway, here's what's in there: 

```ts
// TS code:
function f1(x: string | undefined = undefined) {}
function f2(x?: string) {}

// JS code that's emitted by the transpiler
function f1(x = undefined) { }
function f2(x) { }
```

The `= undefined` is different between the two, so I assumed it was better to match the runtime output whenever possible.

ptomato: Probably inconsequential because the length of that function is not observable, but I can see that divergences might make it harder to debug.

lib/calendar.ts
justingrant: Here's what the equivalent will look like in ES6 classes: https://github.com/js-temporal/temporal-polyfill/pull/113/files#diff-0751307e1b8622b923dfb7f18f7aed915f397cd1f6a67a5465d9672db48839f5R478

Is that better?

lib/calendar.ts
```diff
@@ -1232,10 +1353,31 @@ const nonIsoHelperBase: NonIsoHelperBase = {
...
+          max: number;
+        };
+  };
+}
+
+const helperHebrew: HelperPerCalendarImpl = ObjectAssign({}, helperSharedImpl as HelperPerCalendarImpl, {
   id: 'hebrew',
   calendarType: 'lunisolar',
-  inLeapYear(calendarDate /*, cache */) {
+  inLeapYear(calendarDate: CalendarYearOnly /*, cache: OneObjectCache */) {
```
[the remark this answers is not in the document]
justingrant: 👍  I did this in #113, so I'll plan to fix it there instead of here.

lib/calendar.ts
```diff
@@ -455,9 +447,213 @@ impl['iso8601'] = {
...
+
+/**
+ * `NonIsoImpl` - The generic top-level implementation for all non-ISO
+ * calendars. This type has the same shape as the 262-only ISO calendar
+ * implementation, which means the `Calendar` class implementation can swap out
+ * the ISO for non-ISO implementations without changing any `Calendar` code.
+ */
+interface NonIsoImpl {
+  dateFromFields(
+    this: NonIsoImplWithHelper,
```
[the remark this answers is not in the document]
justingrant: Yep!  See #113.

justingrant (commented):
I just pushed changes to resolve open issues in code review comments, so unless there are additional issues I'll plan to merge this and then finalize #113 to convert calendar.ts to ES6.

lib/calendar.ts
ptomato: I have a vague hunch that this is more complicated than it needs to be, but I'm not prepared to spend time to prove it, so I won't complain anymore :smile:

justingrant: "I have a vague hunch that this is more complicated than it needs to be" could be the motto of half of my time on this planet, I think. ;-)

lib/ecmascript.ts
justingrant: OK, I checked in @12wrigja's suggestion.

lib/intrinsicclass.ts
justingrant: Yeah. I ended up checking in a change here that's the same as the other case you commented about: turn off the ESLint rule for this line (and a few lines below in a similar loop) and add a comment explaining why.

lib/calendar.ts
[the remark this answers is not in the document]
justingrant: This type got renamed in a later commit. If you're still concerned about the interface names, let's discuss over in #113 which has a lot more renaming!

lib/timezone.ts
12wrigja: Ah gotcha - yeah I missed that it was reassigned later. I agree that this solution is fine, but certainly far less than ideal.

PR #111, opened 2021-12-15
Catch up on proposal-temporal PRs

justingrant:
Three commits here to catch up with recent PRs over in proposal-temporal:

* Throw RangeError for invalid offset strings - tc39/proposal-temporal#1976
* Remove zoneRequired param - tc39/proposal-temporal#1979
* Refactor ecmascript.ts - tc39/proposal-temporal#1976 and tc39/proposal-temporal#1980 (the latter was a subset of the lines covered by the former PR, so I just combined into one commit)

### DIFF
--- a/lib/ecmascript.ts
+++ b/lib/ecmascript.ts
@@ -315,9 +315,9 @@ function FormatCalendarAnnotation(id: string, showCalendar: Temporal.ShowCalenda
   return `[u-ca=${id}]`;
 }
 
-function ParseISODateTime(isoString: string, { zoneRequired }: { zoneRequired: boolean }) {
-  const regex = zoneRequired ? PARSE.instant : PARSE.datetime;
-  const match = regex.exec(isoString);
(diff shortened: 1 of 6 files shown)


ptomato (approved):
I think the commit message about proposal-temporal PRs 1980 & 1976 should refer to 1980 & 1975 instead. Otherwise looks good, a correct port of the proposal-temporal PRs. I'm not 100% sure about some of the TypeScript stuff, maybe James would take a look at that.

justingrant:
> I think the commit message about proposal-temporal PRs 1980 & 1976 should refer to 1980 & 1975 instead.

Oops, fixed in latest force-push.

> I'm not 100% sure about some of the TypeScript stuff, maybe James would take a look at that.

@12wrigja  is out until after the new year and I need this PR to keep moving fwd on #109, so I'll plan to merge now and @12wrigja can take a peek when he gets back.

PR #112, opened 2021-12-16
Port non-ISO calendar fixes from proposal-temporal PRs #1977 and #1983

justingrant:
Port of https://github.com/tc39/proposal-temporal/pull/1977 and https://github.com/tc39/proposal-temporal/pull/1983 to fix a bunch of smaller issues with non-ISO calendar. 

This is (hopefully!) the last ported PR so that #109 can be limited to type changes only. 

We should merge this PR only after https://github.com/tc39/proposal-temporal/pull/1977 is approved and merged.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -603,8 +603,9 @@ export namespace Temporal {
     readonly [Symbol.toStringTag]: 'Temporal.Instant';
   }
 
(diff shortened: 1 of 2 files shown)


ptomato (approved):
:+1:

PR #113, opened 2021-12-17
Refactor non-ISO calendars to use ES6 classes

justingrant:
The implementation of non-ISO calendars has always been a bit sloppy and confusing since I wrote it as a proof-of-concept a year ago. Now that non-ISO calendar implementations are fully typed in #109, it's easier and safer to refactor them into a more maintainable set of ES6 classes. The inheritance structure is below, with bold classes are actual ICU calendars, the non-bold ones are base classes. 

* HelperBase - shared code for all non-ISO calendars
  * **HebrewHelper**
  * **PersianHelper**
  * **IndianHelper**
  * GregorianBaseHelper
    * **RocHelper**
    * **BuddhistHelper**
    * **GregoryHelper**
    * **JapaneseHelper**
    * OrthodoxBaseHelper
      * **EthiopicHelper**
      * **EthioaaHelper**
      * **CopticHelper**
  * ChineseBaseHelper
    * **ChineseHelper**
    * **DangiHelper**
  * IslamicBaseHelper
    * **IslamicHelper**
    * **IslamicUmalquraHelper**
    * **IslamicTblaHelper**
    * **IslamicCivilHelper**
    * **IslamicRgsaHelper**
    * **IslamicCcHelper**

This PR removes a lot of extra TS type code that is not needed using classes where types for inherited methods is implicit. 

This is a pretty big PR but I wasn't sure how to split it into multiple commits given that things are pretty intertwined... once I started pulling on the ES6 thread, the whole sweater came with it. 😄

### DIFF
--- a/lib/calendar.ts
+++ b/lib/calendar.ts
@@ -35,10 +35,8 @@ const IntlDateTimeFormat = globalThis.Intl.DateTimeFormat;
 const ArraySort = Array.prototype.sort;
 const MathAbs = Math.abs;
 const MathFloor = Math.floor;
-const ObjectAssign = Object.assign;
 const ObjectEntries = Object.entries;
 const ObjectKeys = Object.keys;
-const ReflectApply = Reflect.apply;
 
 interface CalendarImpl {
   year(date: Temporal.PlainDate | Temporal.PlainYearMonth): number;
@@ -445,63 +443,41 @@ impl['iso8601'] = {
     if (!HasSlot(date, ISO_YEAR)) date = ES.ToTemporalDate(date);
     return ES.LeapYear(GetSlot(date, ISO_YEAR));
   }
-};
+} as CalendarImpl;
 
 // Note: Built-in calendars other than iso8601 are not part of the Temporal
 // proposal for ECMA-262. These calendars will be standardized as part of
-// ECMA-402. Code below here includes an implementation of these calendars order
-// to validate the Temporal API and to get feedback. However, non-ISO calendar
-// implementation is subject to change because these calendars are
-// implementation-defined.
+// ECMA-402. Code below here includes an implementation of these calendars to
+// validate the Temporal API and to get feedback. However, native non-ISO
+// calendar behavior is at least somewhat implementation-defined, so may not
+// match this polyfill's output exactly.
 //
 // Some ES implementations don't include ECMA 402. For this reason, it's helpful
 // to ensure a clean separation between the ISO calendar implementation which is
 // a part of ECMA 262 and the non-ISO calendar implementation which requires
 // ECMA 402.
 //
-// To ensure this separation, the implementation is split. The `NonIsoImpl`
-// interface is the top-level implementation for all non-ISO calendars. This
-// type has the same shape as the ECMA 262-only ISO calendar implementation so
-// can use the same callers, tests, etc.
-//
-// A derived interface `NonIsoImplWithHelper` adds a `helper` property that
-// includes the remaining non-ISO implementation properties and methods beyond
-// the ISO implementation above. The `helper` property's shape is a base
-// singleton object common to all calendars (`HelperSharedImpl`) that's extended
-// (interface `HelperPerCalendarImpl`) with implementation that varies for each
-// calendar.
-//
-// Typing of individual methods in the interfaces below uses the `this`
-// "parameter" declaration definition, which is a fake parameter (stripped by TS
-// during compilation and not visible at runtime) that tells TS what type `this`
-// is for a method. For historical reasons, the initial implementation of
-// non-ISO calendars mirrored the code style of the previous ISO-only
-// implementation which didn't use ES6 classes. Using the `this` parameter is a
-// hack to delay converting this file to use ES6 classes until the code was
-// fully typed to make a `class` refactoring easier and safer. We'll probably do
-// this conversion in the future. (PRs welcome!)
+// To ensure this separation, the implementation is split. A `CalendarImpl`
+// interface powers both ISO and non-ISO calendars, and that interface is
+// extended by `NonIsoImpl` which adds a `helper` property which contains
+// non-ISO implementation.
 
 /**
- * `NonIsoImpl` - The generic top-level implementation for all non-ISO
- * calendars. This type has the same shape as the 262-only ISO calendar
- * implementation, which means the `Calendar` class implementation can swap out
- * the ISO for non-ISO implementations without changing any `Calendar` code.
+ * Generic top-level implementation for all calendars. The `Calendar` class can
+ * use the same code to call ISO and non-ISO implementations.
  */
-interface NonIsoImpl {
+interface CalendarImpl {
   dateFromFields(
-    this: NonIsoImplWithHelper,
     fieldsParam: Params['dateFromFields'][0],
     options: NonNullable<Params['dateFromFields'][1]>,
     calendar: Temporal.Calendar
   ): Temporal.PlainDate;
   yearMonthFromFields(
-    this: NonIsoImplWithHelper,
     fieldsParam: Params['yearMonthFromFields'][0],
     options: NonNullable<Params['yearMonthFromFields'][1]>,
     calendar: Temporal.Calendar
   ): Temporal.PlainYearMonth;
   monthDayFromFields(
-    this: NonIsoImplWithHelper,
     fieldsParam: Params['monthDayFromFields'][0],
     options: NonNullable<Params['monthDayFromFields'][1]>,
     calendar: Temporal.Calendar
@@ -509,7 +485,6 @@ interface NonIsoImpl {
   fields(fieldsParam: string[]): Return['fields'];
   mergeFields(fields: Params['mergeFields'][0], additionalFields: Params['mergeFields'][1]): Return['mergeFields'];
   dateAdd(
-    this: NonIsoImplWithHelper,
     date: Temporal.PlainDate,
     years: number,
     months: number,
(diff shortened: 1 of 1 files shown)


lib/calendar.ts
```diff
@@ -464,52 +462,40 @@ impl['iso8601'] = {
...
 /**
  * `NonIsoImpl` - The generic top-level implementation for all non-ISO
- * calendars. This type has the same shape as the 262-only ISO calendar
- * implementation, which means the `Calendar` class implementation can swap out
+ * calendars. Other than the `helper` property which contains non-ISO-specific
+ * functionality, this type has the same shape as the 262-only ISO calendar
+ * implementation. This means the `Calendar` class implementation can swap out
  * the ISO for non-ISO implementations without changing any `Calendar` code.
  */
 interface NonIsoImpl {
```
12wrigja: Could this extend Calendar or CalendarImpl? They seem similar at first glance.

lib/calendar.ts
```diff
@@ -445,7 +443,7 @@ impl['iso8601'] = {
     if (!HasSlot(date, ISO_YEAR)) date = ES.ToTemporalDate(date);
     return ES.LeapYear(GetSlot(date, ISO_YEAR));
   }
-};
+} as Omit<NonIsoImpl, 'helper'>;
```
12wrigja: I vaguely recall being able to create these implementations using something like the following (untested):

```
impl['iso8601'] = (new (class implements CalendarImpl, Omit<NonIsoImpl, 'helper'> {
  .. impl here ...
})());
```

which might help TS better understand the inheritance / 'this' values more? unsure.

justingrant:
Oops, forgot to commit before latest force-push. Reopening!

justingrant:
FYI @ptomato and @12wrigja - Reviewing this PR is a little easier if you ignore whitespace:
image

lib/calendar.ts
```diff
@@ -1185,15 +1111,35 @@ const helperSharedImpl: HelperSharedImpl = {
...
+      weeks?: number | undefined;
+      days?: number | undefined;
+    },
+    overflow: Overflow,
+    cache: OneObjectCache
+  ): FullCalendarDate {
     const { year, month, day } = calendarDate;
-    const addedMonths = this.addMonthsCalendar({ year: year + years, month, day }, months, overflow, cache);
-    const initialDays = days + weeks * 7;
+    const addedMonths = this.addMonthsCalendar({ year: year + (years ?? 0), month, day }, months ?? 0, overflow, cache);
```
justingrant, line 1132: I believe this (and the same change in the line below) was the only runtime change found in this PR. I'll plan to port it over to proposal-temporal.

lib/calendar.ts
```diff
@@ -1852,170 +1818,170 @@ function isGregorianLeapYear(year: number) {
...
-            // last era always gets all "leftover" (older than epoch) years,
-            // so no need for a comparison like below.
-            eraYear = year - e.anchorEpoch.year + (e.hasYearZero ? 0 : 1);
-            return true;
-          }
-          const comparison = helperSharedImpl.compareCalendarDates(adjustedCalendarDate, e.anchorEpoch);
-          if (comparison >= 0) {
-            eraYear = year - e.anchorEpoch.year + (e.hasYearZero ? 0 : 1);
+abstract class GregorianBaseHelper extends HelperBase {
+  id: BuiltinCalendarId;
```
12wrigja, line 1822: Doesn't this need to be abstract to force implementations to specify it?

lib/calendar.ts
```diff
@@ -445,71 +443,48 @@ impl['iso8601'] = {
     if (!HasSlot(date, ISO_YEAR)) date = ES.ToTemporalDate(date);
     return ES.LeapYear(GetSlot(date, ISO_YEAR));
   }
-};
+} as CalendarImpl;
```
12wrigja, line 446: Could this instead be something like

```
impl['iso8601'] = new (class implements CalendarImpl {
 .. impl content here ...
 })();
```

which avoids casting and aligns better with the helper registration?

lib/calendar.ts
justingrant: TS wasn't happy about making these props `abstract`, because of this error: 

> Abstract property 'id' in class 'GregorianBaseHelper' cannot be accessed in the constructor.

I guess if properties are assigned in the constructor, then they can't be abstract.

12wrigja: ah - I didn't notice the duplication there. Would it be worth making it a readonly constructor property instead of specifying it twice?

```
constructor(readonly id: BuildinCalendarId, originalEras: InputEra[]) {
	super();
	// no need for this.id = id;
	// or adding the property
}
```

justingrant: If all the constructor params were setting properties without code, sure that'd work. But the `originalEras` parameter is split into two properties by the `adjustEras` function.  I'm not sure it makes things any more readable to only use the `readonly` trick for `id` but then run code for the era-related properties. Seems clearer to me to just explicitly show what's happening. What do you think?

lib/calendar.ts
justingrant: Actually, the cast is unnecessary because the `impl` array is typed as `Record<BuiltinCalendarId, CalendarImpl>`. I removed it; see the latest commit.

re: making it a class, the problem is that TS's `class` syntax doesn't automatically type class member functions. So I'd need to add types to every parameter of every method, and after all that work we'd end up with exactly the same level of type safety as the current code. Also, it'd have quite a different runtime emit compared to proposal-temporal.

I've observed while refactoring this file that TS `class` is great when you want to inherit functionality or create multiple instances. But when creating non-inherited singletons (like this object here), it seems like `class` adds work without adding much value because class members don't snap to interface types.

So I'm not sure how best to proceed. Seems like there are these options: 
1) Leave as-is
2) Convert both the ISO and non-ISO `XxxImpl` instances to anonymous classes per your suggestion above, but otherwise leave as-is.
3) Convert `nonIsoHelper` from a cloned singleton into an abstract base class and remove the "helper" concept because each non-ISO calendar would have its own class derived from NonIsoImpl. The original reason for the current split between `XxxImpl` and `helper` was to ensure that the non-ISO implementation was the same shape as the  ISO one. I wanted it to be really obvious when the NonIsoImpl object was calling into a method that didn't exist in ISO. But we don't need to stick with this architecture if we don't think it's helpful.

My preference is for (1) because it's easiest and involves the least deviation from existing proposal-temporal code. But I don't have a very strong opinion either way.  What do you think?

justingrant: > because the `impl` array is typed as `Record<BuiltinCalendarId, CalendarImpl>`. I removed it; see the latest commit.

Actually, I refined the type of the impl array further so that the record is CalendarImpl for the ISO calendar and NonIsoImpl for other calendars. More precise.

lib/calendar.ts
ptomato: It seemed to me like the previous code was actually better! (Besides the fact of not being typed, that is...)
I guess if we were to do `{ years = 0 }: { years?: number | undefined }` TypeScript is not smart enough to figure that the `years` that you get in the body of the function can only be a `number`?

By the way, do we need `years?: number | undefined` rather than just `years?: number`? It seems like the question mark operator should cover that.

lib/calendar.ts
```diff
@@ -1509,69 +1496,92 @@ const helperHebrew: HelperPerCalendarImpl = ObjectAssign({}, helperSharedImpl as
...
-});
+}
 
-const helperPersian = ObjectAssign({}, helperSharedImpl, {
-  id: 'persian',
-  calendarType: 'solar',
-  inLeapYear(calendarDate: CalendarYearOnly, cache?: OneObjectCache) {
+// There are 6 Islamic calendars with the same implementation in this polyfill.
+// They vary only in their ID. They do emit different output from the underlying
+// Intl implementation, but our code for each of them is identical.
```
ptomato, line 1538: I didn't realize this! I guess in a correct implementation they would have to have different Hijri correction days?

ptomato (approved):
Big improvement!

lib/calendar.ts
justingrant: I don't know exactly why they return different results, but they do return different results. My naive understanding of Islamic calendars is that the months change based on when the new moon happens, and "when the new moon happens" depends on who is doing the observing and where those observations are happening.

Furthermore, there is a "tabular" calendar which I believe means that they have some sort of mathematical formulae to determine the lunar cycles, vs. an actual observational mechanism to determine the month boundaries. https://en.wikipedia.org/wiki/Tabular_Islamic_calendar

For Temporal's purposes, I don't think we really care much about why these are different, only that they are different and we can't treat them all the same.

lib/calendar.ts
justingrant: You're exactly right. I think what happened here was that an intermediate state of this class had an externally-declared `interface` and I couldn't figure out how to get the `= 0` defaults to work with that external declaration. But later I removed that declaration it for DRY reasons (now just relying on the class itself) but forgot to revert the default params changes.  Latest commit restores the old code which I agree was better. Thanks for catching this!

lib/calendar.ts
justingrant: After thinking about it overnight, this PR is already pretty big so I think the best thing would be to merge this as-is, and then consider (2) or (3) above in a separate PR which would be smaller in scope so would be easier to review. I opened #119 to discuss, and I'll resolve this discussion now.

lib/calendar.ts
justingrant: This is the last outstanding issue before merging and I don't think there's a clear obvious answer either way, so I'm planning to leave as-is for now. Happy to revisit later.

PR #116, opened 2022-01-02
GetFormatterParts tests and error cases

bananer:
As a follow-up to #97 this adds tests for the relevant code parts and lets it throw on unexpected input (instead of failing with NaN values later somewhere else)

Should a browser change the format again, it will fail with a helpful error message in the correct place.

### DIFF
--- a/lib/ecmascript.ts
+++ b/lib/ecmascript.ts
@@ -2861,16 +2861,42 @@ export function GetIANATimeZonePreviousTransition(epochNanoseconds: JSBI, id: st
   return result;
 }
 
+// ts-prune-ignore-next TODO: remove this after tests are converted to TS
 export function parseFromEnUsFormat(datetime: string) {
-  const [month, day, year, era, hour, minute, second] = datetime.split(/[^\w]+/);
-  return {
-    year: era.toUpperCase().startsWith('B') ? -year + 1 : +year,
-    month: +month,
-    day: +day,
-    hour: hour === '24' ? 0 : +hour, // bugs.chromium.org/p/chromium/issues/detail?id=1045791
-    minute: +minute,
-    second: +second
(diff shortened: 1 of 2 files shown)


lib/ecmascript.ts
```diff
@@ -2861,20 +2861,50 @@ export function GetIANATimeZonePreviousTransition(epochNanoseconds: JSBI, id: st
...
 export function GetFormatterParts(timeZone: string, epochMilliseconds: number) {
   const formatter = getIntlDateTimeFormatEnUsForTimeZone(timeZone);
   // Using `format` instead of `formatToParts` for compatibility with older clients
   const datetime = formatter.format(new Date(epochMilliseconds));
-  const [month, day, year, era, hour, minute, second] = datetime.split(/[^\w]+/);
-  return {
-    year: era.toUpperCase().startsWith('B') ? -year + 1 : +year,
-    month: +month,
-    day: +day,
-    hour: hour === '24' ? 0 : +hour, // bugs.chromium.org/p/chromium/issues/detail?id=1045791
```
ptomato, line 2874: It'd be nice to preserve this link in a comment in the new code, so that we know why the check for 24 is present.

ptomato (approved):
Thanks for the follow-up, :+1: from me. (Personally I still wish we could use formatToParts here, but I get that we should probably do that later when there is wider adoption.)

lib/ecmascript.ts
bananer: Thanks for the review, I added that comment back.

justingrant (approved):
Thanks for this PR!

justingrant:
Oops. forgot to squash this into one commit before merging. Sorry for the GIT timeline noise, future readers!

PR #121, opened 2022-01-25
Include the valueOf function in Temporal.Duration.

12wrigja:
This was missed out in https://github.com/js-temporal/temporal-polyfill/commit/46b663cfe40c6b9381263d166615f8d5f7fadef5.


ptomato (approved):
My bad!

12wrigja:
Absolutely no problem :) At first I thought maybe they were all missing, but it turns out only this one was.

PR #122, opened 2022-01-26
Change type-only imports to use import type

### LINKED ISSUE #117: Change type-only imports to use `import type`
There are a few places where type-only imports use `import`:
* These two imports in intl.ts
```ts
import { Temporal, Intl } from '..';
import { DateTimeFormatParams as Params, DateTimeFormatReturn as Return } from './internaltypes';
```
* `import { Temporal } from '..';` in almost every code file

When imports are only types, it's probably best to use `import type` instead of `import`. See https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export for context. 

It's optional, but can't hurt and may help in some cases with bundlers, babel, etc.

NotWoods:
Closes  #117

Updates type-only imports in the TypeScript source code, and turns on a typescript rule to show an error if any new imports in the old format are introduced.

### DIFF
--- a/lib/calendar.ts
+++ b/lib/calendar.ts
@@ -21,7 +21,7 @@ import {
   HasSlot,
   SetSlot
 } from './slots';
-import { Temporal } from '..';
+import type { Temporal } from '..';
 import type {
   BuiltinCalendarId,
   CalendarParams as Params,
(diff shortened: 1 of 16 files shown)


ptomato (approved):
Thanks! This looks correct to me.

12wrigja (approved):

Issue #124, opened 2022-01-30
DateTimeFormatOptions typing override missing calendar


RandomEngy:
The TC39 docs [have this example](https://tc39.es/proposal-temporal/docs/plainmonthday.html#toLocaleString):

```js
monthDay.toLocaleString(locales, { calendar: monthDay.calendar });
```

But it doesn't compile for me with the polyfill library unless I do this:

```js
monthDay.toLocaleString(locales, { calendar: monthDay.calendar.id });
```

The `options` parameter takes `calendar: string | undefined` while `monthDay.calendar` is a `Temporal.CalendarProtocol`. I thought the example code might have been incorrect, but it turns out that [if you pass in a `Temporal.Calendar` there it is interpreted correctly](https://github.com/tc39/proposal-temporal/pull/2023).

The polyfill library doctors the Intl.DateTimeFormatOptions. It looks like it redeclares timeZone, but I think also should be redeclaring the calendar property to also accept `Temporal.CalendarProtocol`.

```typescript
  export interface DateTimeFormatOptions extends Omit<globalThis.Intl.DateTimeFormatOptions, 'timeZone'> {
    timeZone?: string | Temporal.TimeZoneProtocol;
    // TODO: remove the props below after TS lib declarations are updated
    dayPeriod?: 'narrow' | 'short' | 'long';
    dateStyle?: 'full' | 'long' | 'medium' | 'short';
    timeStyle?: 'full' | 'long' | 'medium' | 'short';
  }
```

justingrant:
Good catch. Fix coming soon.

Issue #118, opened 2022-01-08
Update type declarations for "Timezone-like" and "Calendar-like" parameters to support ZDT property bags


justingrant:
The current public types for `TimeZone.from` look like this:

https://github.com/js-temporal/temporal-polyfill/blob/e046ccdfd5ff4a2f048195894e279aefc64e565c/index.d.ts#L1125

But the spec and the polyfill also accept a ZDT-like property bag too, per tc39/proposal-temporal#925. Specifically, they accept any object with a `timeZone` property and will use that property's value as the time zone, with one exception: per https://github.com/tc39/proposal-temporal/issues/925#issuecomment-709544520, recursive bags like `{timeZone: {timeZone: 'Asia/Tokyo'}}` are disallowed. 

Therefore, the declaration should also allow a `{ timeZone: TimeZoneProtocol | string }` parameter type.

This issue tracks the work to: 
1. Figure out how to encode the "no recursive bags" in TS types. One possible way (not sure this will work) could be something like this:
```ts
type NoRecursiveTimeZones = { timeZone: { timeZone: never } };
. . .
  static from(timeZone: TimeZoneProtocol | string | { timeZone: TimeZoneProtocol } | NoRecursiveTimeZones): Temporal.TimeZone | TimeZoneProtocol; 
```
2. Use that declaration throughout index.d.ts wherever TimeZone-like parameters are accepted
3. Port the changes to proposal-temporal, and verify that the docs are correct for methods that accept a time zone parameter.
4. Remove any now-unnecessary type casts, like the one @ptomato found in in https://github.com/js-temporal/temporal-polyfill/pull/109/files#r779978091.

justingrant:
Note that the same issue applies to calendars too. Updating title accordingly.

PR #126, opened 2022-02-04
Update TS types of TimeZone-like/Calendar-like params to match spec and polyfill

justingrant:
Extends TS tsupport for property-bag form of Calendar-like and TimeZone-like params in Temporal methods. Fixes https://github.com/js-temporal/temporal-polyfill/issues/118

Also allows `Temporal.Calendar` instances to be used in the `Intl.DateTimeFormatOptions` type. (Previously only string IDs were allowed.) Fixes https://github.com/js-temporal/temporal-polyfill/issues/124.

Note that fixing https://github.com/tc39/proposal-temporal/issues/2005 will further extend the types that can be passed into the `calendar` and `timeZone` properties of `Intl.DateTimeFormatOptions`. But that requires a spec change and polyfill code change, so in this PR we'll only extend `Intl.DateTimeFormatOptions` to what's supported by the current spec and polyfill.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -189,7 +189,7 @@ export namespace Temporal {
 
   export type InstantToStringOptions = Partial<
     ToStringPrecisionOptions & {
-      timeZone: TimeZoneProtocol | string;
+      timeZone: TimeZoneLike;
     }
   >;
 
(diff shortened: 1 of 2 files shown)


ptomato (approved):
LGTM, a couple of questions though:
- #118 mentions removing a cast: https://github.com/js-temporal/temporal-polyfill/pull/109/files#r779978091, should that be part of this PR?
- does this encode the "no recursive property bags" rule or is that coming in a follow-up?

justingrant:
> * #118 mentions removing a cast: https://github.com/js-temporal/temporal-polyfill/pull/109/files#r779978091, should that be part of this PR?

Good catch!  Thanks for remembering that. Fixed now in latest commit. (Both the timezone and the calendar cast could be removed.)

> * does this encode the "no recursive property bags" rule or is that coming in a follow-up?

I couldn't figure out how to encode that in TS types. Tried a few ways to do it, none of them seemed to work. So leaving that for someone with better TypeScript Fu than me.

ptomato:
> leaving that for someone with better TypeScript Fu than me.

Well, certainly not me. I salute whoever it is!

justingrant:
> > * does this encode the "no recursive property bags" rule or is that coming in a follow-up?
> 
> I couldn't figure out how to encode that in TS types. Tried a few ways to do it, none of them seemed to work. So leaving that for someone with better TypeScript Fu than me.

Actually, I take that back. If an object bag is used, then its `timeZone` or `calendar` property, respectively, must be `string | TimeZoneProtocol` or `string | CalendarProtocol`.  The two protocol types already have `timeZone?: never` and `calendar?: never` properties defined which should disallow recursion. It's not perfect; a `timeZone: undefined` property will pass TS, but any non-undefined value will cause a compile error. I tried making those `never` properties non-optional, but this broke other code so I figured we can just take the W with the current types.

PR #130, opened 2022-02-05
Remove duplicate CalendarImpl definition.

12wrigja:
I think this got copied as part of a previous calendar refactoring and the original never cleaned up.


justingrant (approved):
Whoops!  Thanks for catching this.

PR #131, opened 2022-02-05
Enable noImplicitOverride tsc flag.

12wrigja:
Also sort the keys in tsconfig.json to make it easier to read and diff between codebases.


justingrant (approved):
Cool!  The overrides make things clearer.

justingrant:
Also, looks like this needs a rebase for tsconfig.json

ptomato (approved):

tsconfig.json
```diff
@@ -4,29 +4,29 @@
     "allowUnreachableCode": false,
     "downlevelIteration": true,
     "importHelpers": true,
-    "sourceMap": true,
+    "importsNotUsedAsValues": "error",
     "inlineSources": false,
+    "lib": ["dom", "ES2020"],
```
justingrant, line 9: Hey James - Sorry for not catching this earlier. Does this config mean that the polyfill won't support downlevel clients that don't have ES2020?

12wrigja: No - this is just the tsconfig. This output is not published to NPM untouched - it passes through Babel first. We downlevel some of the artifacts published to NPM using Babel: https://github.com/js-temporal/temporal-polyfill/blob/d8dfacb47a562c127156bb53f2790e66d973181d/rollup.config.js#L147

12wrigja: And this particular line was already present in tsconfig (see L19 on the "before" side).

justingrant: OK cool.

PR #134, opened 2022-02-08
Fix `TimeZone.p.getXxxTransition()` worst-case perf

### LINKED ISSUE #60: TimeZone GetNextTransition / GetPreviousTransition for far-future and far-past dates
The current implementation of TimeZone's `GetNextTransition` / `GetPreviousTransition` methods behave poorly for very large or very small inputs, including possible denial-of-service issues caused by O(1) loops of expensive code. 

Examples: 
```js 
// Repro 1: far-future date, get previous transition
Temporal.TimeZone.from('Etc/GMT+10').getPreviousTransition(new Temporal.Instant(0n).add({hours: 24 * 1e8}));
  // expected: ???
  // actual: runs for a loooooooong time, probably over an hour (denial of service attack?) then will return null

// Repro 2: far-past date, get next transition
Temporal.TimeZone.from('Etc/GMT+10').getNextTransition(new Temporal.Instant(0n).add({hours: -24 * (1e8-1)}));
  // expected: ???
  // actual: runs for a loooooooong time, probably over an hour (denial of service attack?) then will return null

// Repro 3: future date in year 3021, get next transition
Temporal.TimeZone.from('Etc/GMT+10').getNextTransition(new Temporal.Instant(0n).add({hours: 24 * 365 * 1000}));
  // expected: ???
  // actual: null
```

There are a few issues here: 
1. Can we prevent a denial-of-service attack caused by this expensive, O(1) loop?
2. DST looking forward more than a few decades in the future seems inherently unreliable, and certainly not worth exposing the user's computer to a DoS opportunity.

Should we simply limit future dates to N years in the future (N = 100? N = 50?  N = 20?), and throw a RangeError for anything past that?

BTW, we already limit the problem for too-small dates (There was no DST before 1847 C.E.) when calling `GetPreviousTransition`.  We should use the same smarts in `GetNextTransition` to avoid Repro #2 above.

justingrant:
When passed very far-past or very far-future dates,`TimeZone.p.get(Next|Previous)Transition()` currently takes minutes of 100% CPU to finish because those methods include a loop that makes expensive system calls for every 2 weeks until it finds a transition. For periods where there are no transitions, this loop will run millions of times which is time consuming.

Note that native implementations have access to the TZDB data directly so this is not an issue there. We could in theory bundle TZDB data (or simply bundle a shorter list of per-timezone data showing the earliest and oldest TZDB rules for each) but doing so would add a lot of complexity (and at least some bundle size) to this polyfill for minimal benefit, so solving the problem with optimization is, IMHO, preferable.  

This PR improves worst-case perf ~1500x by:
* Skipping looping for any dates before 1847 CE, which is the first entry in the TZDB.
* Optimizing handling of dates 10 years or later than the current system time. Offset changes are only planned a few years in advance, so when provided a far-future date, we only need to check one year away to see if there's DST projected forward from the most recent rule in that TZ. If there's not, then when calling `getNextTransition` we know there's no transitions (and can return `null`) and when calling `getPreviousTransition` we can skip all the way back to current time + 10 years because we know that no transitions will be in the skipped period.

These optimizations reduce a worst-case loop of 273K years to 175 years, or about 1500x faster.  That translates to a worst-case call time of ~80ms on my older MacBook Pro, which is still not great but is probably good enough for now given this is a corner case of an unusual API.

Fixes #60.

### DIFF
--- a/test/timezone.mjs
+++ b/test/timezone.mjs
@@ -217,6 +217,114 @@ describe('TimeZone', () => {
       }
     });
   });
+  const checkTime = (limitMsecs, func) => {
+    const now = Date.now();
+    func();
+    const msecs = Date.now() - now;
+    if (msecs > limitMsecs) assert(false, `Expected ${limitMsecs}ms or less, actual: ${msecs}ms`);
+  };
+  describe('Far-future transitions (time zone currently has DST)', () => {
+    const zone = new Temporal.TimeZone('America/Los_Angeles');
+    const inst = Temporal.Instant.from('+200000-01-01T00:00-08:00');
+    it('next transition is valid', () => {
+      const nextTransition = zone.getNextTransition(inst, zone);
+      const zdtTransition = nextTransition.toZonedDateTimeISO(zone);
+      equal(zdtTransition.offset, '-07:00');
+      equal(zdtTransition.month, 3);
+      equal(zdtTransition.subtract({ nanoseconds: 1 }).offset, '-08:00');
+    });
+    it('getNextTransition takes less than 800ms', () => {
+      checkTime(800, () => zone.getNextTransition(inst, zone));
+    });
+    it('previous transition is valid', () => {
+      const prevTransition = zone.getPreviousTransition(inst, zone);
+      const zdtTransition = prevTransition.toZonedDateTimeISO(zone);
+      equal(zdtTransition.offset, '-08:00');
+      equal(zdtTransition.month, 11);
+      equal(zdtTransition.subtract({ nanoseconds: 1 }).offset, '-07:00');
+    });
+    it('getPreviousTransition takes less than 800ms', () => {
+      checkTime(800, () => zone.getPreviousTransition(inst, zone));
+    });
+  });
+  describe('Far-future transitions (time zone has no DST now, but has past transitions)', () => {
+    const zone = new Temporal.TimeZone('Asia/Kolkata');
+    const inst = Temporal.Instant.from('+200000-01-01T00:00+05:30');
+    it('next transition is valid', () => {
+      const nextTransition = zone.getNextTransition(inst, zone);
+      equal(nextTransition, null);
+    });
+    it('getNextTransition takes less than 800ms', () => {
+      checkTime(800, () => zone.getNextTransition(inst, zone));
+    });
+    it('previous transition is valid', () => {
+      const prevTransition = zone.getPreviousTransition(inst, zone);
+      const zdtTransition = prevTransition.toZonedDateTimeISO(zone);
+      equal(zdtTransition.offset, '+05:30');
+      equal(prevTransition.toString(), '1945-10-14T17:30:00Z');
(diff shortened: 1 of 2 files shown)


justingrant:
Given that this is a perf PR, I figured it'd make sense to add actual perf tests for the changes.  Actual worst-case call perf is about 80ms on my older MacBook Pro, so just to be safe in CI I bumped it up to 800ms which hopefully should be enough for slow CI systems.

If this kind of test is a bad idea, then I'm happy to revert.

ptomato (approved):
This seems clever!

I'm personally not a fan of the "must take less than 800 ms" tests, but I'll defer to you and James on that. If I were setting it up, I'd probably want to have it run on PRs only, and compare the performance before and after the change.

I'd still love to see an approach based on https://github.com/tc39/proposal-temporal/pull/1370 when the time is right!

test/timezone.mjs
```diff
@@ -217,6 +217,114 @@ describe('TimeZone', () => {
...
+      const zdtTransition = prevTransition.toZonedDateTimeISO(zone);
+      equal(zdtTransition.offset, '+05:30');
+      equal(prevTransition.toString(), '1945-10-14T17:30:00Z');
+      equal(zdtTransition.subtract({ nanoseconds: 1 }).offset, '+06:30');
+    });
+    it('getPreviousTransition takes less than 800ms', () => {
+      checkTime(800, () => zone.getPreviousTransition(inst, zone));
+    });
+  });
+  describe('Far-future transitions (time zone has never had any offset transitions', () => {
```
12wrigja, line 271: missing close parens

test/timezone.mjs
```diff
@@ -217,6 +217,114 @@ describe('TimeZone', () => {
...
+    });
+    it('previous transition is valid', () => {
+      const prevTransition = zone.getPreviousTransition(inst, zone);
+      equal(prevTransition, null);
+    });
+    it('getPreviousTransition takes less than 800ms', () => {
+      checkTime(800, () => zone.getPreviousTransition(inst, zone));
+    });
+  });
+  describe('Far-past transitions (time zone has never had any offset transitions', () => {
```
12wrigja, line 310: same

12wrigja (commented):
It would be interesting to be able to measure this in terms of the number of operations happening - I wonder if the JSBI static functions could be spied on and called through, and all the spies call counts summed up to give a useful measurement.

That being said, I have no qualms with checking in your current tests, and if they end up being flaky we can always disable them.

12wrigja (approved):

justingrant:
> It would be interesting to be able to measure this in terms of the number of operations happening - I wonder if the JSBI static functions could be spied on and called through, and all the spies call counts summed up to give a useful measurement.

Seems reasonable to investigate, but I won't have time to do it!  😄 

> That being said, I have no qualms with checking in your current tests, and if they end up being flaky we can always disable them.

Sounds good. After fixing the typos you found then I'll merge this PR, and we can always remove the tests if they're problematic later.

PR #136, opened 2022-02-12
Fix `Calendar.fields` return type (in index.d.ts) to match the spec

justingrant:
Port of https://github.com/tc39/proposal-temporal/pull/2056


12wrigja (approved):

justingrant:
> The CalendarImpl change seems fine regardless. The other one, I'm not sure about, see comment on [tc39/proposal-temporal#2056](https://github.com/tc39/proposal-temporal/pull/2056).

I synced up the content of this PR with the conclusion of [tc39/proposal-temporal#2056](https://github.com/tc39/proposal-temporal/pull/2056) so we should be good to go.

PR #144, opened 2022-03-17
Remove erroneous options from Safari's Intl.DateTimeFormat resolvedOptions.

12wrigja:
This causes Test262 errors in Safari for
ZonedDateTime/prototype/toLocaleString/options-conflict (and possibly
others).

### DIFF
--- a/lib/intl.ts
+++ b/lib/intl.ts
@@ -208,11 +208,30 @@ function adjustFormatterTimeZone(
   if (!timeZone) return formatter;
   const options = formatter.resolvedOptions();
   if (options.timeZone === timeZone) return formatter;
-  // Existing Intl isn't typed to accept Temporal-specific options, but will not
-  // break at runtime if we pass them. Also, the lib types for resolved options
-  // are less restrictive than the types for options. For example, `weekday` is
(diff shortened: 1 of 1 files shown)


12wrigja:
Note: I only have very indirect access to Safari via Google's testing infrastructure (I'm not even sure what version is being used :|), and the contents of the PR does fix test failures there. I would appreciate if someone else with access to Safari would validate whether these changes are needed for newer Safari versions:

```
new Intl.DateTimeFormat('en', {'dateStyle': 'long'}).resolvedOptions();
``` 

This object should not contain keys that conflict as per https://tc39.es/proposal-intl-datetime-style/#table-datetimeformat-components, but in some version of Safari appears to do so.

ptomato:
This looked very familiar and indeed I already reported it to Safari: https://bugs.webkit.org/show_bug.cgi?id=231041
So I guess it should work correctly in Safari versions released after October 2021.

It seems like it'd be good to carry this workaround until the broken versions are mostly obsolete, though I don't know if we've ever thought about an explicit policy on carrying workarounds...

ptomato (approved):

PR #153, opened 2022-05-08
Lunisolar math fixes

### LINKED ISSUE #149: Oddness in year arithmetic with lunisolar calendars
I noticed something odd about adding and subtracting years in a lunisolar calendar:
```js
const date = Temporal.PlainDate.from({ year: 5782, monthCode: 'M08', day: 2, calendar: 'hebrew' });
console.log(date.toLocaleString('en-u-ca-hebrew'));  // => 2 Iyar 5782
console.log(date.add({years: 1}).toLocaleString('en-u-ca-hebrew'));  // => 2 Sivan 5783
```
It seems like the ordinal month is being preserved, but it seems more likely that the expectation would be for the month code to be preserved. (5782 is a leap year in the Hebrew calendar, so Iyar is month 9. Next year, Sivan is month 9 and Iyar is month 8.) But then again, it seems like this could be dependent on the custom of whoever uses each calendar.

justingrant:
This PR fixes two bugs with lunisolar calendars: 

1. https://github.com/js-temporal/temporal-polyfill/commit/932810be1306a30d4c95fc88c54c64abe7fb1bf9 fixes #149 which could happen when adding/subtracting years to/from a lunisolar calendar date.
2. https://github.com/js-temporal/temporal-polyfill/commit/80611d584b880d7c9af68327b49bdc9ede050eed fixes another bug in the Hebrew calendar that was exposed by new tests for (1)

The tests for (2) can't pass without (1), so it was most convenient to put the fixes into separate commits in the same PR. Hope this is OK, given that each of these are tiny changes. If not, I could split into two PRs.

### DIFF
--- a/lib/calendar.ts
+++ b/lib/calendar.ts
@@ -1423,7 +1423,7 @@ class HebrewHelper extends HelperBase {
         } else {
           month = monthCodeNumberPart(monthCode as string);
           // if leap month is before this one, the month index is one more than the month code
-          if (this.inLeapYear({ year }) && month > 6) month++;
+          if (this.inLeapYear({ year }) && month >= 6) month++;
           const largestMonth = this.monthsInYear({ year });
           if (month < 1 || month > largestMonth) throw new RangeError(`Invalid monthCode: ${monthCode}`);
         }
--- a/lib/calendar.ts
+++ b/lib/calendar.ts
@@ -1106,13 +1106,14 @@ abstract class HelperBase {
     return calendarDate;
   }
   addCalendar(
-    calendarDate: CalendarYMD,
+    calendarDate: CalendarYMD & { monthCode: string },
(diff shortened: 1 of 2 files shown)


lib/calendar.ts
```diff
@@ -1106,13 +1106,14 @@ abstract class HelperBase {
     return calendarDate;
   }
   addCalendar(
-    calendarDate: CalendarYMD,
+    calendarDate: CalendarYMD & { monthCode: string },
```
12wrigja, line 1109: Nit: is `{monthCode: string}` something that would benefit from being a named type and reused elsewhere?

12wrigja (approved):
LG for TS code style.

ptomato (approved):
Makes sense. I'm a bit fuzzy on the calendar calculations though, so I didn't look too closely.

PR #158, opened 2022-05-31
Add types to exports/./import to support Typescript Node16

jonasb:
Before this Typescript would fail to resolve the types for the package when configured to use module: "Node16" (new in Typescript 4.7).

`main.ts:1:26 - error TS7016: Could not find a declaration file for module '@js-temporal/polyfill'.`

Repro available at https://github.com/jonasb/typescript-error-repro/tree/typescript-4.7-esm-node16-js-temporal

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,10 @@
   "exports": {
     ".": [
       {
-        "import": "./dist/index.esm.js",
+        "import": {
+          "types": "./index.d.ts",
+          "default": "./dist/index.esm.js"
+        },
         "default": "./dist/index.cjs"
       },
       "./dist/index.cjs"


12wrigja:
Thanks for the PR, and for creating a repro.

Looking around the community, this seems like it's unfortunately intentional (https://github.com/microsoft/TypeScript/issues/46334#issuecomment-981876926 implies that if there is an exports object that a `types` key is now required in the exports object instead of falling back to the top-level types). Seems similar to https://github.com/framer/motion/issues/1543.

I patched your PR and tested against a couple small test projects and they seem to compile/run just fine, so I think this is fine to merge.

12wrigja (approved):

everett1992:
I hit  `Could not find a declaration file for module '@js-temporal/polyfill'` today and this change fixes the error. When will this be published to npm?

12wrigja:
I've been in the middle of merging a bunch of upstream fixes, but that might take longer to release so I'll put together a patch release with this fix sometime in the next day or so.

12wrigja:
@jonasb  it seems this might not have fixed things, based on the report in #177 ? Can you verify that this PR did fix things for your repro?

everett1992:
I can confirm it's fixed for my use case.

PR #162, opened 2022-06-23
Fix mathematical error when computing System UTC Epoch nanoseconds.

12wrigja:
This bug was introduced in a25a878665dc7b73e101f69d5b127a5439595653 as a
typo - division drops the last 6 digits of the current milliseconds,
when instead we wanted to use them.

This was causing Temporal/Now/instant/return-value-value to be flaky in my testing, as sometimes time appeared to go backward by a small number of nanoseconds.


ptomato (approved):

ptomato:
Good catch!

(Separately, I think we should probably remove the fake nanoseconds eventually, since this is intended for production. There's probably some rollup transform that will give us `process.hrtime` in Node and a lower-precision equivalent in the browser.)

PR #163, opened 2022-06-28
Port various changes from tc39/proposal-temporal polyfill

12wrigja:
Unlike other PR's, I intend to merge this PR without squashing, to retain all the commit information for better history.


ptomato (approved):
The way you preserved the commits looks really useful!

I didn't review it exhaustively, but LGTM.

It seems like it's not strictly necessary to drop the old tests when we pull in new test262 tests, but it probably doesn't hurt either. You might also consider squashing the commits that only drop old tests and import new test262 tests, into one commit at the end, since it looks like at least half the commits here are comprised of those. But this is fine too.

justingrant:
IMO, I don't think we should delete the old tests until we have a faster way of running Test262. 

Otherwise, looks good. 

Have there been other commits since this PR was opened that should be included?

12wrigja:
There are other commits to cherry-pick in after this, but this was a good stopping point. I'm going through the history of new commits to upstream sequentially, and ended up blocked on trying to get one of the later commits to type-check properly. I do intend to send more PRs to try and catch up to the upstream repo's state. I've also created a small handful of scripts to help accelerate that process and enable verification that our rebased commits match their corresponding upstream commits (hence the UPSTREAM_COMMIT tag in the descriptions) which I'll tidy up and commit here along with some general docs on how to rebase upstream commits so that others can help do this sort of maintenance work.

I do generally wish that 262 was faster, but having duplicated tests that then fail as changes are being cherry-picked in becomes very time-consuming (having to determine if the failure is expected or not / whether there were changes missed in the upstream test file, or to the upstream 262 test, etc).

justingrant:
> I do generally wish that 262 was faster, but having duplicated tests that then fail as changes are being cherry-picked in becomes very time-consuming (having to determine if the failure is expected or not / whether there were changes missed in the upstream test file, or to the upstream 262 test, etc).

Makes sense. Are you suggesting that we should also remove older demitasse tests that were previously removed months ago, or only for new commits?

BTW, I just spent a few mins trying to speed up Test262 on my computer (it's a new M1 Max MacBook Pro) and I found that the following changes sped it up by about 35%:
* On MacOS, double the threads (more than 2x didn't seem to help)
* Increase max thread limit to 32 (my laptop reports 10 CPUs=>20 threads, but there are higher-end CPUs out there)

Should I PR those changes?

```sh
if [ "$(uname)" = 'Darwin' ]; then
  threads=$(sysctl -n hw.logicalcpu)
  ((threads = threads * 2))
else
  threads=$(nproc --ignore 1)
fi
if [ $threads -gt 32 ]; then threads=32; fi
```

12wrigja:
> Are you suggesting that we should also remove older demitasse tests that were previously removed months ago, or only for new commits?

I imagine at some point it would make sense, but for now my strategy has been:
 - if a PR upstream changes a test, accept it's changes (done manually)
 - run tests after each ported commit. If tests fail, investigate. This has lead to some tests being deleted, discovering of a "missing" PR, etc.
 
 I haven't done a diff between the two corresponding branch HEADs to see what is different and merge all the changes, but eventually maybe this will be the case. 

Eventually, the only functionality I'd expect to exist in non-test262 tests are going to be implementation-detail-specific tests.

 > Should I PR those changes?
 
 Sure! I don't use any Apple hardware so I don't think this will apply to me, but if it increases your velocity go for it. It might also be worth filing a tracking bug to consider using a V8 snapshot to speed up the initialization time each node execution takes. That being said, the strategy I've been looking at is trying to eliminate the need to restart the VM in between tests, at the cost of potentially being inaccurate in the case something is changed that can't be restored (e.g. module-local state, etc).

justingrant:
>  I don't use any Apple hardware so I don't think this will apply to me, but if it increases your velocity go for it.

Done. https://github.com/js-temporal/temporal-polyfill/pull/165

Interestingly, increasing threads also boosts CI performance 30%-40%, so seems like a good idea to do in general. It might be worth trying the same trick in your local dev environment to see if it helps there too. 

>  It might also be worth filing a tracking bug to consider using a V8 snapshot to speed up the initialization time each node execution takes. 

Do you want to file this?  I'm not familiar with the snapshot feature so I think you might be able to file a clearer issue than I could. Thanks!

ptomato:
Has the snapshot feature landed in a released Node.js yet?

PR #165, opened 2022-07-07
Speed up Test262 by using more threads

justingrant:
On my Mac (2021 M1 Max 10-core MBP), setting threads to 2x the number of
CPU cores resulted in a 35%-40% speedup when running Test262. More
than 2x didn't seem to help.

I'll test this to see if it improves CI performance too, and if not
then I'll revise this commit to only make this change on MacOS.


justingrant:
FWIW, the latest test run (a few hours after my previous runs) still shows good speedup. 
Previous: https://github.com/js-temporal/temporal-polyfill/runs/7143762564 - 10m 52s
This PR: https://github.com/js-temporal/temporal-polyfill/runs/7227471019 - 6m 42s

If we're happy with this change, then I can port it over to proposal-temporal too.

12wrigja:
🎉 Nice speedups!

test/test262.sh
```diff
@@ -20,10 +20,14 @@ TIMEOUT=${TIMEOUT:-10000}
 
 if [ "$(uname)" = 'Darwin' ]; then
   threads=$(sysctl -n hw.logicalcpu)
+  ((threads = threads * 2))
```
12wrigja, line 23: It's interesting to me that this seems to speed things up as apparently the M1 chips don't have hyperthreading? https://apple.stackexchange.com/questions/425668/does-the-m1-chip-apple-silicon-use-hyper-threading-a-k-a-simultaneous-multi

12wrigja: Ah good point. I'll experiment with unconditionally doubling this number and send a PR if it makes a big difference.

12wrigja (approved):

test/test262.sh
12wrigja: I tried unconditionally doubling the threads on my workstation (from 11 to 22) and it didn't make much of a difference: 3m36s to 3m34s.

Issue #164, opened 2022-07-06
Check whether this polyfill runs afoul of Dual Package hazards


12wrigja:
Context: https://nodejs.org/api/packages.html#dual-package-hazard

I was pondering this in the context of #155, specifically whether if two versions of the polyfill sources (one with jsbi, and one with bigint) were to be loaded into the page whether they would correctly interoperate with each other, at the cost of extreme code-size regression (two copies of Temporal srcs, potentially two copies of JSBI!).

My guess is that no, currently they would not, primarily because an object from one instance would not satisfy the slot-based checks when passed to the other: https://github.com/js-temporal/temporal-polyfill/blob/56df93fba8d1a1c4e059eaa302da36392ad66d12/lib/slots.ts#L153

Setting aside the code-size problems, if it were possible to at least have the results be correct when interoperating I would consider publishing a second package that contains a pre-"transpiled" bundle that does not contain JSBI.

12wrigja:
If/when we get to a future where Temporal has shipped presumably this is not a problem because each version of the package will simply provide a reference to the native Temporal instance.

tars0x9752:
> an object from one instance would not satisfy the slot-based checks

This sounds right if some kind of slots related things passed to the other. But is "passed to the other" possible? The slot itself looks rely on "module-local state" but I think each slots (I mean "slots containers" or I should say weakmap keys) are isolated between instances and they look quite private. So I haven't figured out how to reproduce this yet.

I just tried a little bit but at least it didn't happen while simply importing two different versions. 

```js
import { Temporal as Temporal1 } from 'temporal-0.3.0'
import { Temporal as Temporal2 } from 'temporal-0.4.1'

const a = Temporal1.Now.instant()
const b = Temporal2.Now.instant()
const c = Temporal1.Now.instant()
const d = Temporal2.Now.instant()

console.log('a', a.epochSeconds) // no errors are thrown.
console.log('b', b.epochSeconds) // no errors are thrown.
console.log('c', c.epochSeconds) // no errors are thrown.
console.log('d', d.epochSeconds) // no errors are thrown.
```

ptomato:
What happens if you do something like
```js
const a = Temporal1.Now.instant();
const round = Temporal2.Instant.prototype.round;
round.call(a, 'second');
```

tars0x9752:
Thanks! It threw the `TypeError: invalid receiver` that is exactly what I was looking for. So the dual package hazard is definitely possible.

```js
[...]

Issue #155, opened 2022-05-12
FR: provide a "modern" ESM build without jsbi


taralx:
There's a babel plug-in for it, but it would be nice to have a ready-to-import version.

12wrigja:
https://github.com/js-temporal/temporal-polyfill/issues/138#issuecomment-1085369314 seems related.

For your use-case, are you planning on using this polyfill:
 - only within applications (that don't themselves become NPM packages) and can made the decision whether they need JSBI
 - only within an ecosystem of packages where native BigInt is considered available on the platform?
 
 Are you aware of any other packages that have this sort of a problem and how they went about solving it? I'm not opposed to having a solution, but I'd prefer not to create a bifurcation in the ecosystem where there are some packages using this dependency assuming that native BigInt is available only to find that someone who wants to depend on those needs JSBI.
 
 Would it work for your use-case to have JSBI defer to native BigInt if available, but still pay the code-size cost of JSBI itself? This used to be a feature in JSBI but was removed.

taralx:
I'm using it on browser, where we expect all users to have bigint (we need newer features anyway). Having it defer would be ok, but we're trying to keep download sizes down as well, preferably without introducing an additional transpile step.

12wrigja:
I'd probably lean towards making more outputs that are listed as conditional exports (https://nodejs.org/api/packages.html#conditional-exports) and could be imported like so:

```
import {Temporal} from '@js-temporal/polyfill/bigint';
```

Or some such.

I don't want to tie this to the module system used as it can be quite surprising which tools chose to use what thing we specify in package.json (see https://github.com/js-temporal/temporal-polyfill/issues/29 for example).

12wrigja:
Though I think this strategy will lead to weird bifurcation issues within the ecosystem and likely weird issues if both versions are in use at once (besides the expected doubled code-size).

taralx:
I haven't tried it but it almost certainly works. It's just that right now we don't have a transpile requirement, and I was hoping to keep it that way. 🙂

12wrigja:
Would you mind describing your current dev & prod build setups in more detail? I'd like to understand why adding a transpilation step here is something you want to avoid. I can generally understand the desire to have fewer tools/processes in place, but any general solution to the bifurcation problem I presented above likely requires ecosystem-wide tooling support. If your goal is to reduce download code-size then I'd have thought you would already have transpilation / minification in place for prod builds.

taralx:
[...]

PR #167, opened 2022-07-11
Add BigInt bundle

tars0x9752:
Related: #155

For the environments that have native bigint support, This PR will add an subpath export with a separate native bigint build. As far as I believe this shouldn't change/affect existing exports/builds. You can import the bigint build as follows:

```
import { Temporal } from '@js-temporal/polyfill/bigint'
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@js-temporal/polyfill",
-  "version": "0.3.0",
+  "version": "0.4.1",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@js-temporal/polyfill",
-      "version": "0.3.0",
+      "version": "0.4.1",
       "license": "ISC",
       "dependencies": {
         "jsbi": "^4.1.0",
@@ -25,6 +25,7 @@
         "@rollup/plugin-typescript": "^8.3.0",
         "@typescript-eslint/eslint-plugin": "^5.7.0",
         "@typescript-eslint/parser": "^5.7.0",
+        "babel-plugin-transform-jsbi-to-bigint": "^1.4.0",
         "eslint": "^8.4.1",
         "eslint-config-prettier": "^8.3.0",
         "eslint-plugin-prettier": "^4.0.0",
@@ -2268,6 +2269,15 @@
       },
       "peerDependencies": {
         "@babel/core": "^7.0.0-0"
+      }
+    },
+    "node_modules/babel-plugin-transform-jsbi-to-bigint": {
+      "version": "1.4.0",
+      "resolved": "https://registry.npmjs.org/babel-plugin-transform-jsbi-to-bigint/-/babel-plugin-transform-jsbi-to-bigint-1.4.0.tgz",
+      "integrity": "sha512-59f6ClwQBY/SKMVwKV/xoAqH/gnyZ6C5gSSPsEUCLPvvZULCDSE5Y3T62wjNTbpBcFzh0ReYz2mIe3u44mW9UQ==",
+      "dev": true,
+      "engines": {
+        "node": ">=8"
       }
     },
     "node_modules/backo2": {
@@ -7066,6 +7076,12 @@
       "requires": {
         "@babel/helper-define-polyfill-provider": "^0.3.0"
       }
+    },
+    "babel-plugin-transform-jsbi-to-bigint": {
+      "version": "1.4.0",
+      "resolved": "https://registry.npmjs.org/babel-plugin-transform-jsbi-to-bigint/-/babel-plugin-transform-jsbi-to-bigint-1.4.0.tgz",
+      "integrity": "sha512-59f6ClwQBY/SKMVwKV/xoAqH/gnyZ6C5gSSPsEUCLPvvZULCDSE5Y3T62wjNTbpBcFzh0ReYz2mIe3u44mW9UQ==",
+      "dev": true
     },
     "backo2": {
       "version": "1.0.2",
--- a/package.json
+++ b/package.json
@@ -16,7 +16,8 @@
         "default": "./dist/index.cjs"
       },
       "./dist/index.cjs"
-    ]
+    ],
+    "./bigint": "./dist/bigint/index.esm.js"
   },
   "types": "./index.d.ts",
   "scripts": {
@@ -90,6 +91,7 @@
     "@rollup/plugin-typescript": "^8.3.0",
     "@typescript-eslint/eslint-plugin": "^5.7.0",
     "@typescript-eslint/parser": "^5.7.0",
+    "babel-plugin-transform-jsbi-to-bigint": "^1.4.0",
     "eslint": "^8.4.1",
     "eslint-config-prettier": "^8.3.0",
     "eslint-plugin-prettier": "^4.0.0",
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,6 +1,6 @@
 import commonjs from '@rollup/plugin-commonjs';
 import { nodeResolve } from '@rollup/plugin-node-resolve';
-import babel from '@rollup/plugin-babel';
+import babel, { getBabelOutputPlugin } from '@rollup/plugin-babel';
 import replace from '@rollup/plugin-replace';
 import { terser } from 'rollup-plugin-terser';
 import { env } from 'process';
@@ -57,15 +57,20 @@ const external = [
   ...Object.keys(pkg.peerDependencies || {})
 ].map((dep) => new RegExp(dep + '*'));
 
-function outputEntry(file, format) {
+function outputEntry(file, format, plugins = []) {
   return {
     name: libName,
     file,
     format,
     exports: 'named',
-    sourcemap: true
+    sourcemap: true,
+    plugins
   };
 }
+
+const bigintBabelConfig = {
+  plugins: ['transform-jsbi-to-bigint']
+};
 
 const es5BundleBabelConfig = {
   babelHelpers: 'bundled',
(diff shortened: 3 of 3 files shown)


package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "@js-temporal/polyfill",
-  "version": "0.3.0",
+  "version": "0.4.1",
```
tars0x9752, line 3: I don't know the exact reason why this was `v0.3.0`. But I guess it better to match` package.json`'s version.

tars0x9752:
I just noticed that I should deal with the following issue. (Currently, this PR works with JS but TS. )
https://github.com/microsoft/TypeScript/issues/33079

tars0x9752:
https://github.com/js-temporal/temporal-polyfill/pull/167#issuecomment-1180091059 fixed.

12wrigja:
Thanks for this PR! I haven't cloned this to experiment with it, but this looks about right.

That being said, I'd prefer to fix https://github.com/js-temporal/temporal-polyfill/issues/164 first, as anyone trying to use this version along other library code that makes use of the non-bigint version will likely end up with broken code.

tars0x9752:
> use this version along other library code

Ah, I get it. So transitive deps can be a potential dual package hazard risk. At first I simply thought it would be fine as long as users don't import both bigint and non-bigint build at the same time. But it's true that two of them may be mixed at the same time as transitive deps if other library code depend on a diffrent temporal-polyfill build.

Thanks for your comment! I’ll do some research.

tars0x9752:
I could be wrong, but after a little research, it looks like you can use both bigint and jsbi version at the same time.

Here's my thought:
First, I think this package doesn't have any possible shared state between dependencies. Every exports are either a class or a stateless function, not instance. And all static methods of the exported class seem stateless. 

Also, since there's no code like `instanceof SomeKindOfTemporalClass` in this polyfill, The "different instance from different package problem" like [this](https://github.com/GeoffreyBooth/dual-package-hazard) will not happen.

### bigint and JSBI

According to `index.d.ts`, there are only 8 bigint interfaces exposed.

List of bigint interfaces that are exposed as a method's or constructor's parameter:
- `Instant.fromEpochMicroseconds`'s parameter
- `Instant.fromEpochNanoseconds`'s parameter
- `Instant` constructor's parameter
- `ZonedDateTime` constructor's parameter

About these parameters, The type definition only accepts `bigint`, so if the user is using TS, JSBI cannot be passed. Even in the case of JS, if you try to pass an instance of JSBI, the `instanceof JSBI` guard should be false, and you will get an error and not be able to pass it. And if you pass an actual bigint as a parameter with JSBI version, it OK because theres's `ToBigInt` and `ToBigInt` abstract operation implementation will convert bigint type into JSBI in JSBI version.

List of bigint interfaces that are exposed as a get only accessor property:
- `Instant.epochMicroseconds` property
- `Instant.epochNanoseconds` property
- `ZonedDateTime.epochMicroseconds` property
- `ZonedDateTime.epochNanoseconds` property

About these properties, it may use JSBI in their object's internal slot, but when it is called via getter by the user, it must goes through `ToBigIntExternal`. `ToBigIntExternal` will use bigint instead of JSBI if there is a bitint in `globalThis` even if it imported from JSBI version.

So the mixed up code like this will work:

```js
import { Temporal } from "@js-temporal/polyfill/bigint";
import { Temporal as JSBITemporal } from "@js-temporal/polyfill";

const a = new Temporal.Instant(JSBITemporal.Now.instant().epochNanoseconds)
const b = new JSBITemporal.Instant(Temporal.Now.instant().epochNanoseconds)
```

I'm sorry if I missed or misunderstood something. Cheers.

12wrigja:
> Also, since there's no code like instanceof SomeKindOfTemporalClass in [this](https://github.com/GeoffreyBooth/dual-package-hazard) polyfill, The "different instance from different package problem" like this will not happen.

I think that dual-package hazards are still likely because of the way that slots are used and implemented for branding - see https://github.com/js-temporal/temporal-polyfill/issues/164#issue-1296604290. They aren't instanceof checks, but they do heavily rely on module-local state which would be duplicated if there were dual packages. Let's keep the discussion of the general dual-package hazards over there, but the discussion of JSBI/bigint compatibility is good to have here.

Great points r.e. mixing native BigInt and JSBI - I think you are right r.e. mixing them. In particular, we do need to be very careful accepting native `bigint`, as by default `JSBI.BigInt(<some native bigint>)` will throw.

My understanding is that package aliases can be used to allow two versions of a project to be imported at the same time into a project - would this work to setup a set of tests we can use to exercise both versions at the same time and verify compatibility between the two versions?

tars0x9752:
> Let's keep the discussion of the general dual-package hazards over there, but the discussion of JSBI/bigint compatibility is good to have here.

Ok!

> My understanding is that package aliases can be used to allow two versions of a project to be imported at the same time into a project - would this work to setup a set of tests we can use to exercise both versions at the same time and verify compatibility between the two versions?

Do you mean aliases like this? If so, Yes. I think It should do the job.

```
"dependencies": {
  ...
  "temporal-0.3.0": "npm:@js-temporal/polyfill@0.3.0",
  "temporal-0.4.1": "npm:@js-temporal/polyfill@0.4.1"
}
```

Or If you are talking about the bigint and jsbi versions in this branch, we may simply use the package self-referencing to do that.
https://nodejs.org/api/packages.html#self-referencing-a-package-using-its-name

```js
// In the test file:
import { Temporal } from "@js-temporal/polyfill/bigint";
import { Temporal as JSBITemporal } from "@js-temporal/polyfill";
```

Note: If we want to use self-referencing in TS file, then we need TS 4.7 or higher.
https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/#package-json-exports-imports-and-self-referencing

tars0x9752:
A single extensive test suite sounds good! But not sure what to test actually. Also, I just noticed there's a loader hook `resolve.source.mjs` exists and found the code `specifier === PKG.name`. Probably it'll interrupt self-reference behavior. So I think we must add a different npm script for the test suite. In any case, I'll take a look first.

tars0x9752:
Well, `from` doesn't seem to work either.

```js
const a = Temporal.Now.zonedDateTime('iso8601')
const b = TemporalBigInt.ZonedDateTime.from(a); // TypeError: invalid result at CalendarDateFromFields
```

So dual package hazard can happen quite easily actually. Now I feel that it might be better to release the bigint build as a completely different package rather than a subpath export until we find some way around the dual package hazard.

12wrigja:
My concern about releasing multiple packages is that they would likely be compatible according to TypeScript, and so you could easily end up with code that compiles that doesn't work at all at runtime. This only applies if packages use Temporal types as part of their public API surface I suppose, but I'd rather not release something that has this sort of a footgun.

That being said, I've had some thoughts on how to resolve these hazards, and I'll put my thoughts into a comment over in the package hazard issue shortly.

tars0x9752:
I agree. That makes sense.

celluj34:
Would it be possible to publish two _separate_ npm packages, one that includes JSBI and one that doesn't?

rjgotten:
> Would it be possible to publish two _separate_ npm packages, one that includes JSBI and one that doesn't? Does that help the dual-package problem?

Seconding this.
Right now there's `@js-temporal/polyfill` which for least friction for new users (also considering BigInt is well-supported nowadays) should just be the build without JSBI.
Simply have a `@js-temporal/polyfill-jsbi` next to that with JSBI for legacy browser support, where the public API surface has the same shape.

For those cases where existing code must continue to work using the JSBI version: instruct users to use NPM aliases and install `@js-temporal/polyfill-jsbi` _as_ `@js-temporal/polyfill` for all dependencies. I.e.

```jsonc
{
  // ...
  "dependencies": {
    // ...
    "@js-temporal/polyfill" : "npm:@js-temporal/polyfill-jsbi@x.y.z"
  },
  "overrides": {
    // ...
    "@js-temporal/polyfill" : "$@js-temporal/polyfill"
  }
}
```

Yes; this would require pinning to a version. But you're already well into exceptional-case territory at that point anyway.

12wrigja:
I agree that `@js-temporal/polyfill` should be the version sans JSBI.

> Would it be possible to publish two separate npm packages, one that includes JSBI and one that doesn't? Does that help the dual-package problem?

We could ship two packages, but:
  - I don't think it would change anything with respect to dual-package hazards, as those happen regardless of whether we use exports or different NPM packages, as hazards occur when there's two distinct bundles of JS code and interactions between the bundles don't work.
 - we'd probably need a way to indicate that those two packages are meant to be compatible with each other (so that two hypothetical incompatible versions of the project can't be used together, at least not without surfacing that problem to developers somehow via npm versioning constraints).
 
 R.e. aliases - that is a very interesting idea.... is there a way to make aliases apply to an entire transitive closure (tc) of dependencies? If so, then publishing two separate packages might be a solution for users who need to force the use of JSBI throughout their application (e.g. by forcing any libraries in their tc that were using the bigint package to instead use the jsbi package). I suppose that principle doesn't scale to the libraries themselves (they might directly create/manipulate bigints, so even if we swap out Temporal that code would still not work at runtime) so maybe it is a moot point?

12wrigja:
It might be possible to do use NPM overrides in this manner: https://docs.npmjs.com/cli/v9/configuring-npm/package-json?v=true#overrides

Something like (hypothetical):

```json
{
  "dependencies": {
    "@js-temporal/polyfill-no-bigint",
  },
  "overrides": {
     "@js-temporal/polyfill":  "$js-temporal/polyfill-no-bigint",
  }
}
```

if I understand the docs properly.

celluj34:
According to caniuse.com, only Internet Explorer and Opera Mini don't support JSBI: https://caniuse.com/bigint

Microsoft themselves has now disowned IE, so why does this lib need to cater to it? Going a step further, once Temporal becomes stage 4 or better, the only browsers that are going to be able to use Temporal _are already going to support BigInt_.

12wrigja:
> so why does this lib need to cater to it?

When I joined this project, there were certain interested parties (my employer Google, for example) that still support IE, or older versions of "evergreen" browsers that don't have native BigInt.

rjgotten:
> if I understand the docs properly.

Yup. You understood correctly.
Or at least: you understood them the same way I understood them. 😄

khawarizmus:
Google has dropped IE support as of July 15, 2024. Is there still a need for a JSBI version?

12wrigja:
IE is not the only browser without native BigInt support that Google still supports (e.g. FireFox 65, Safari Desktop 13.1).

Maybe within a year or two the requirements will change again and we will finally cross the threshold where all browsers support native BigInt, but ultimately I still think the medium term solution for this project is to make multiple packages, one with and one without JSBI. We would ship the package **without** JSBI by default and document a mechanism for people who need JSBI to override the package to point to the other version transitively for their builds.

PR #168, opened 2022-07-14
Fixed a possible type error if higher TS version is used in the future.

tars0x9752:
Just found a possible small type error. In the current TS version, its type is considered as `Partial<any>` so it's not an error for now, but the type may be not correct actually and I guess it may become a type error in the future.


12wrigja (approved):

tars0x9752:
Sorry, I forgot the lint. Fixed it.

12wrigja:
Thanks!

PR #169, opened 2022-07-15
Editorial: Simplify operations (#2248)

12wrigja:
Similar to other rebases, I intend for this to be merged as-is without squashing.

The typing for GetTemporalUnit is interesting - ideally, I'd like to find a way to avoid all the casts within the function itself and avoid needing overloads, but I can't seem to get TS to recognize that when the default empty array (typed as `never[]`) is used as the `extraValues` property that it expands to `never` in the return type (which is the union identity).

### DIFF
--- a/lib/ecmascript.ts
+++ b/lib/ecmascript.ts
@@ -223,31 +223,22 @@ const BUILTIN_CASTS = new Map<PrimitivePropertyNames, BuiltinCastFunction>([
   ['offset', ToString]
 ]);
 
-const ALLOWED_UNITS: readonly Temporal.DateTimeUnit[] = [
-  'year',
-  'month',
-  'week',
-  'day',
-  'hour',
-  'minute',
-  'second',
-  'millisecond',
-  'microsecond',
-  'nanosecond'
+// each item is [plural, singular, category]
+const SINGULAR_PLURAL_UNITS = [
+  ['years', 'year', 'date'],
+  ['months', 'month', 'date'],
+  ['weeks', 'week', 'date'],
+  ['days', 'day', 'date'],
+  ['hours', 'hour', 'time'],
+  ['minutes', 'minute', 'time'],
+  ['seconds', 'second', 'time'],
+  ['milliseconds', 'millisecond', 'time'],
+  ['microseconds', 'microsecond', 'time'],
+  ['nanoseconds', 'nanosecond', 'time']
 ] as const;
-type PluralAndSingularUnitTuple<U extends Temporal.DateTimeUnit> = readonly [Temporal.PluralUnit<U>, U];
-const SINGULAR_PLURAL_UNITS: readonly PluralAndSingularUnitTuple<Temporal.DateTimeUnit>[] = [
-  ['years', 'year'],
-  ['months', 'month'],
-  ['weeks', 'week'],
-  ['days', 'day'],
-  ['hours', 'hour'],
-  ['minutes', 'minute'],
-  ['seconds', 'second'],
-  ['milliseconds', 'millisecond'],
-  ['microseconds', 'microsecond'],
-  ['nanoseconds', 'nanosecond']
-] as const;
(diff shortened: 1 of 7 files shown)


lib/ecmascript.ts
```diff
@@ -872,7 +863,22 @@ export function ToSecondsStringPrecision(options: Temporal.ToStringPrecisionOpti
   unit: UnitSmallerThanOrEqualTo<'minute'>;
   increment: number;
 } {
-  const smallestUnit = ToSmallestTemporalUnit(options, undefined, ['year', 'month', 'week', 'day', 'hour']);
+  const smallestUnit = GetTemporalUnit(options, 'smallestUnit', 'time', undefined);
+  // This condition is impossible according to type-checking, so we have to
+  // widen the return from GetTemporalUnit to trick tsc.
+  // This could potentially be gated behind DEBUG to remove unnecessary code iff
+  // users are comfortable relying on the typechecker instead.
+  if ((smallestUnit as Temporal.TimeUnit) === 'hour') {
```
ptomato, line 871: Not sure I get why this is necessary from just reading the comment and looking at the signature of the first overload of GetTemporalUnit. I'd expect the return type of that overload to be `Temporal.TimeUnit | 'auto' | undefined` which seems like it shouldn't make the condition impossible.

lib/ecmascript.ts
```diff
@@ -917,87 +923,80 @@ export function ToSecondsStringPrecision(options: Temporal.ToStringPrecisionOpti
...
+type UnitTypeMapping = {
+  date: Temporal.DateUnit;
+  time: Temporal.TimeUnit;
+  datetime: Temporal.DateTimeUnit;
+};
+
+export function GetTemporalUnit<
+  T extends keyof UnitTypeMapping,
+  D extends typeof REQUIRED | UnitTypeMapping[T] | 'auto' | undefined,
+  R extends Exclude<D, typeof REQUIRED> | UnitTypeMapping[T]
```
ptomato, line 943: If I'm reading this correctly, `Exclude<D, typeof REQUIRED>` already is unioned with `UnitTypeMapping[T]`, so that might be redundant?

ptomato (approved):
Looks good, just remarking on some oddities that I noticed which may be just my inexperience with TS.

lib/ecmascript.ts
12wrigja: This is leftover from a previous iteration of the GetTemporalUnit overloads which I screwed up. Thanks for catching - you are right about the return type and this not needing the cast.

lib/ecmascript.ts
12wrigja: I don't believe it's redundant because of the way that the binding D happens.

If I write

`GetTemporalUnit(o, 'smallestUnit', 'time', 'hours')`

then `D` is bound to `'hours'`, not to the whole union type space.

Therefore, we need to re-add that space back into the return type (as the allowed values also include all `TimeUnit` values).

justingrant:
Sorry I'm coming late to this party!  I had it starred in my email app but had a busy few weeks and didn't get to it until now. 

The typing of these Largest/Smallest unit methods was one of the 2 hardest TS challenges when I originally did the TS conversion. The combination of allowing and disallowing options was painful to make typed. 

I have one question on the new implementation. If I remember correctly, some of the methods allowed `auto` but others didn't.  Has that distinction been retained in the new TS types, so we'll catch the case of someone passing `auto` to a method that shouldn't allow it?

12wrigja:
> I had it starred in my email app

In the future, feel free to drop a line here saying something like "id like to review this before it's merged" and I'm happy to wait for your review.

> some of the methods allowed

I think you are referring to `ToSmallestTemporalUnit` didn't accept `auto`, but `ToLargestTemporalUnit` did. I didn't notice that when I was merging this PR, and dropped that detail. There's another commit upstream that further modifies the new impl signature, so I'll incorporate this change into that PR.

PR #170, opened 2022-07-16
dual-pkg

tars0x9752:
#164
This is not an actual PR. Just to sharing the code to investigate the dual package hazard.

### DIFF
--- a/try-dual-pkg/consumer/index.mjs
+++ b/try-dual-pkg/consumer/index.mjs
@@ -14,6 +14,6 @@ const zdtC = createCurrentZDT();
 const zdtD = createZDT(2022, 1, 2);
 
 /* eslint-disable no-console */
-// console.log(getDurationInHours(zdtA, zdtB)); // This will throw TypeError: invalid result
-console.log(getDurationInHours(zdtB, zdtC)); // This will not throw because both are esm
-console.log(getDurationInHours(zdtA, zdtD)); // This will not throw because both are cjs
+console.log(getDurationInHours(zdtA, zdtB)); // This will throw TypeError: invalid result
+// console.log(getDurationInHours(zdtB, zdtC)); // This will not throw because both are esm
+// console.log(getDurationInHours(zdtA, zdtD)); // This will not throw because both are cjs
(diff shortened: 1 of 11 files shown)


try-dual-pkg/lib-esm/index.mjs
```diff
@@ -0,0 +1,9 @@
+import { Temporal } from '@js-temporal/polyfill';
```
12wrigja, line 1: Could this just be 

`export * from '@js-temporal/polyfill';`

 ?

try-dual-pkg/lib-cjs/index.cjs
```diff
@@ -0,0 +1,12 @@
...
+function createZDT(year, month, day) {
+  return Temporal.ZonedDateTime.from({
+    timeZone: 'Asia/Tokyo',
+    year,
+    month,
+    day
+  });
+}
+
+module.exports = { createZDT };
```
12wrigja, line 12: Same here, but in some CJS-compatible complete re-export?

try-dual-pkg/consumer/index.mjs
```diff
@@ -0,0 +1,20 @@
+/* eslint-disable */
+
+import { Temporal } from '@js-temporal/polyfill';
```
12wrigja, line 3: Temporal as TemporalDirect ?

try-dual-pkg/consumer/index.mjs
```diff
@@ -0,0 +1,20 @@
...
+import { createCurrentZDT, getDurationInHours } from 'lib-esm';
```
12wrigja, line 4: Temporal as TemporalESMReExport

try-dual-pkg/consumer/index.mjs
```diff
@@ -0,0 +1,20 @@
...
+import { createZDT } from 'lib-cjs';
```
12wrigja, line 5: Temporal as TemporalCJSReExport

tars0x9752:
Closed because it does not have to be open.

PR #174, opened 2022-07-20
More upstream ports

12wrigja:
These commits will be merged in, not squashed.

### DIFF
--- a/lib/ecmascript.ts
+++ b/lib/ecmascript.ts
@@ -14,6 +14,7 @@ const StringCtor = String;
 const NumberMaxSafeInteger = Number.MAX_SAFE_INTEGER;
 const ObjectCreate = Object.create;
 const ObjectDefineProperty = Object.defineProperty;
+const ObjectGetOwnPropertyDescriptor = Object.getOwnPropertyDescriptor;
 const ObjectIs = Object.is;
 const ReflectApply = Reflect.apply;
 
@@ -22,8 +23,8 @@ import JSBI from 'jsbi';
 
 import type { Temporal } from '..';
 import type {
+  AllTemporalLikeTypes,
   AnyTemporalLikeType,
-  PrimitivePropertyNames,
   UnitSmallerThanOrEqualTo,
   CalendarProtocolParams,
   TimeZoneProtocolParams,
@@ -37,7 +38,7 @@ import type {
   DurationParams,
   PlainDateTimeParams,
   PlainYearMonthParams,
-  FieldRecord,
+  PrimitiveFieldsOf,
   BuiltinCalendarId
 } from './internaltypes';
 import { GetIntrinsic } from './intrinsicclass';
@@ -197,7 +198,7 @@ function abs(x: JSBI): JSBI {
 }
 
 type BuiltinCastFunction = (v: unknown) => string | number;
-const BUILTIN_CASTS = new Map<PrimitivePropertyNames, BuiltinCastFunction>([
+const BUILTIN_CASTS = new Map<AnyTemporalKey, BuiltinCastFunction>([
   ['year', ToIntegerThrowOnInfinity],
   ['month', ToPositiveInteger],
   ['monthCode', ToString],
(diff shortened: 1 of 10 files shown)


ptomato (approved):
I don't get all of the typing stuff, but this seems to implement the changes from upstream as well as allow us to remove some casts.

lib/ecmascript.ts
```diff
@@ -1130,73 +1158,94 @@ function MergeLargestUnitOption<T extends Temporal.DateTimeUnit>(
...
-export function ToPartialRecord<B extends AnyTemporalLikeType>(bagParam: B, fieldsParam: ReadonlyArray<keyof B>) {
-  // External callers are limited to specific types, but this function's
-  // implementation uses generic property types. The casts below (and at the
-  // end) convert to/from generic records.
-  const bag = bagParam as Record<PrimitivePropertyNames & keyof B, string | number | undefined>;
-  const fields = fieldsParam as ReadonlyArray<keyof B & PrimitivePropertyNames>;
+type FieldCompleteness = 'complete' | 'partial';
+type AllKeys<U> = Keys<U extends unknown ? U : never>;
+type AnyTemporalKey = Exclude<AllKeys<AnyTemporalLikeType>, symbol>;
+type Keys<T> = T extends object ? keyof T : never;
```
justingrant, line 1164: Usually `object` is avoided because `null` is `object`. I think `Record<string, unknown>` is most typically used.

lib/ecmascript.ts
```diff
@@ -1130,73 +1158,94 @@ function MergeLargestUnitOption<T extends Temporal.DateTimeUnit>(
...
+  return result as unknown as Owner<Exclude<FieldKeys, undefined>>;
 }
 
 // field access in the following operations is intentionally alphabetical
 export function ToTemporalDateFields(
   bag: Temporal.PlainDateLike,
-  fieldNames: readonly (keyof Temporal.PlainDateLike)[]
-) {
-  const entries: [keyof Temporal.PlainDateLike, 0 | undefined][] = [
+  fieldNames: readonly (keyof Temporal.PlainDateLike & PrimitivePropertyNames)[]
```
justingrant, line 1246: The `keyof Temporal.X & PrimitivePropertyNames` pattern is used in a lot of different places in the code. Should it be abstracted into a generic type e.g. `PrimitivePropsOnly<T>`?  Or would doing this break IDE autocomplete and/or TS compilation due to limitations in TS's understanding of generics?

lib/ecmascript.ts
```diff
@@ -1238,9 +1287,9 @@ export function ToTemporalDateTimeFields(
 
 export function ToTemporalMonthDayFields(
   bag: Temporal.PlainMonthDayLike,
-  fieldNames: readonly (keyof Temporal.PlainMonthDayLike)[]
-) {
-  const entries: [keyof Temporal.PlainMonthDayLike, 0 | undefined][] = [
+  fieldNames: readonly (keyof Temporal.PlainMonthDayLike & PrimitivePropertyNames)[]
+): Required<Temporal.PlainMonthDayLike> {
```
justingrant, line 1291: `PlainMonthDayLike` includes a `calendar` field. Does `PrepareTemporalFields` return a `calendar` field?

justingrant (approved):
I only did a quick review, but what I saw looked good. Only a few suggestions added.

lib/ecmascript.ts
12wrigja: I abstracted it, and it both compiles and autocomplete works correctly still. Thanks for the suggestion - this was useful to solve the comment r.e. calendars as well.

lib/ecmascript.ts
12wrigja: It doesn't, but the typing here could have led you to think it did. I made use of the `PrimitiveFieldsOf` type to remove that (as it now better matches the field key inputs).

lib/ecmascript.ts
12wrigja: Great point - done.

PR #175, opened 2022-07-21
Implements globalThis slots

tars0x9752:
#164
Implements globalThis slots mentioned in https://github.com/js-temporal/temporal-polyfill/issues/164#issuecomment-1185775627 to fix the slots issue with dual package hazard.

### DIFF
--- a/lib/slots.ts
+++ b/lib/slots.ts
@@ -150,9 +150,9 @@ interface SlotsToTypes {
 
 type SlotKey = keyof SlotsToTypes;
 
-const globalSlots = new WeakMap();
+const globalSlots = new WeakMap<Slots[keyof Slots]['usedBy'], Record<keyof Slots, Slots[keyof Slots]['value']>>();
 
-function _GetSlots(container: Slots[SlotKey]['usedBy']) {
+function _GetSlots(container: Slots[keyof Slots]['usedBy']) {
   return globalSlots.get(container);
 }
 
@@ -161,9 +161,9 @@ const GetSlotsSymbol = Symbol.for('@@Temporal__GetSlots');
 // expose GetSlots to avoid dual package hazards
 (globalThis as any)[GetSlotsSymbol] ||= _GetSlots;
 
-const GetSlots = (globalThis as any)[GetSlotsSymbol];
+const GetSlots = (globalThis as any)[GetSlotsSymbol] as typeof _GetSlots;
 
-function _CreateSlots(container: Slots[SlotKey]['usedBy']): void {
+function _CreateSlots(container: Slots[keyof Slots]['usedBy']): void {
   globalSlots.set(container, Object.create(null));
 }
 
@@ -172,7 +172,7 @@ const CreateSlotsSymbol = Symbol.for('@@Temporal__CreateSlots');
 // expose CreateSlots to avoid dual package hazards
 (globalThis as any)[CreateSlotsSymbol] ||= _CreateSlots;
 
-export const CreateSlots = (globalThis as any)[CreateSlotsSymbol];
+export const CreateSlots = (globalThis as any)[CreateSlotsSymbol] as typeof _CreateSlots;
 
 // TODO: is there a better way than 9 overloads to make HasSlot into a type
 // guard that takes a variable number of parameters?
@@ -294,7 +294,7 @@ export function GetSlot<KeyT extends keyof Slots>(
   container: Slots[typeof id]['usedBy'],
   id: KeyT
 ): Slots[KeyT]['value'] {
-  const value = GetSlots(container)[id];
+  const value = GetSlots(container)?.[id];
   if (value === undefined) throw new TypeError(`Missing internal slot ${id}`);
   return value;
 }
@@ -303,11 +303,15 @@ export function SetSlot<KeyT extends SlotKey>(
   id: KeyT,
   value: Slots[KeyT]['value']
 ): void {
-  const slot = GetSlots(container);
+  const slots = GetSlots(container);
 
-  if (id in slot) {
+  if (slots === undefined) {
+    throw new TypeError('Missing slots for the given container');
+  }
+
+  if (id in slots) {
     throw new TypeError(`${id} already has set`);
   }
 
-  slot[id] = value;
+  slots[id] = value;
 }
--- a/lib/slots.ts
+++ b/lib/slots.ts
@@ -150,7 +150,12 @@ interface SlotsToTypes {
 
 type SlotKey = keyof SlotsToTypes;
 
-const slots = new WeakMap();
+const globalTemporalSlotMapKey = Symbol.for('@@Temporal__slots__private_do_not_use');
+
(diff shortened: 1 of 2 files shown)


12wrigja:
Generally, this LGTM. I have a small perf

@ptomato and @justingrant - this makes the slot map accessible outside the module (and so I suppose users could mess with said slot map to cause all sorts of problems). Of course this diverges from the spec, but is that an acceptable change to you both / do either of you have thoughts on how two versions of this polyfill could "communicate" with each other to avoid dual package hazards?

ptomato:
`Symbol.for('@@Temporal__slots__private_do_not_use')` seems "private" enough for friendly code, but not for adversarial code. I don't know if we want it to be a goal of this polyfill to be robust against adversarial code. Maybe we should discuss that.

I'm also curious about something — I can no longer find the comment but at one time I suggested that we migrate over time away from a private slots WeakMap and towards "Impl" objects that share their private members with each other. (This is not happening imminently; I haven't done any work whatsoever towards it.) I wonder if the dual package hazard would still be present in that case? I suspect it would, and would no longer be able to be solved by a global symbol.

12wrigja:
> migrate over time away from a private slots WeakMap and towards "Impl" objects

This sounds similar to #105 which was trying to replace the usage of GetIntrinsic with circular dependencies - maybe that helps narrow down where your comment was?

I think that would have similar problems - at a high-level, if we intend to verify that something "is a specific Temporal thing", then we need access to that from independent executions of a script in order to avoid dual-(multi-?)package hazards, and more generally we need to ensure that any local state that can change the output of function calls remains in sync between the N versions. Slots was the first thing that came to mind, but I haven't delved deeply into the other pieces of module-local state to see whether there's more problems here. Conversely, I think the [caches](https://github.com/js-temporal/temporal-polyfill/blob/85b988fdbcdaee51a1dc9e09a64842c4f1ffedf1/lib/calendar.ts#L594-L655) we have would not be a problem though they would affect the performance of an application as the caches would be out of sync.

12wrigja:
> but not for adversarial code. I don't know if we want it to be a goal of this polyfill to be robust against adversarial code. Maybe we should discuss that.

Agreed, which is why I wanted your opinion. Unlike the capturing of VM intrinsic functions I don't think there's a way to make sure that the WeakMap we get back from this Symbol can only be changed by the polyfill sources - anyone could easily get this weakmap even after the polyfill sources have loaded and adjust the content.

12wrigja:
I think we should move forward with this change, and we can consider alternatives to the static Symbol key (release build time hashing?) if needed.

My only concern is setting up some CI testing that covers this, which I think should probably end up as part of
https://github.com/js-temporal/temporal-polyfill/pull/167.

ptomato:
I wonder if we could avoid exposing the slot map, by instead exposing a function that gets a value from the slot map given an object and a slot name, and another function that sets all the slots at once of an object that wasn't previously in the slot map? That way, slots couldn't be overwritten once an object was created, so the objects would still be immutable even if user code accessed the private symbols.

12wrigja:
I think that's a great idea - reducing the API surface down to only what we need seems reasonable.

@tars0x9752 is that something you can look into? We'd need two symbols, one for each exposed function, but that seems pretty reasonable. We probably also want to adjust the map to be "write-once" - if something already has set a key we should throw.

tars0x9752:
I'm not quit sure yet, but I'll look into it when I have time. And let me clarify one thing: When you say "expose", I read it as "expose through globalThis". Am I correct?

12wrigja:
Right - in the same way we are currently making the map itself accessible via the globalThis[symbol].

tars0x9752:
I did a quick test, and I might be wrong, but just exposing `CreateSlots` and `GetSlots` seems to work even with the dual package. I'll do some more research when I have more time. BTW, Am I allowed to completely rewrite `SetSlot` and `CreateSlots` and where they are used?

ptomato:
Depends on what rewriting you have in mind, I guess :smile: but that seems like it should be OK?

ptomato:
I'd have probably changed CreateSlots to just take an argument that provides all the slots and their values at once, and removed SetSlot altogether. But I don't see anything wrong with doing it this way, either.

tars0x9752:
Thanks for your responses. I've just updated this branch as well. And for now, I minimized the changes as in [the comment here](https://github.com/tars0x9752/temporal-polyfill/commit/b52ad6190db920ebaf0b670a0ac1c4f538599f77#r86146175) instead. Let me know if you prefer to remove SetSlot altogether.

lib/slots.ts
```diff
@@ -287,5 +303,15 @@ export function SetSlot<KeyT extends SlotKey>(
...
   value: Slots[KeyT]['value']
 ): void {
-  GetSlots(container)[id] = value;
+  const slots = GetSlots(container);
+
+  if (slots === undefined) {
+    throw new TypeError('Missing slots for the given container');
+  }
+
+  if (id in slots) {
```
12wrigja, line 312: Wouldn't it be better to write this as 

const existingSlot = slots[id];
if (existingSlot) throw new TypeError(...);
slots[id] = value;

12wrigja (approved):
I'm so sorry @tars0x9752 - I didn't realize this was waiting for my review. One small comment, but otherwise this LGTM.

12wrigja:
One more question: you confirmed this didn't have any dual-package hazards - is it possible (in a follow-up PR) to run an automated test for some of these things?

justingrant (approved):
Nice!

tars0x9752:
> One more question: you confirmed this didn't have any dual-package hazards - is it possible (in a follow-up PR) to run an automated test for some of these things?

Sorry, honestly I forgot about the details of this PR. So I don't think I'll be able to provide such a follow-up PR anytime soon.

lib/slots.ts
tars0x9752: Fixed it as suggested!

tars0x9752:
Oops, it looks like test262 is failing in Node v18 and above. :thinking:  I'll look into it again when I have more time.

tars0x9752:
Hmm, I have a feeling that the problem is related to A NARROW NO-BREAK SPACE (U+202f), do you know what the cause is?

justingrant:
> Hmm, I have a feeling that the problem is related to A NARROW NO-BREAK SPACE (U+202f), do you know what the cause is?

Arrrgh, this was caused by a regression in V8 (https://bugs.chromium.org/p/chromium/issues/detail?id=1416538) which broke Test262. Specifically broke changes that I made recently to prevent Node-version-specific behavior from breaking tests. Sigh.

The fix will have to be made inside Test262. 

As long as those are the only tests that are failing, @12wrigja I think it's OK to ignore those failures since they're unrelated to this PR.

12wrigja:
I've sent https://github.com/js-temporal/temporal-polyfill/pull/230 to add those to the expected failure list which will fix the CI.

Your changes otherwise look good to me, so I'm going to rebase this in. Thanks for contributing!

PR #176, opened 2022-07-30
Spec ports

12wrigja:


lib/ecmascript.ts
```diff
@@ -567,13 +567,29 @@ export function ParseTemporalDurationString(isoString: string) {
...
 // ts-prune-ignore-next TODO: remove if test/validStrings is converted to TS.
 export function ParseTemporalInstant(isoString: string) {
-  const { year, month, day, hour, minute, second, millisecond, microsecond, nanosecond, offset, z } =
+  let { year, month, day, hour, minute, second, millisecond, microsecond, nanosecond, offset, z } =
     ParseTemporalInstantString(isoString);
 
+  if (!z && !offset) throw new RangeError('Temporal.Instant requires a time zone offset');
+  // At least one of z or offset is defined, but TS doesn't seem to understand
+  // that we only use offset if z is not defined (and thus offset must be defined).
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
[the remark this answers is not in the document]
12wrigja: I don't think typescript's inference is good enough with destructuring of those sorts of unions to enable better type-checking than this. If we rewrote this code to avoid destructuring the returned object then it might stand a chance. :)

12wrigja:
R.e. merging the test262 ports: I'm reluctant to drop each of those commits because I do think they provide value in their current form. Without this, test262 would simply jump from one commit to another as the last commit in the series, which would almost definitely cause test262 not to pass at some of these interim commits.

lib/ecmascript.ts
```diff
@@ -4591,10 +4606,16 @@ export function DifferenceTemporalPlainYearMonth(
...
   const fieldNames = CalendarFields(calendar, ['monthCode', 'year'] as const);
-  const otherFields = ToTemporalYearMonthFields(other, fieldNames);
-  const thisFields = ToTemporalYearMonthFields(yearMonth, fieldNames);
-  const otherDate = CalendarDateFromFields(calendar, { ...otherFields, day: 1 });
-  const thisDate = CalendarDateFromFields(calendar, { ...thisFields, day: 1 });
+  const otherFields = ToTemporalYearMonthFields(other, fieldNames) as Required<Temporal.PlainYearMonthLike> & {
+    day: number;
+  };
+  otherFields.day = 1;
+  const thisFields = ToTemporalYearMonthFields(yearMonth, fieldNames) as Required<Temporal.PlainYearMonthLike> & {
```
justingrant, line 4613: A few questions about this cast:
* The return type of `ToTemporalYearMonthFields` is `Required<PrimitiveFieldsOf<Temporal.PlainYearMonthLike>>`, not `Required<Temporal.PlainYearMonthLike>`. To avoid mismatches like this, I will often use the return type directly, e.g. `ReturnType<ToTemporalYearMonthFields> & { day: 1 }`.
* If we use this cast twice, should we create a local type declaration and re-use it?

justingrant (approved):
Looks good!  Only one minor suggestion.

lib/ecmascript.ts
```diff
@@ -5234,12 +5262,16 @@ export function AddDurationToOrSubtractDurationFromPlainYearMonth(
 
   const calendar = GetSlot(yearMonth, CALENDAR);
   const fieldNames = CalendarFields(calendar, ['monthCode', 'year'] as const);
-  const fields = ToTemporalYearMonthFields(yearMonth, fieldNames);
+  const fields = ToTemporalYearMonthFields(yearMonth, fieldNames) as Required<Temporal.PlainYearMonthLike> & {
```
justingrant, line 5265: Same issue here with this cast as above.

lib/ecmascript.ts
12wrigja: Great points - done. Also added an explanation for why this is even needed.

lib/ecmascript.ts
12wrigja: Done - same as above.

PR #179, opened 2022-08-17
Remove outdated expected failure from test262.

12wrigja:
This test has long since been removed from Test262.


ptomato (approved):
Indeed.

Issue #177, opened 2022-08-02
Types not found when `moduleResolution: node16` is set for typescript


sorgloomer:
**Describe the bug**

Getting 

```
main.ts:1:26 - error TS7016: Could not find a declaration file for module '@js-temporal/polyfill'. '.../node_modules/@js-temporal/polyfill/dist/index.cjs' implicitly has an 'any' type.
```

when trying to import in a project with `moduleResolution: node16`.

**To Reproduce**

*package.json*

```
{
  "dependencies": {
    "@js-temporal/polyfill": "^0.4.2",
    "typescript": "4.7.4"
  }
}
```

*tsconfig.json*

```
{
  "compilerOptions": {
    "moduleResolution": "Node16",
    "noImplicitAny": true
  }
}
```

*main.ts*

```
import { Temporal } from '@js-temporal/polyfill';
```

and then run:

```
tsc
```

Output I get:

```
$ tsc
main.ts:1:26 - error TS7016: Could not find a declaration file for module '@js-temporal/polyfill'. '.../node_modules/@js-temporal/polyfill/dist/index.cjs' implicitly has an 'any' type.
  Try `npm i --save-dev @types/js-temporal__polyfill` if it exists or add a new declaration (.d.ts) file containing `declare module '@js-temporal/polyfill';`

1 import { Temporal } from '@js-temporal/polyfill';
                           ~~~~~~~~~~~~~~~~~~~~~~~


Found 1 error in main.ts:1
```

**Additional context**

My intent in this example is to import Temporal using require, and this is what's happening, but typescript doesn't seem to pick up the `.d.ts` file.

Typescripts esm support is not without controversies. One of them is that they are more strict about the location of `.d.ts` files: https://github.com/microsoft/TypeScript/issues/49160 . The most straightforward solution seems to be moving and duplicating `index.d.ts` to `dist/index.d.cts`, `dist/index.esm.d.ts` and `dist/index.umd.d.ts`, but there might be better ways I am not aware of.

12wrigja:
Based on https://github.com/telegraf/telegraf/commit/6ddc3977d3681eb6154a327e73a216ad4c61faa3 I think we can do something very similar to https://github.com/js-temporal/temporal-polyfill/pull/158, but also for the `require` import.

If you are familiar with how to repack npm bundles and use `npm link`, do you want to try this out and send a PR to fix?

It would be ideal if we could re-use the same index.d.ts file for all the references in `exports`.

[...]

PR #181, opened 2022-08-20
Add require section inside exports

shrujalshah28:
Partially Fix: #177

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,11 @@
           "types": "./index.d.ts",
           "default": "./dist/index.esm.js"
         },
+        "require": {
+          "types": "./index.d.ts",
+          "default": "./dist/index.cjs"
+        },
+        "types": "./index.d.ts",
         "default": "./dist/index.cjs"
       },
       "./dist/index.cjs"


package.json
```diff
@@ -13,6 +13,11 @@
           "types": "./index.d.ts",
           "default": "./dist/index.esm.js"
         },
+        "require": {
+          "types": "./index.d.ts",
```
12wrigja, line 17: I tried repro-ing this locally but couldn't get compilation to succeed when this was a .d.ts file. It did work when it was a .d.cts file, but that would mean duplicating the types file we maintain, and I'd prefer if we didn't have to do that (or, if we only did it as part of a build process so we can have a single source of truth).

shrujalshah28: With this, I know we still have an error but after this change, we can get intelligence support in IDE.

shrujalshah28: Do you think we can merge this?
I am open to submitting a follow-up PR to address your suggestion? or Do you think I just submit that change in this PR?

shrujalshah28: https://github.com/js-temporal/temporal-polyfill/pull/184

12wrigja: Sorry for the delayed responses - I was on vacation.

Do I understand correctly that with just this change you were seeing type information in an editor when you weren't before, even though compilation shouldn't be succeeding? That sounds like a bug to me.

Do you have a minimal reproduction + instructions that I can follow?

12wrigja: I'm not opposed at all to finding and merging a fix for this issue, I just want to be sure that it's the right combination of fixes.

12wrigja:
Closing this out as we instead merged https://github.com/js-temporal/temporal-polyfill/pull/184.

PR #183, opened 2022-08-29
More spec ports, and small updates to GetTemporalUnit typings.

12wrigja:
Like other spec port PR's, I will merge this as-is without squashing to retain history.

### DIFF
--- a/lib/ecmascript.ts
+++ b/lib/ecmascript.ts
@@ -24,7 +24,6 @@ import JSBI from 'jsbi';
 
 import type { Temporal } from '..';
 import type {
-  AllTemporalLikeTypes,
   AnyTemporalLikeType,
   UnitSmallerThanOrEqualTo,
   CalendarProtocolParams,
@@ -198,6 +197,14 @@ function abs(x: JSBI): JSBI {
   return x;
 }
 
+export function ArrayPush<NewValue>(
+  arr: ReadonlyArray<NewValue>,
+  ...newItem: readonly NewValue[]
+): ReadonlyArray<NewValue> {
+  ArrayPrototypePush.apply(arr, newItem as any[]);
+  return arr;
+}
+
 type BuiltinCastFunction = (v: unknown) => string | number;
 const BUILTIN_CASTS = new Map<AnyTemporalKey, BuiltinCastFunction>([
   ['year', ToIntegerThrowOnInfinity],
@@ -223,6 +230,15 @@ const BUILTIN_CASTS = new Map<AnyTemporalKey, BuiltinCastFunction>([
   ['era', ToString],
   ['eraYear', ToInteger],
   ['offset', ToString]
+]);
+
+const BUILTIN_DEFAULTS = new Map([
+  ['hour', 0],
+  ['minute', 0],
+  ['second', 0],
+  ['millisecond', 0],
+  ['microsecond', 0],
+  ['nanosecond', 0]
 ]);
 
 // each item is [plural, singular, category]
@@ -1075,7 +1091,7 @@ export function ToRelativeTemporalObject(options: {
       'second',
       'year'
     ] as const);
-    const fields = ToTemporalDateTimeFields(relativeTo, fieldNames);
+    const fields = PrepareTemporalFields(relativeTo, fieldNames, []);
     const dateOptions = ObjectCreate(null);
     dateOptions.overflow = 'constrain';
     ({ year, month, day, hour, minute, second, millisecond, microsecond, nanosecond } = InterpretTemporalDateTimeFields(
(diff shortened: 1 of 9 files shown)


12wrigja:
FYI that I'll be on vacation till the 6th of September (2022), so I won't respond till then. There's no rush on this review.

I've asked some colleagues to take a look at some of the more complex type-system implementation I used for `PrepareTemporalFields` to see if there's any way to simplify it further whilst retaining the useful properties the current impl has:

 - preparing fields returns an object with all the given field names as required properties with the types from the "potential owners" of that field.
  - all the other properties of the potential owners that weren't specificed in the field name list are marked as optional - this allows them to be assigned to without casting / complain from tsc. This isn't perfect - it does mean in some cases casts are still required to change the properties type after assignment (e.g. ecmascript.ts L4574 is a good example).

lib/ecmascript.ts
```diff
@@ -1180,56 +1201,88 @@ type AllKeys<U> = Keys<U extends unknown ? U : never>;
 type AnyTemporalKey = Exclude<AllKeys<AnyTemporalLikeType>, symbol>;
 type Keys<T> = T extends Record<string, unknown> ? keyof T : never;
 type Values<T> = T[keyof T];
-type Entry<K> = K | readonly [K, (string | number | undefined)?];
-type ResolvedFieldTypes<T> = T extends Entry<infer K> ? K : never;
 type Owner<T extends AnyTemporalKey> = Values<{
```
shicks, line 1204: I think this could maybe be simplified to

```typescript
type Owner<K extends AnyTemporalKey> = K extends unknown ? OwnerOf<K, AnyTemporalLikeType> : never;
type OwnerOf<K extends AnyTemporalKey, T> = T extends Record<K, unknown> ? T : never;
```

At which point maybe you don't even need `AllTemporalLikeTypes` or `Values` at all anymore?

lib/ecmascript.ts
```diff
@@ -1180,56 +1201,88 @@ type AllKeys<U> = Keys<U extends unknown ? U : never>;
...
 >(
   bag: Partial<Record<FieldKeys, unknown>>,
-  fields: FieldSpec,
-  completeness: FieldCompleteness = 'complete',
+  fields: ReadonlyArray<FieldKeys>,
+  requiredFields: ReadonlyArray<RequiredFieldKeys> | 'partial',
   { emptySourceErrorMessage }: FieldPrepareOptions = { emptySourceErrorMessage: 'no supported properties found' }
-): Owner<Exclude<FieldKeys, undefined>> {
-  const result: Partial<Record<AnyTemporalKey, string | number | undefined>> = ObjectCreate(null);
+): OwnerT {
```
shicks, line 1282: Is this providing extra typings beyond the first two overloads, or is it just for the implementation? (I'm a little fuzzy on this part).  The parameter list makes me think it's the latter.

Assuming that's the case, you've got two implementations.  The only differences between them are the `requiredFields` parameter and the return type.  Given that, you could potentially consolidate into a single signature with an extra parameter `<R extends 'partial'|readonly RequiredFieldKeys[]>` and return `R extends 'partial' ? Partial<OwnerT> : Owner<RequiredFieldKeys[number]>`?

Not sure if that's an improvement or not.

shicks (commented):
See also the [final state of the playground](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgGIHt3IN4ChkHIAeAXMgM5hSgDmA3PoQJ5kgCuAtgEbQOHIB3Ml0wAbCHBAMAvrlyhIsRCgBCcKDkYFSydt15bkALzKVqIeoaHIR6cZJlywTAA4oAClHQuAPABUAGmQAaQA+ZABeQz9kCCJIEAATcmRJJmQAfmQACmDY+IgklIBrCCZ0GGQYrL8AbWCAXWQyNiSIGFAIRIBKZt0IADcDZzc0HwBBIIAlSOQB9GBE8Iic9Roycd6I8KmGXBGUAFUQYHQQP3QASXBocggEMFOQH0Plw2zD-ITk1JB0rNQL3CrEG0F6cW+KUBoBg0GQr0y8L6rWKIHQAik+1cKCmEHIdiG-je-BiEMKP2yADpqWsQUMoLUGltwii0RjETEyNh6shQMhSuVKn4GmQ6o1pAwDmhRHAwAkgZFDLj8aJCdy8nyBRV4Sczhdroo7g8nkCRchPN4XkEwtJQntEvcZVAUAgzpRiGRUDK5YUfBh0AAfNRQO24AD0Yf4UcIAD0Mk5sWhgBBRIkAPJcABW9zAqC8HDTGNuPkLIGggSTKcSwTKKTJRQoVFoy2QyoJEBLRagMQAZDgeXzUMnUzWmCk+1rKqXyw0ALQZMgAUSICFEbAdnbL3fqDSCrQdHTLS1tDAdq-ULrdYGQLErqYz2YeefQBa75B82CIC90nB4UCCt5mLQQRGN+eh-tIQQAOREFBoYRtG0Zxgmoyju+fgtqSBQNrirpQIkPgwnCwRBGk4RZHkdLDIm4yiKIaFErMDFYZCyCsuiIAcn0Zb0qGZ5OpeIDukYACMZAMf6QbqPBkaIfwyH8ReyCukJ15GAATBsdESZgUkhgwCFybG8ZYqMtGiH42IpCstT+kEwYNJKNF-JZowrOZrl4rU4HQI5pkoOMfyjrMy6ruuHYMYFTCeaEQTkEw3B2KGUoAGpwGueKMSsdSTlUfn+cg05QD4eT1j8UWji2pXYT8aSIkVaYwCVQRRTF3GglATmjA1TXVaxQEWEEGGzCx5IpLh6D4c1bEgKiHHkVU7X0vajpKSpwkAMxkEVPhQQIcEGbJRnIHGQA) I was working with while experimenting.  I have a few links to earlier versions of it throughout the review (as well as a few other links to separate playgrounds, so there's that).

lib/ecmascript.ts
```diff
@@ -1180,56 +1201,88 @@ type AllKeys<U> = Keys<U extends unknown ? U : never>;
 type AnyTemporalKey = Exclude<AllKeys<AnyTemporalLikeType>, symbol>;
 type Keys<T> = T extends Record<string, unknown> ? keyof T : never;
```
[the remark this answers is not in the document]
12wrigja: I tried using this, but ran into weirdness when the keys were optional: https://www.typescriptlang.org/play?ts=4.7.4#code/JYOwLgpgTgZghgYwgAgGIHt3IN4ChkHIAeA-AFzIDOYUoA5sgD7ICuIAJhDKBOwNz5CATwogWAWwBG0Jqw5ce-QQQDu5ZJMwAbCHBCy2nbiF4CAvrlyhIsRCgBCcKDmXFRE6c+aGFJpYWQALwpqWhAGb3ljU1cVCk10HT0DKMVzSzAhAAcUAEkQGGgoXgBpCCFKAB4AFQA+ZABeZGrkCCJIDkpkACUIBHQodkrQQucSgBo5AGsQdBUQepJkEuRRCAA3aAFcTJzl8oB5GDKKmvqmlraO9i7e-sHK0PpJthm5heQlqfL0GGbV5AmTZQba7FAnI4nSgAdWAYAAFr0AI4sYDFIZ1Rr-K4QTrIPRCT49CAotG8M6tdq4m7E+5DJ7hF4gN7zRbIb5CX7E0noilrYEAoFbSycBBaJwofogahBCj5UboqGVDDoWoCAD06oC2sIAD0SLhReLisgpTLAgBGCgQ47lKoqtW4TU6nX6w19Y2S9DSsBBABM1sOtoqsIRyNRvIdGq1LoCbvdYolpu95oAzHKCkVSnbKo4oI7nbG9QajUmzb7AgAWQNCSE5vMFmNF5Bu0sm8tBACsNbrIbhiJJEfJDejzYI+qAA

I did manage to clean up AllKeys in favour of Keys just distributing, but it uses keyof and not infer.

lib/ecmascript.ts
12wrigja: That was very helpful! I worked this in, though because all the TemporalLike types have optional properties I needed to make some adjustments to get `T extends Record<K...>` to work. PTAL.

lib/ecmascript.ts
```diff
@@ -1180,56 +1201,88 @@ type AllKeys<U> = Keys<U extends unknown ? U : never>;
...
+// UnionToIntersection. See https://fettblog.eu/typescript-union-to-intersection/
+type UnionToIntersection<U> = (
+  U extends any
+    ? // TS converts a union to an intersection because types in function
+      // arguments are contravariant. We extract this type using infer and return.
+      (k: U) => void
+    : never
+) extends (k: infer I) => void
+  ? I
+  : never;
```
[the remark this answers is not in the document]
12wrigja: I managed to re-work this to avoid using UnionToIntersection, but I've kept this in mind and made more use of unknown over never.

I did use never in Owner/OwnerOf, and I'm wondering whether there's a difference there and those should also be unknown?

lib/ecmascript.ts
```diff
@@ -1180,56 +1201,88 @@ type AllKeys<U> = Keys<U extends unknown ? U : never>;
...
+
+// Flattens two union types into a single type with optional values
+// i.e. FlattenUnion<{ a: number, c: number } | { b: string, c: number }> = { a?: number, b?: string, c: number }
+// From https://www.roryba.in/programming/2019/10/12/flattening-typescript-union-types.html
+type FlattenUnion<T> = ForceResolve<{
+  [K in keyof UnionToIntersection<T>]: K extends keyof T
+    ? T[K] extends any[]
+      ? T[K]
+      : T[K] extends object
+      ? FlattenUnion<T[K]>
```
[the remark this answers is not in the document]
12wrigja: No - turns out I don't need this. Removed.

lib/ecmascript.ts
```diff
@@ -1180,56 +1201,88 @@ type AllKeys<U> = Keys<U extends unknown ? U : never>;
...
+      : T[K]
+    : UnionToIntersection<T>[K] | undefined;
```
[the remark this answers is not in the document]
12wrigja: `Prop` was the missing link for me - being able to do the distributed keyof and then unioning the results together was exactly what this needed. Thanks.

lib/ecmascript.ts
```diff
@@ -1180,56 +1201,88 @@ type AllKeys<U> = Keys<U extends unknown ? U : never>;
...
+}>;
+// ForceResolve copies the keys and values of a given object so that TS will
+// stop using type names in error messages / autocomplete. Generally, those
+// names can be more useful, but sometimes having the primitive object shape is
+// significantly easier to reason about (e.g. deeply-nested types).
+// ESLint doesn't like the Function type here, but it is necessary
```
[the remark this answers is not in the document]
12wrigja: Not necessary. Reading over the docs for the eslint rule, it seems I'm now including/remapping class declaration types?

lib/ecmascript.ts
```diff
@@ -1180,56 +1201,88 @@ type AllKeys<U> = Keys<U extends unknown ? U : never>;
...
+// eslint-disable-next-line @typescript-eslint/ban-types
+type ForceResolve<T> = T extends Function ? T : { [K in keyof T]: T[K] };
```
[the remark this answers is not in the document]
12wrigja: Documented.

lib/ecmascript.ts
```diff
@@ -1180,56 +1201,88 @@ type AllKeys<U> = Keys<U extends unknown ? U : never>;
...
+type FieldObjectFromOwners<OwnerT, FieldKeys> = FlattenUnion<
```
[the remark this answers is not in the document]
12wrigja: I've added docs all over the place to try and make sense of the type-magic going on, and I'd appreciate if you could review them for accuracy.

lib/ecmascript.ts
```diff
@@ -1180,56 +1201,88 @@ type AllKeys<U> = Keys<U extends unknown ? U : never>;
...
+  {
+    // Keys within FieldKeys are always present - either because that key was
+    // required to be present in the bag given to the function or a default value
+    // was used instead.
+    [K in keyof OwnerT as K & FieldKeys]-?: OwnerT[K];
```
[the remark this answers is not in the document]
12wrigja: I did need to remove undefined, and separately force-add undefined / property optionality before joining them together. That being said, I think it's now slightly more complicated than it was before.

lib/ecmascript.ts
```diff
@@ -1180,56 +1201,88 @@ type AllKeys<U> = Keys<U extends unknown ? U : never>;
...
-  FieldKeys extends ResolvedFieldTypes<FieldSpec[number]>,
-  FieldSpec extends ReadonlyArray<Entry<AnyTemporalKey>>
+  FieldKeys extends AnyTemporalKey,
+  OwnerT extends Owner<Exclude<FieldKeys, undefined>>
 >(
   bag: Partial<Record<FieldKeys, unknown>>,
-  fields: FieldSpec,
-  completeness: 'partial',
+  fields: ReadonlyArray<FieldKeys>,
+  requiredFields: 'partial',
```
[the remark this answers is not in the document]
12wrigja: I'm copying the method signature from the [spec](https://github.com/tc39/proposal-temporal/blob/b4d9abc9d77419d3f2075887905b452299e5fa23/spec/abstractops.html#L1631), which calls it requiredFields.

lib/ecmascript.ts
12wrigja: Thanks - only having the one signature is a large improvement in readability.

lib/ecmascript.ts
```diff
@@ -1176,60 +1196,89 @@ function MergeLargestUnitOption<T extends Temporal.DateTimeUnit>(
...
+  // Re-mapping doesn't work very well for functions, so exclude them
+  T extends (...args: never[]) => unknown
+    ? T
+    : // Re-map all the keys in T to the same value. This forces TS into no longer
+      // using type aliases, etc.
+      { [K in keyof T]: T[K] };
+
+type FieldObjectFromOwners<OwnerT, FieldKeys extends AnyTemporalKey> = Resolve<
+  {
+    -readonly // All keys in FieldKeys are required and don't include undefined.
```
shicks, line 1255: This comment seems to apply to the line below, _not_ to `-readonly`, which I find surprising.  Similar a few lines down.

lib/ecmascript.ts
```diff
@@ -1176,60 +1196,89 @@ function MergeLargestUnitOption<T extends Temporal.DateTimeUnit>(
...
-  FieldSpec extends ReadonlyArray<Entry<AnyTemporalKey>>
+  FieldKeys extends AnyTemporalKey,
+  OwnerT extends Owner<FieldKeys>,
+  // Constrains the Required keys to be a subset of the given field keys
+  // This could have been written directly into the parameter type, but that
+  // causes an unintended effect where the required fields are added to the list
+  // of field keys, even if that key isn't present in 'fields'.
+  RequiredFieldKeys extends FieldKeys,
+  RequiredFields extends ReadonlyArray<RequiredFieldKeys> | FieldCompleteness,
+  ReturnT extends RequiredFields extends 'partial' ? Partial<OwnerT> : FieldObjectFromOwners<OwnerT, FieldKeys>
```
shicks, line 1272: This is new syntax to me: `T extends A extends B` - is that the same as `T extends A&B`?

Also, this (still?) looks like "return-only generics".  It sounds like you said this is intended to be inferred from usage.  The problem with inferred return-only-generics is that it allows requiring arbitrary constraints on the type that comes out of a function, which obviously is infeasible.  See this [toy example](https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXwHMQMBJDAHgBUA+ACgEoAueSgbgFgAoLrVDEGIihgEAERwF4Aby7x4AI1gBrBswBuOLMA6cAvly5g8AZwzxgE5uMkBeQsTIMdFggDpFMFfVbwA9L-gMAAsBEAByY3hUHHgAdygAT3tSDAZ4I2QIYHgABxxjYyx5CCSlaNjAmN5IZFB4KAVlNIBbYiCcYFcuoA) for how this can happen even accidentally, when the type is inferred rather than explicitly specified.  I'd argue that return-only-generics can be appropriate if the generic is nested in a template parameter of the returned type (e.g. `declare function emptyList<T>(): T[]`) but this is not what's happening here.

Stepping back, what is the intended usage pattern here?  You've got _a lot_ of parameters, with no defaults - are they all intended to be inferred, or specified directly?

It can be useful to pull out named intermediate types (like `ReturnT`) as generics, but I'd expect them to be assigned directly via a default (and marked `private` if only [that feature](https://github.com/microsoft/TypeScript/issues/26242#issuecomment-1230903819) would land, which it never will).  But I'm puzzled about what this is doing.  Writing this with `=` instead of `extends` doesn't necessarily make it safer from tampering (i.e. intentionally specifying something arbitrary) but it _does_ prevent accidental misuse by calling it with code that with back-infer a more specific type than this is prepared to return.

I dunno if you're interested in this, but it's actually possible to make arguments mostly-private by writing them as `declare function foo<NoExplicitArgsPastThisPoint extends typeof privateSymbol, ...>` ([example](https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEYD2A7AzgF3gBxgSwDcoMQBlATwFsAjJCALngFcU8BHJhNK2iAbgCwAKGGhIsBADMWYDHlTxJAHgBySAKIAPbBDxg8GAIIwA5mnghNJFMHMZy2EEkk58REhRp0ANPAAqAHwAFLAmjH4AlOEA2gC6gkJi0HCKMnIKJkp+vmpaOnoGxmYWViA2dg5OLriExGQ8dPAAvPD2js6utR4NEL4Aqs3+ccGh4VHwfQnCyOhYmoOSQQAsAEwRCTOY8OSDmZj4KCYAPihMNCAwwQDkmlfrwkA)), which prevents tampering with them, other than passing an explicit `never`.

lib/ecmascript.ts
12wrigja: Prettier really didn't like where I had placed my comments (one each above the { for the type), so I've combined all the comments together into one large comment which it doesn't seem to want to move around.

lib/ecmascript.ts
12wrigja: It's not new syntax, just very hard to read :| 

It should be 

ReturnT extends (RequiredFields extends 'partial' ? Partial<OwnerT> : FieldObjectFromOwners<OwnerT, FieldKeys>).

Thanks for the discussion r.e. generic bounds and return-only-generics. Following that advise, I've moved as much of the type-magic out of the generic param list to helper types, retaining only FieldKeys and RequiredFields as the two parameters that need inference.

12wrigja: Steve and I discussed this offline and the new implementation LGTM.

lib/ecmascript.ts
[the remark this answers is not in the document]
12wrigja: Clarified where using `never` is critical / correct, vs where it was being used as an impossible error state by changing the latter to use string literals (which would cause all sorts of errors).

12wrigja:
@justingrant @ptomato I think Steve and I are done with back-and-forth on this, and I'd appreciate if you could take a look. Thanks!

lib/ecmascript.ts
```diff
@@ -1243,86 +1296,24 @@ export function PrepareTemporalFields<
         // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
         value = BUILTIN_CASTS.get(property)!(value);
       }
-      result[property] = value as string | number | undefined;
-    } else if (completeness !== 'partial') {
-      if (wrappedFieldRecord.length === 1) {
+      result[property] = value;
+    } else if (requiredFields !== 'partial') {
+      // TODO: using .call in this way is not correctly type-checked by tsc.
+      // We might need a type-safe Call wrapper?
```
ptomato, line 1302: Could es-abstract's Call be the thing to use here?

ptomato (approved):
LGTM although I didn't deep-dive into all the changes to the typing of PrepareTemporalFields.

Separately, if it turns out that the types of PrepareTemporalFields are that convoluted, it may be worth deviating from the spec text there (in a way that's not observable from JS, of course) and having separate operations for PreparePlainDateFields, etc. if their types are easier to express. But that would be out of scope for this PR.

lib/ecmascript.ts
12wrigja: The typings for es-abstract's Call look useful. I'll investigate this and follow up with a PR once I've figured out if there's enough of these mis-typed calls in our codebase to justify using the es-abstract version or whether a local function definition is enough.

PR #184, opened 2022-08-30
add script to copy types for commonjs module system

shrujalshah28:
Fix: #177

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,10 @@
           "types": "./index.d.ts",
           "default": "./dist/index.esm.js"
         },
+        "require": {
+          "types": "./index.d.cts",
+          "default": "./dist/index.cjs"
+        },
         "default": "./dist/index.cjs"
       },
       "./dist/index.cjs"
@@ -25,7 +29,7 @@
     "testValidStrings": "npm run build && node --experimental-modules --experimental-specifier-resolution=node --no-warnings --icu-data-dir ./node_modules/full-icu/ --loader ./test/resolve.source.mjs test/validStrings.mjs",
     "build": "rm -rf dist/* tsc-out/* && tsc && rollup -c rollup.config.js",
     "prepare": "npm run build",
-    "prepublishOnly": "npm run build",
+    "prepublishOnly": "node copy-types.mjs && npm run build",
     "update": "npx npm-check-updates -u -x @pipobscure/demitasse && npm install",
     "playground": "TEMPORAL_PLAYGROUND=1 npm run build && node --experimental-modules --no-warnings --icu-data-dir node_modules/full-icu -r ./dist/playground.cjs",
     "lint": "eslint . --ext ts,js,mjs,.d.ts --max-warnings 0 --cache \"$@\" && npm run prettier",
@@ -69,6 +73,7 @@
   "license": "ISC",
   "files": [
     "index.d.ts",
+    "index.d.cts",
     "dist",
     "lib",
     "CHANGELOG.md"
--- a/copy-types.mjs
+++ b/copy-types.mjs
@@ -1,0 +1,3 @@
+import { cp } from 'node:fs/promises';
+
+await cp('index.d.ts', 'index.d.cts');


shrujalshah28:
I just tested the script, but Haven't tried to publish it.

package.json
```diff
@@ -13,6 +13,10 @@
           "types": "./index.d.ts",
           "default": "./dist/index.esm.js"
         },
+        "require": {
+          "types": "./index.d.cts",
+          "default": "./dist/index.cjs"
+        },
```
12wrigja, line 19: I think the types key is also needed for the "default" case below, also pointing to "./index.d.cts"?

12wrigja:
@shrujalshah28 Let's focus the review here, and drop https://github.com/js-temporal/temporal-polyfill/pull/181/files as is duplicated with this.

12wrigja:
Based on the way other projects are fixing their types for Node's ESM support, I think this is the right fix. Once the key is included for the default this should be good to merge.

@shrujalshah28 if you have a repo setup to use Node's ESM support that I can use to verify this fixes things, that would be most helpful.

shrujalshah28:
@12wrigja Currently I use ESM in private repository but tomorrow I can create small repository to verify changes.

12wrigja:
Thanks. I'd be especially interested to see if it's different from the repro in https://github.com/js-temporal/temporal-polyfill/issues/177#issue-1325208301 - I think this is the same problem / fix, but want to be sure.

shrujalshah28:
@12wrigja Here is [repo](https://github.com/shrujalshah28/temporal-polyfill-typescript-issue-177).

package.json
shrujalshah28: I don't think so. When I try to add that key, It (VSCode) did not suggest adding that key.

shrujalshah28: For now, everything is working without that type. But if you think we can add that too.

12wrigja:
With your repro repo, I think we would also need fixes to JSBI (our only dependency) in order for this to work - without also changing the JSBI package.json, the temporal project doesn't seem to compile anymore:

```                                  
> @js-temporal/polyfill@0.4.2 build
> rm -rf dist/* tsc-out/* && tsc && rollup -c rollup.config.js
                                                   
lib/duration.ts:21:18 - error TS2307: Cannot find module 'jsbi' or its corresponding type declarations. 
                                                   
21 import JSBI from 'jsbi';                                                                            
                    ~~~~~~
                                                   
lib/ecmascript.ts:23:18 - error TS2307: Cannot find module 'jsbi' or its corresponding type declarations.
                                                                                                       
23 import JSBI from 'jsbi';                                                                            
                    ~~~~~~                         

lib/instant.ts:9:18 - error TS2307: Cannot find module 'jsbi' or its corresponding type declarations.

9 import JSBI from 'jsbi';
```

JSBI's package.json is set up a bit differently than ours and doesn't have exports at all. I'm wondering whether introducing exports could be a breaking change.

shrujalshah28:
Ohh, I am not an expert on this but I will try this and find the solution.

> I'm wondering whether introducing exports could be a breaking change.
Yes

12wrigja:
I'm also an owner for JSBI, so I'll chat with the other maintainers over the next couple days and see if we can fix JSBI up as well. It's very unfortunate that this has a transitive effect on the ecosystem.

shrujalshah28:
@12wrigja When I try to build in my branch, It works fine. I am able to build packages.

```
some\path\temporal-polyfill>npm run build

> @js-temporal/polyfill@0.4.2 build
> tsc && rollup -c rollup.config.js


tsc-out/index.js → ./dist/index.esm.js, ./dist/index.cjs...
created ./dist/index.esm.js, ./dist/index.cjs in 2.1s

tsc-out/index.js → ./dist/index.umd.js...
Browserslist: caniuse-lite is outdated. Please run:
  npx browserslist@latest --update-db
  Why you should do it regularly: https://github.com/browserslist/browserslist#browsers-data-updating
created ./dist/index.umd.js in 7.1s
```

package.json
12wrigja: Let's leave the types key out of this PR, as it doesn't seem to be needed to fix your problems. If someone can find a repro that shows we need it then I'll accept a PR then.

12wrigja (approved):

PR #191, opened 2022-10-25
Fix the link to other polyfills

MidnightDesign:
The anchor of the link to other polyfills in the tc39/proposal-temporal repository has changed from `polyfill` to `polyfills`.


ptomato (approved):
Good catch, thanks!

PR #193, opened 2022-11-13
Fix sourcemap warnings in VSCode debugger

### LINKED ISSUE #192: Invalid sourcemap paths for ESM and CJS bundles
The current release (~~0.4.2~~ EDIT 0.4.3) emits a sourcemap with invalid file paths, which results in console errors in IDEs like VSCode that prefer file references to inline source references in sourcemaps. I haven't kept up with recent changes in build steps, I believe this is a new problem. 

Here's the first few lines of `node_modules/@js-temporal/polyfill/dist/index.esm.js.map`:

```
{"version":3,"file":"index.esm.js","sources":["../tsc-out/intrinsicclass.js","../tsc-out/slots.js","../tsc-out/calendar.js","../tsc-out/ecmascript.js","../tsc-out/regex.js","../tsc-out/intl.js","../tsc-out/instant.js","../tsc-out/plaindate.js","../tsc-out/plaindatetime.js","../tsc-out/duration.js","../tsc-out/plainmonthday.js","../tsc-out/now.js","../tsc-out/plaintime.js","../tsc-out/timezone.js","../tsc-out/plainyearmonth.js","../tsc-out/zoneddatetime.js","../tsc-out/legacydate.js","../tsc-out/index.js"]
```

To fix the problem, we should change `../tsc-out` to `../lib`.  I've run into this problem before, and if my memory is correct there's either a TSC and/or a rollup setting that will allow you to set the correct directory path for the sourcemap. If that's too hard, we could just post-process the output in a build step.

justingrant:
Fixes #192. by adding the `tsc-out` folder to the package, which will solve the sourcemap warnings you'll see in VSCode in the debugger console. 

It's possible that there may be some other way to fix sourcemaps without making the package larger, but at least in the meantime we can fix the debugger console warnings while we look for a better solution if one exists.

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,6 +76,7 @@
     "index.d.cts",
     "dist",
     "lib",
+    "tsc-out",
     "CHANGELOG.md"
   ],
   "dependencies": {


justingrant:
Weird, there are test failures in Node 16, but this PR doesn't make any runtime code changes, so those must be existing failures in the codebase, perhaps caused by a time zone DB update in a Node 16 update?  Or perhaps we made a change recently to make Node 17 work well, but it broke Node 16 because the same TZDB hasn't been applied there yet?

ptomato (commented):
This seems well-justified to me, but I don't know much about the topic of packaging NPM modules, so I'd prefer to leave the review to @12wrigja.

The test failures look almost certainly due to a TZDB update — in the most recent release, some time zones' pre-1970 data was consolidated. See https://github.com/tc39/test262/commit/0bc0ca189da481b3cb8afe76d79e1f73ce2b3c00

12wrigja:
I'll look a bit more closely at the sourcemap setup later today, but wanted to ask if it's possible to pin ourselves to particular versions of this TZ data (by pinning specific Node versions?) so we don't have randomly failing tests.

ptomato:
That should be possible, but it would additionally be good to replace these with tests that won't depend on specific time zone data.

justingrant:
Should we hold this PR waiting for all-green fix for the node 16 time zones?  Or OK to merge this one?

ptomato:
I don't think it's necessary to wait for the TZDB fix, but I do think James still wanted to look at the sourcemap setup.

justingrant:
Replaced by #194

PR #197, opened 2022-12-02
Stop relying on IANA timezone database information in failing tests.

12wrigja:


ptomato (approved):
Yep, we did something similar in test262.

PR #198, opened 2022-12-06
Remove redundant non-null assertion operators.

12wrigja:


lib/calendar.ts
```diff
@@ -334,7 +334,7 @@ export class Calendar implements Temporal.Calendar {
   static from(item: Params['from'][0]): Return['from'] {
     return ES.ToTemporalCalendar(item);
   }
-  [Symbol.toStringTag]!: 'Temporal.Calendar';
```
justingrant, line 337: Does the reason we originally added this assertion (see https://github.com/js-temporal/temporal-polyfill/pull/53#issuecomment-924524613) not apply anymore? Does TS now emit `[Symbol.toStringTag]` even without the `!` assertion?

justingrant (commented):
If the assertions aren't needed anymore, OK to remove them... but should validate that they don't break the original reason why we put them there. See comment above.

lib/calendar.ts
justingrant: Note that this is not actually a non-null assertion, it's a "[definite assignment assertion](https://www.typescriptlang.org/docs/handbook/2/classes.html)" that means that this field will be initialized by non-constructor code.

PR #201, opened 2022-12-08
Split Build and Test steps in CI.

12wrigja:
Make use of reusable GitHub workflows and matrix testing strategies.


justingrant:
I admit I don't fully understand what this PR is doing. Where did Test262 runs go?

12wrigja:
I haven't finished work on this - I was just pushing it to see what the GitHub CI does. The idea was the split all the execution of build steps and testing steps so that builds would only ever be run using newer versions of Node, while testing against the artifacts could then use older versions. I realized later this doesn't make sense for Demitasse tests, as those don't actually build using Rollup (using Node), they only need transpilation to succeed (and are then executed by a particular Node version that reads the transpiled library sources along with the test sources). 
After the changes in https://github.com/js-temporal/temporal-polyfill/pull/202 I don't think this exact structure is necessary (though something similar might be a good cleanup just to reduce the duplication within our workflows).

I commented out the test262 runs temporarily because I though it was causing GH to think the workflow file was broken (there's some weird edge-cases where you get no feedback about this either in the GH UI or in the project-wide actions/workflows UI).

justingrant:
Is this PR still relevant given the other changes we made to use matrix tasks in CI?

PR #202, opened 2022-12-08
Stop unnecessarily running Rollup when running Demitasse tests.

12wrigja:


justingrant:
Sorry @12wrigja I'm behind on reviews.  Is the idea behind this PR that we don't use the rollup output in Demitasse tests so we can omit it?  If yes, then do you know _why_ we don't use that output?  Wondering if we should be using the same bundled code for testing that we are publishing on npm?

12wrigja:
That is the idea - currently, the Node resolver will resolve to the transpiled JS that tsc outputs. I believe this is exactly the same as how the spec polyfill runs (ran?) their Demitasse tests.

We could use one of the build artifacts for Demitasse tests (the ESM build output is probably suitable here), but I don't think it's going to get us much besides testing that bundling works properly, unlike say the test262 tests that use the transpiled es5 output, which also catch problems in our transpilation setup.

I saw we had adopted matrixes in the test setup but haven't looked particularly carefully at them. I'll take a look and provide my thoughts there too soon.

12wrigja:
Closing this as after some rebasing we don't really have very many Demitasse tests left, and this isn't really worth the effort to fix.

PR #203, opened 2022-12-09
CI Tests for Node 18 and 19

justingrant:
Makes CI tests pass in Node 18 and Node 19.  It was a little more complicated than https://github.com/tc39/proposal-temporal/pull/2448 because TS code and demitasse tests required some changes.

* Update CI scripts to test Node 19, 18, 16, and 14 using GH's "matrix" strategy.
* Add non-ISO-calendar snapshot data for Node 18/19
* Make demitasse tests resilient to format changes in CLDR 42 (which is included in ICU 72 which is used in GH Actions' deployments of Node 19)
* Switch to use the optimized Test262 runner from [`@js-temporal/temporal-test262-runner`](https://github.com/js-temporal/temporal-test262-runner)
* Snap to the latest Test262, which contains PRs that make those tests work with CLDR 42: 
  * https://github.com/tc39/test262/pull/3750
  * https://github.com/tc39/test262/pull/3751
  * https://github.com/tc39/test262/pull/3762
* Adds a new `test/expected-failures-todo-migrated-code.txt` file that tracks Test262 failures due to this repo being behind proposal-temporal in merging the latest normative changes to the Temporal spec. As upstream changes are migrated here, tests should be removed from that file.


test/resolve.source.mjs
```diff
@@ -5,11 +5,19 @@
...
     specifier = new URL(specifier.replace('../lib/', '../tsc-out/'), import.meta.url).toString();
+    if (!specifier.endsWith('.js')) specifier += '.js';
+  } else if (
+    context.parentURL &&
+    context.parentURL.includes('/tsc-out/') &&
+    specifier.startsWith('./') &&
+    !specifier.endsWith('.js')
+  ) {
+    specifier += '.js';
   }
```
12wrigja, line 21: Can you document an example of what URIs were being resolved here?

12wrigja (approved):

test/resolve.source.mjs
justingrant: (copying from our chat room, sorry for the dupe!) @12wrigja It looks like Node 19 requires a `.js` file extension on all imports, and TSC does not add `.js` to imports as part of transpilation, although babel does. So our demitasse tests (which depend on TSC, unlike our rollup packaging which depends on babel) break on Node 19. There are at least two ways to fix this:

1. Add .js to all our relative imports, like [this commit](https://github.com/js-temporal/temporal-polyfill/pull/203/commits/8bd2c67324da27695e319074d512720857a996b3)
2. Leave our existing polyfill code as-is, and fix the problem in our custom resolver used for demitasse tests, like [this commit](https://github.com/js-temporal/temporal-polyfill/pull/203/commits/15e072e999a6cd9b4a7e0dd617bb30947b1a14a1#diff-b377bd07b17032f760a56c4027320cbcf6401c49f4b4313ae9513f7e74049ee3).

Got a preference?

justingrant:
All Demitasse tests now pass. Node 19 made many changes in the output of `Intl.DateTimeFormat`. It took a bit of time to write node-version-agnostic code for affected tests.

But I'm stumped with Test262, which is failing in Node 19 but when I look at the test output I don't see any failures, only successes and expected failures. @ptomato , you're a Test262 expert. Any ideas about what may be going on?

justingrant:
> But I'm stumped with Test262, which is failing in Node 19 but when I look at the test output I don't see any failures, only successes and expected failures. @ptomato , you're a Test262 expert. Any ideas about what may be going on?

Mystery solved by copying https://github.com/tc39/proposal-temporal/blob/main/polyfill/runtest262.mjs into this repo. Now I can see the failures, which seem to be related to the Intl format changes that caused the Node 19 Demitasse failures.

justingrant:
PR filed in Test262 for the failures in Node 19: https://github.com/tc39/test262/pull/3750

I'm not sure what's going on with the `test262-es5` and `test262-opt` runs. I assume these are rollup config issues?

test/resolve.source.mjs
justingrant: Per our offline convo, I removed (1) and left (2) above.

justingrant:
This PR should be ready to go, but unfortunately because Test262 changes were needed, it's blocked on catching up with the last few months of normative changes from proposal-temporal, so that we can pass the last few months of Test262 changes. When we've caught up with the normative changes then I can rebase this PR and it should be good to merge. 

I'll change to a draft PR in the meantime.

justingrant:
@12wrigja All tests now pass because I just added a new expected-failures file to include the 897 (!!!) Test262 tests that fail because of commits that we have yet to migrate over from proposal-temporal. Seems like we have two options:

1. Merge this PR as-is, and as we migrate commits from proposal-temporal, then remove lines from the new expected-failures file. 
2. Wait until proposal-temporal commits are migrated, then merge this PR (minus the new expected-failures file).

Which would you prefer?  Personally I like (1) because it's one less rebase to worry about in the future, but I'm flexible either way.

12wrigja:
Let's proceed with 1 - I don't want this work to bit-rot, and my understanding is the 262 runner will flag tests that are supposed to fail but pass and we can remove the relevant lines then.

PR #204, opened 2022-12-22
Use optimized Test262 from proposal-temporal

justingrant:
This PR switches to use the the optimized Test262 runner library [`@js-temporal/temporal-test262-runner`](https://github.com/js-temporal/temporal-test262-runner) that was based on the original code from proposal-temporal: https://github.com/tc39/proposal-temporal/blob/a4728e3b8567d4f62ef518863a23dfb790fe8a83/polyfill/runtest262.mjs

I split this PR out of the Node 19 PR (#203) because this runner change can be merged right away. #203 can't be merged until we make a bunch of polyfill code changes to support Node 19 and to catch up to the latest Node-19-friendly rev of Test262.


ptomato (changes requested):
I have a pretty strong hunch that the failures are because we need to load alternative expected-failures files for the "opt" and "es5" test runs, due to limitations in Terser (in the former case) and in ES5 JavaScript (in the latter case).

justingrant:
> we need to load alternative expected-failures files for the "opt" and "es5" test runs, due to limitations in Terser (in the former case) and in ES5 JavaScript (in the latter case).

@ptomato Yep, that was it. It's fixed now for all three variations. I also updated this PR to use the new `@js-temporal/temporal-test262-runner` library instead of copypasta.

12wrigja (approved):
This is beautiful! Thanks for doing this.

ptomato (approved):
Ditto.

PR #210, opened 2023-01-14
Rebase updates 1

12wrigja:


12wrigja:
Given how Justin already updated test262, I'm tempted to squash/delete all the testing-related changes as those do eventually get turned off (as they were all migrated over to test262).

lib/calendar.ts
```diff
@@ -2314,16 +2313,12 @@ class NonIsoCalendar implements CalendarImpl {
...
-    // a code.
     const cache = new OneObjectCache();
-    const fields = ES.PrepareTemporalFields(
-      fieldsParam,
-      ['day', 'era', 'eraYear', 'month', 'monthCode', 'year'],
-      ['day']
-    );
+    // For lunisolar calendars, either `monthCode` or `year` must be provided
+    // because `month` is ambiguous without a year or a code.
+    const fieldNames = this.fields(['day', 'month', 'monthCode', 'year']) as readonly ES.AnyTemporalKey[];
```
justingrant, line 2319: Same comment as above re: casting.

lib/ecmascript.ts
```diff
@@ -1200,7 +1213,7 @@ interface FieldPrepareOptions {
   emptySourceErrorMessage: string;
 }
 
-type AnyTemporalKey = Exclude<Keys<AnyTemporalLikeType>, symbol>;
+export type AnyTemporalKey = Exclude<Keys<AnyTemporalLikeType>, symbol>;
```
justingrant, line 1216: If we don't use this exported type per comment above, then shouldn't export this. And if it is shared, then we may want to move it into `intnernaltypes.d.ts`.

runtest262.mjs
```diff
@@ -19,7 +19,8 @@ const result = runTest262({
   test262Dir: 'test262',
   polyfillCodeFile: 'dist/script.js',
   expectedFailureFiles,
-  testGlobs: process.argv.slice(2)
+  testGlobs: process.argv.slice(2),
+  failOnUnexpectedPasses: true
```
justingrant, line 23: I don't think we need `failOnUnexpectedPasses`.... instead we can just change the test runner to always fail on unexpected passes. Anything unexpected should fail the test suite.

test/validStrings.mjs
```diff
@@ -7,34 +7,9 @@
...
-  'islamic-tbla',
-  'islamic-civil',
-  'islamic-rgsa',
-  'iso8601',
-  'japanese',
-  'persian',
-  'roc',
-  'islamicc'
-];
+const timezoneNames = Intl.supportedValuesOf('timeZone');
```
justingrant, line 11: Is it OK that this test won't run in Node 18?  Or would we use a polyfill?  Or should we conditionally use supportedValuesOf if available, and use the old hardcoded list if not?

justingrant (approved):
Overall looks great!  A few minor questions/comments.

lib/calendar.ts
```diff
@@ -2283,12 +2284,9 @@ class NonIsoCalendar implements CalendarImpl {
   ): Temporal.PlainDate {
     const overflow = ES.ToTemporalOverflow(options);
     const cache = new OneObjectCache();
-    // Intentionally alphabetical
-    const fields = ES.PrepareTemporalFields(
-      fieldsParam,
-      ['day', 'era', 'eraYear', 'month', 'monthCode', 'year'],
-      ['day']
-    );
+    const fieldNames = this.fields(['day', 'month', 'monthCode', 'year']) as readonly ES.AnyTemporalKey[];
```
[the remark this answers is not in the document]
justingrant: > note that `AnyTemporalKey` is too broad here, because it includes time units, timeZone/offset, calendar, etc. Should narrow it to just calendar units?

Actually, the current spec sometimes passes time units down to `fields`, so `AnyTemporalKey` is probably OK for now. I did ask @ptomato if the spec should be changed to disallow time units, but that's beyond the scope of this PR.

> Instead of casting the return value here, should we change the type declaration of the fields method parameter and return value?

Given the notes above, I take back this suggestion. Let's just leave as-is, and depending on how the time units thing works out, we can always follow up with another PR to change the types.

lib/calendar.ts
justingrant: Never mind. See https://github.com/js-temporal/temporal-polyfill/pull/210#discussion_r1070726942.

lib/calendar.ts
```diff
@@ -2314,16 +2313,12 @@ class NonIsoCalendar implements CalendarImpl {
...
     const cache = new OneObjectCache();
-    const fields = ES.PrepareTemporalFields(
-      fieldsParam,
-      ['day', 'era', 'eraYear', 'month', 'monthCode', 'year'],
-      ['day']
-    );
+    // For lunisolar calendars, either `monthCode` or `year` must be provided
+    // because `month` is ambiguous without a year or a code.
+    const fieldNames = this.fields(['day', 'month', 'monthCode', 'year']) as readonly ES.AnyTemporalKey[];
+    ArrayPrototypeSort.call(fieldNames);
```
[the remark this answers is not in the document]
12wrigja: AFAICT, ES.Call does some runtime validation of arguments lists and makes sure that the function-thing is really a function, both of which are satisfied by static type-checking here instead.

Also, that function comes from es-abstract, which we dropped the dependency on a while back.

lib/calendar.ts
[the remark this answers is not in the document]
12wrigja: Thanks for following up on this. I'm going to resolve these comments and we can revisit if the spec changes.

Was that ask in an issue over on the spec repo? If not, should it be so we can track it here?

runtest262.mjs
12wrigja: Right - this will be fixed by updates to the runner in https://github.com/js-temporal/temporal-test262-runner. I'll wait for that to settle first and sync/rebase all this before merging.

lib/ecmascript.ts
12wrigja: Done.

test/validStrings.mjs
12wrigja: Looking at what this test does, I'm not sure what value we'd get from running it in multiple versions of Node unless somehow the Temporal parsers rely on something node-version-specific?

lib/calendar.ts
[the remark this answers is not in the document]
justingrant: Sorry, no issue for that, but @ptomato said in chat that he was planning to look at removing those time units as part of other ongoing work in the other repo.

test/validStrings.mjs
justingrant: Makes sense to me. Do we need to do anything special so that these tests won't fail when we run Demitasse tests in earlier Node versions?

justingrant: Or should we just run Demitasse tests only in the latest Node and not bother running across multiple versions, now that Demitasse tests are only a tiny fraction of things covered by Test262?

justingrant:
All looks good, but before merging this I think we should probably update the Test262 runner (and use it in this PR) so that unexpected Test262 test passes will cause a CI break.

12wrigja:
> All looks good, but before merging this I think we should probably update the Test262 runner (and use it in this PR) so that unexpected Test262 test passes will cause a CI break.

That's the plan :)

test/validStrings.mjs
12wrigja: I don't think this particular test is worth running in earlier versions, and currently validStrings is being run separately from the other Demitasse tests: https://github.com/js-temporal/temporal-polyfill/blob/445bfd3d4efabec2d23372c9cc92f8d8619c822e/.github/workflows/test.yml#L77-L88

Once this PR lands I'll look over the remaining tests in Demitasse and see whether there's any strong reason to keep running them in older Node versions.

lib/calendar.ts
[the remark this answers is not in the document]
12wrigja: Gotcha. I'll resolve this then and presume that if this ends up changing that it will be reflected in another rebase in the future.

12wrigja:
Hooray! The new test runner changes caught things:
https://github.com/js-temporal/temporal-polyfill/actions/runs/3953451177/jobs/6769740242

justingrant:
> Hooray! The new test runner changes caught things:
> https://github.com/js-temporal/temporal-polyfill/actions/runs/3953451177/jobs/6769740242

Cool!  Glad we waited to merge.

12wrigja:
I'd like to move forward with merging these changes as I have another couple rebase PRs in the works. Any objections?

justingrant:
Nope!  Go ahead and merge when ready!

PR #212, opened 2023-01-18
Parameterize test262 config.

12wrigja:
This will allow it to be re-used over in the runner repo for basic CI testing.


justingrant:
I think there may be a simpler way to solve this problem without changing this repo. Let's continue the convo over in https://github.com/js-temporal/temporal-test262-runner/pull/6#discussion_r1072986445

PR #214, opened 2023-01-25
Even more rebase updates

12wrigja:
These changes start to clean up the expected failure lists we have. Some extra scrutiny on which tests were or were not removed would be helpful, as I'm never entirely certain which tests should work after each change.


lib/ecmascript.ts
```diff
@@ -5317,19 +5326,46 @@ function RoundNumberToIncrement(quantity: JSBI, increment: number, mode: Tempora
   let { quotient, remainder } = divmod(quantity, JSBI.BigInt(increment));
```
ptomato, line 5326: I had some internal tests for RoundNumberToIncrement in the other polyfill in order to more easily test it directly, without going through Temporal entry points - maybe worth porting those over to here?

ptomato (approved):
All looks plausible to me. I was a bit surprised that the first commit only had a change to validStrings, but so did the upstream commit, so I guess the code already worked that way?

index.d.ts
```diff
@@ -1,6 +1,15 @@
 export namespace Temporal {
   export type ComparisonResult = -1 | 0 | 1;
-  export type RoundingMode = 'halfExpand' | 'ceil' | 'trunc' | 'floor';
+  export type RoundingMode =
```
justingrant, line 3: Looks like we also need to make this change in proposal-temporal. @12wrigja do you want to PR this?  If not then I can do it.

justingrant (approved):
Looks great, thanks!

index.d.ts
12wrigja: I can send it - thanks for the reminder.

12wrigja: Opened https://github.com/tc39/proposal-temporal/pull/2489

lib/ecmascript.ts
12wrigja: That sounds useful, though I can't find them in the spec repo. Mind linking to them? I'll rebase them in in another PR.

ptomato: Here they are: https://github.com/tc39/proposal-temporal/blob/main/polyfill/test/ecmascript.mjs#L41-L67

12wrigja: I have no idea why GitHub search refused to show me that until you linked it to me... thanks :) I'll create a follow-up PR for this.

PR #215, opened 2023-01-26
Remove some unnecessary type assertions.

12wrigja:


ptomato (approved):

12wrigja:
I was going to enable a new type-aware eslint rule to flag type assertions that aren't necessary, but it seems to report inconsistent results between my machine and CI, and I'm not yet sure why. I'll send that in a different PR.

PR #216, opened 2023-01-26
Enable no-unnecessary-type-assertions

12wrigja:


12wrigja:
FYI I enabled the type-assertion check as a separate script and action because on my dev machine at least the type-aware linting takes a lot longer than the existing linting (~1.5m per lint run!) and there's some weird bug around caching that showed some results to not show up until I either cloned the repo again or pushed to CI...

justingrant:
Did you take a look at the recommendations [here](https://duncanleung.com/why-slow-plugin-typescript-eslint-performance-issues/) and here to see if that 1.5 minute linting time could be improved?

What I've seen in the past was that many TS compiling/linting perf issues turned out to be caused by a single type declaration. It might be worth some bisect checks to see if there's an easy fix here. If not easy to find or easy to fix, then I think it's OK to split them like you did.

12wrigja:
Thanks for the pointers. Apparently explicitly specifying includes in tsconfig brings the lint time down significantly, but it doesn't then allow mixing the lint rule config for TS and JS otherwise I get  [an error](https://typescript-eslint.io/linting/troubleshooting/#i-get-errors-telling-me-eslint-was-configured-to-run--however-that-tsconfig-does-not--none-of-those-tsconfigs-include-this-file) from ESLint saying that the JS files aren't included in the TSConfig. Splitting up the rule config so that the ts-eslint rules only apply to TS files seems to fix that.

lib/plaindate.ts
```diff
@@ -247,13 +248,13 @@ export class PlainDate implements Temporal.PlainDate {
...
       nanosecond = 0;
     if (temporalTime !== undefined) {
-      temporalTime = ES.ToTemporalTime(temporalTime);
-      hour = GetSlot(temporalTime, ISO_HOUR);
-      minute = GetSlot(temporalTime, ISO_MINUTE);
-      second = GetSlot(temporalTime, ISO_SECOND);
-      millisecond = GetSlot(temporalTime, ISO_MILLISECOND);
-      microsecond = GetSlot(temporalTime, ISO_MICROSECOND);
-      nanosecond = GetSlot(temporalTime, ISO_NANOSECOND);
+      let temporalTimeConverted = ES.ToTemporalTime(temporalTime);
```
justingrant, line 251: Should this be const?

lib/plaindate.ts
```diff
@@ -216,7 +216,9 @@ export class PlainDate implements Temporal.PlainDate {
   toZonedDateTime(item: Params['toZonedDateTime'][0]): Return['toZonedDateTime'] {
     if (!ES.IsTemporalDate(this)) throw new TypeError('invalid receiver');
 
-    let timeZone, temporalTime;
+    let timeZone: Temporal.TimeZoneProtocol;
+    type TimeZoneAndPlainTimeProps = Exclude<typeof item, Temporal.TimeZoneProtocol | string>;
```
justingrant, line 220: to keep changes from temporal-polyfill minimized, maybe move this type declaration up one line, and then leave the single-line `let` statement to declare both variables?

lib/plaindate.ts
```diff
@@ -225,11 +227,10 @@ export class PlainDate implements Temporal.PlainDate {
         // not a TimeZoneProtocol.
         // TODO: should we check for that shape to improve on the (bad) error
         // message that the caller will get from ToTemporalTimeZone?
-        timeZone = ES.ToTemporalTimeZone(item as Temporal.TimeZoneProtocol);
```
justingrant, line 228: General question for changes in this method: why are these changes helpful?  In general, I like minimizing changes from the original proposal-temporal source, so naively I'd prefer leaving variable names the same, but maybe I'm missing some other benefit of the changes?

justingrant (approved):
Thanks, looks good. A few comments inline.

lib/plaindate.ts
12wrigja: I'm assuming you are referring to the change below that uses the name `temporalTimeConverted`? This particular line just removed a cast, but the variable names should be the same as upstream.

Unfortunately, TS won't narrow the type of `temporalTime` when it sees

```ts
temporalTime = ES.ToTemporalTime(temporalTime);
```

because it can't be sure whether any future reassignments that aren't of type `PlainTime` are desired or an error.

The other way to try and keep this closer to upstream would be to add some assertion functions that will force this narrowing at the cost of adding an extra statement in some places:

```ts
assertIsTemporalPlainTime(arg: unknown): asserts arg is Temporal.PlainTime {}

...

temporalTime = ES.ToTemporalTime(temporalTime);
assertIsTemporalPlainTime(temporalTime); // This is OK because ToTemporalTime always returns a PlainTime

// temporalTime is now a PlainTime
```

This still means adding an extra statement which will mess up rebasing a bit (though other changes in this particular chunk would likely mess this up anyways).

lib/plaindate.ts
12wrigja: Done.

lib/plaindate.ts
12wrigja: If we end up using this pattern I'll change it over to const. I've pushed an alternative that you might like more - PTAL.

lib/plaindate.ts
justingrant: > I'm assuming you are referring to the change below that uses the name `temporalTimeConverted`? This particular line just removed a cast, but the variable names should be the same as upstream.

I was mostly wondering about why any of these changes were needed. What was working before this PR that's now failing with the new lint rules?

BTW, interesting idea of using an assertion function. Would it work (and if yes, would we want to do this?) to define a generic no-op assertion function that could be re-used in all cases where we need to let TS know that a variable's type has changed?

```ts
function noOpAsserter<T>(unused: unknown): asserts unused is T {};
```

lib/plaindate.ts
```diff
@@ -248,6 +248,7 @@ export class PlainDate implements Temporal.PlainDate {
       nanosecond = 0;
     if (temporalTime !== undefined) {
       temporalTime = ES.ToTemporalTime(temporalTime);
+      ES.assertIsPlainTime(temporalTime, 'ToTemporalTime above always returns a PlainTime');
```
justingrant, line 251: Will this justification string end up adding to bundle size? If so, then maybe make this a comment?

If we really wanted to be fancy, we could create a custom lint rule that verified that any use of `ES.assert*` functions was preceded by a comment with a specific format, e.g. "// justification: ".  If you've never written a custom lint rule before, it's surprisingly easy.

lib/plaindate.ts
justingrant: Oh, lol I see you've done something like this!

lib/ecmascript.ts
```diff
@@ -3591,65 +3609,74 @@ export function UnbalanceDurationRelative(
...
           months += oneYearMonths;
           years -= sign;
         }
       }
       break;
     case 'week':
-      if (!calendar) throw new RangeError('a starting point is required for weeks balancing');
+      // These conditions are effectively identical (as relativeTo being defined forces calendar to be defined)
+      // but checking for their truthiness here helps TypeScript and avoids the need for casts below.
+      if (!relativeTo || !calendar) throw new RangeError('a starting point is required for weeks balancing');
```
justingrant, line 3633: Should we use an assertion function here as an alternative to changing the `if` condition?

lib/ecmascript.ts
```diff
@@ -120,6 +120,24 @@ const BUILTIN_CALENDAR_IDS = [
   'gregory'
 ];
 
+/*
+ * These functions are all used to help TS understand when variables should change types,
+ * but TS can't or won't infer this automatically. They should be used sparingly, because
+ * if used incorrectly can lead to difficult-to-diagnose problems.
+ */
+/* eslint-disable @typescript-eslint/no-unused-vars, @typescript-eslint/no-empty-function */
+export function assertIsPlainTime(arg: unknown, justification: string): asserts arg is Temporal.PlainTime {}
```
justingrant, line 129: Do we also want to have a generic function for other types?  Many of our type casts use other types.

lib/plaindate.ts
12wrigja: > What was working before this PR that's now failing with the new lint rules?

I could split these PRs up, as there's two groups of changes to sources here:
 - removing casts that are doing nothing (this is what the linter is catching)
  - changing code that uses casts to avoid them and instead fix whatever the underlying problem(s) they might be masking.
  
  These changes are more in the latter bucket, where casts are being used to get the code to compile and might be hiding problems in code. Generally, I'd prefer to avoid having a bunch of casts scattered throughout the code in favour of either fixing the root cause so inference works correctly or using type-assertion functions to fix type-inference in places where there's a limitation, and documenting why that particular assertion is needed.
  
  I think the generic version would work - I'll update the PR with that.

lib/plaindate.ts
12wrigja: > Will this justification string end up adding to bundle size? If so, then maybe make this a comment?

No - terser removes the function and calls entirely because to Terser/JS it looks like an empty function with all it's args unused.

lib/plaindate.ts
justingrant: Makes sense. Esp. given that the no-op functions are removed in prod, I really like the assertion function thing as alternative to littering casts everywhere. Thanks for finding that!

lib/plaindate.ts
justingrant: Cool. One more thought: having the justification string as an argument might cause naive readers to think that the function actually does something with that string, e.g. some sort of runtime validation. I wonder if we should name the function differently (e.g. `changeType` or `noOpAssert` or something like that) and/or remove the justification arg to clarify this?

lib/ecmascript.ts
12wrigja: Sounds good.

These cases actually can benefit from a stricter assertion (assertExists) for which we can also include some optional runtime type-checks (that get compiled out in prod builds).

justingrant (approved):
Looks good!  Nice removal of type assertions.

justingrant:
BTW, make sure to fix the lint issue

lib/plaindate.ts
12wrigja: I renamed this to `uncheckedAssertNarrowedType` as that is what I intend for this function to allow:
 - it's not being checked at runtime
 - ideally, it should allow "narrowing" down a union of types to one of the union's members. I've not quite found a way to do this in the type system, but I'll continue to iterate on this with the help of some teammates over the coming weeks.

lib/ecmascript.ts
12wrigja: Yes - done.

PR #218, opened 2023-02-02
Use a command line parser in runtest262.

12wrigja:
This allows the test262 runner to update expected-failure files via a command line flag:

```
npm run test262 -- --update-expected-failure-files
```

If we use this a lot we could add that to package.json.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,7 @@
       "devDependencies": {
         "@babel/core": "^7.20.5",
         "@babel/preset-env": "^7.20.2",
-        "@js-temporal/temporal-test262-runner": "^0.8.0",
+        "@js-temporal/temporal-test262-runner": "^0.9.0",
         "@pipobscure/demitasse": "^1.0.10",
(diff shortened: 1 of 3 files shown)


justingrant (approved):

justingrant:
> This allows the test262 runner to update expected-failure files via a command line flag:
> 
> ```
> npm run test262 -- --update-expected-failure-files
> ```

Why is the first `--` needed?

Also, is it getting complicated enough that it may be worth adding a separate CI run to verify that our command-line args work OK for this file?  Also OK not to bother. Your call.

12wrigja:
The first `--` is to split the flags between `npm` itself and whatever the underlying script is. See https://docs.npmjs.com/cli/v9/commands/npm-run-script.

For the moment I don't think I'm going to add tests for the CLI, because I don't really know how I'd verify the wiring there without adjusting the script itself. Maybe some small additions to package.json to fail the test262 script if it doesn't see some "telltale" markers for the test runner?

PR #224, opened 2023-02-24
TypeScript: allow Intl.Locale objects in locale list params

lionel-rowe:
This PR allows `locales` parameters to accept `Intl.Locale` objects (or arrays thereof). This already works fine in plain JS, but the current TS types are overly restrictive.

See also: https://github.com/microsoft/TypeScript/issues/52946

> https://tc39.es/ecma402/#sec-canonicalizelocalelist
>
> `CanonicalizeLocaleList` is called on all such `locales` arguments, iterating through them as an array, checking for the `[[InitializedLocale]]` internal slot (used to identify `Intl.Locale` objects), and reading the `[[Locale]]` internal slot if it exists (for `Intl.Locale` objects, this is the string representation of the locale). Thus `'x'.toLocaleUpperCase('es')` behaves identically to `'x'.toLocaleUpperCase(new Locale('es'))`

Note that `Intl.DateTimeFormat` already accepts `Intl.Locale`s — the casting to `as ConstructorParameters<typeof IntlDateTimeFormat>[0]` is only needed until the TS `lib` issue is fixed.

### DIFF
--- a/index-lt-4.7.4.d.ts
+++ b/index-lt-4.7.4.d.ts
@@ -1,0 +1,20 @@
+export * from './index';
+
+declare global {
+  // `Intl.LocalesArgument` type was recently added to TS — polyfilled here via `typesVersions` in `package.json` for
+  // compatibility with TS versions older than v4.7.4.
+  // TODO: Revisit/remove once TS >= v4.7.4 is widespread enough.
+  namespace Intl {
+    /**
+     * The locale or locales to use
+     *
+     * See [MDN - Intl - locales
+     * argument](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl#locales_argument)
+     */
+    type LocalesArgument =
+      | UnicodeBCP47LocaleIdentifier
+      | globalThis.Intl.Locale
+      | readonly (UnicodeBCP47LocaleIdentifier | globalThis.Intl.Locale)[]
+      | undefined;
+  }
+}
--- a/lib/intl.ts
+++ b/lib/intl.ts
@@ -118,7 +118,10 @@ function DateTimeFormatImpl(
   const hasOptions = typeof optionsParam !== 'undefined';
   const options = hasOptions ? ObjectAssign({}, optionsParam) : {};
   // TODO: remove type assertion after Temporal types land in TS lib types
-  const original = new IntlDateTimeFormat(locale, options as globalThis.Intl.DateTimeFormatOptions);
+  const original = new IntlDateTimeFormat(
+    locale as ConstructorParameters<typeof IntlDateTimeFormat>[0],
+    options as globalThis.Intl.DateTimeFormatOptions
+  );
   const ro = original.resolvedOptions();
 
   // DateTimeFormat instances are very expensive to create. Therefore, they will
@@ -169,7 +172,10 @@ DateTimeFormatImpl.supportedLocalesOf = function (
   locales: Params['supportedLocalesOf'][0],
   options: Params['supportedLocalesOf'][1]
 ) {
-  return IntlDateTimeFormat.supportedLocalesOf(locales, options as globalThis.Intl.DateTimeFormatOptions);
+  return IntlDateTimeFormat.supportedLocalesOf(
+    locales as Parameters<typeof IntlDateTimeFormat['supportedLocalesOf']>[0],
+    options as globalThis.Intl.DateTimeFormatOptions
+  );
 };
 
 const propertyDescriptors: Partial<Record<keyof Intl.DateTimeFormat, PropertyDescriptor>> = {
--- a/lib/intl.ts
+++ b/lib/intl.ts
@@ -117,9 +117,10 @@ function DateTimeFormatImpl(
   }
   const hasOptions = typeof optionsParam !== 'undefined';
   const options = hasOptions ? ObjectAssign({}, optionsParam) : {};
-  // TODO: remove type assertion after Temporal types land in TS lib types
   const original = new IntlDateTimeFormat(
+    // TODO: remove type assertion after TS lib types updated per https://github.com/microsoft/TypeScript/issues/52946
     locale as ConstructorParameters<typeof IntlDateTimeFormat>[0],
+    // TODO: remove type assertion after Temporal types land in TS lib types
     options as globalThis.Intl.DateTimeFormatOptions
   );
   const ro = original.resolvedOptions();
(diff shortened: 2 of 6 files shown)


lib/intl.ts
```diff
@@ -118,7 +118,10 @@ function DateTimeFormatImpl(
   const hasOptions = typeof optionsParam !== 'undefined';
   const options = hasOptions ? ObjectAssign({}, optionsParam) : {};
   // TODO: remove type assertion after Temporal types land in TS lib types
-  const original = new IntlDateTimeFormat(locale, options as globalThis.Intl.DateTimeFormatOptions);
+  const original = new IntlDateTimeFormat(
```
12wrigja, line 121: Is it correct that once https://github.com/microsoft/TypeScript/pull/52996 lands and we update our TS version in use we will be able to stop casting the locale param? If so, would you mind adding a TODO here linking against that PR? Thanks.

12wrigja:
As far as I can tell, this symbol is relatively new? https://stackoverflow.com/questions/73637484/namespace-intl-has-no-exported-member-localesargument

Presumably, if we add this then our .d.ts file will be incompatible with downstream ts users who don't have a relatively up to date TS. This is typically solved using [`typesVersions`](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#version-selection-with-typesversions) which we don't currently have set up. Maybe we could extend the copy-types.mjs script to also set this up?

justingrant:
@12wrigja  We've solved this for other built-in types by duplicating the types in our own .d.ts. Add long as the types match exactly, this won't break TSC in later versions of TS.

Seems simple to do the same thing here. What do you think?

12wrigja:
Duplicating is basically what the fallback for typesVersions would do. I'm just suggesting having multiple versions so that we automatically stay up to date as the upstream Intl typings change.

12wrigja:
And realistically wouldn't the typings that end up in the standard .d.ts for TS itself use the Intl symbols instead of a duplicate?

justingrant:
> I'm just suggesting having multiple versions so that we automatically stay up to date as the upstream Intl typings change.

Given that TS will do [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html), what's the advantage of having separate per-TS-version type declaration files that would outweigh the complexity of maintaining separate files?

For example, `dayPeriod`, `dateStyle`, and `timeStyle` are relatively new so we added them into the polyfill's types, but they work fine on later versions of TS because of declaration merging. 

https://github.com/js-temporal/temporal-polyfill/blob/c0f7349a327b68543797d38e045b1fd8c1e0949b/index.d.ts#L1526-L1533

> And realistically wouldn't the typings that end up in the standard .d.ts for TS itself use the Intl symbols instead of a duplicate?

Yep, but as long as the parts of our types match what's in TS's own `lib.*` type declarations, then everything should work fine. Unless I'm missing some important point?

lib/intl.ts
lionel-rowe: Correct - I've added a TODO but referencing the issue not the PR, in case that PR doesn't get merged for some reason

lionel-rowe:
I've duplicated `LocalesArgument` with a TODO to remove in the future.

justingrant (approved):
Thanks for the PR. LGTM!  @12wrigja feel free to merge if you think this change is OK.

12wrigja:
I'd like to verify this works when compiling against TS versions that don't define LocalesArgument before merging.

12wrigja:
Ultimately, the PR in it's current state works (as the .d.ts doesn't explicitly reference `globalThis.Intl.LocalesArgument`), but I did also manage to get this to work in https://github.com/js-temporal/temporal-polyfill/pull/229 using `typesVersions` and I think it has some advantages:

 - It leaves the index.d.ts file clean and enables automatic forward compatibility with any changes to the upstream definition of Intl.LocalesArgument (the real definition of which is decided by our users, not whatever version of TS we compile against).
  - It moves all the older version cruft into a separate .d.ts file, making cleanup easy (just delete the file and the corresponding lines in package.json and done). In this case, what we are effectively doing is providing the typedef for Intl.LocalesArgument to older versions of TS if it's not available in the standard typings.
  - Given that typings don't tend to make backwards-incompatible changes and we really only need additions to the .d.ts, we can continue to use this pattern for other types that we've coped in so long as we can figure out what TS version it should start applying at.
  
  Thoughts @justingrant?
  
  @lionel-rowe I'd be more than happy for you to copy my WIP PR above and fix it up if you'd still like to contribute this fix.

lionel-rowe:
Thanks @12wrigja — I've updated PR with that fix. I've set the threshold at <4.7.4, as 4.7.4 has the type but 4.6.4 is missing it.

index-lt-4.7.4.d.ts
```diff
@@ -0,0 +1,17 @@
+export * from './index';
+
+declare global {
```
12wrigja, line 3: Nit: a comment here about why this is needed and when we should consider cleaning it up would be helpful.

12wrigja (approved):

12wrigja:
I'm happy with this, but would like @justingrant to take a look too.

12wrigja:
After reading over https://twitter.com/atcb/status/1634653474041503744?s=46&t=QSggAfTZ89VDmJRcZoeQ0A, I think we might need some small adjustments to the build so that this works consistently (as typesVersions is only used when exports is not).

 - https://github.com/js-temporal/temporal-polyfill/blob/main/copy-types.mjs needs updating to copy the new .d.ts file to a .d.cts file
  - exports.import in package.json needs to get a "types@<4.7.4" key pointing to the new .d.ts file
   - exports.require in package.json needs the same key, but pointing to the copied .d.cts file.
   
   Based on https://github.com/microsoft/TypeScript/pull/50890#issuecomment-1256528709.

justingrant (approved):
LGTM. Should we also consider a similar trick for our existing re-declaration below? 

```ts
    // TODO: remove the props below after TS lib declarations are updated
    dayPeriod?: 'narrow' | 'short' | 'long';
    dateStyle?: 'full' | 'long' | 'medium' | 'short';
    timeStyle?: 'full' | 'long' | 'medium' | 'short';
```

jnoordsij:
Just ran into this, as I encountered this typing issue myself when trying to pass some `Intl.LocalesArgument` typed variable to the `toLocaleString` function. Is there any chance of this still getting merged at some point?

justingrant:
@ptomato is this issue now fixed given your latest typings changes?

ptomato:
Yes, I think it's now covered regardless of TypeScript version, because we do
```ts
type LocalesArgument = ConstructorParameters<typeof Intl.DateTimeFormat>[0];
```
So I'll close this, but feel free to reopen it if I got that wrong.

PR #229, opened 2023-03-02
Use Intl.LocalesArgument in typings.

12wrigja:


12wrigja:
Closing as I asked the author of https://github.com/js-temporal/temporal-polyfill/pull/224 if they'd like to clone-and-own this, and they graciously did so.

PR #230, opened 2023-03-22
Temporarily work around Test262 tests broken by V8 bug #1416538

12wrigja:
V8 broke Test262 by releasing a regression that caused the output of `Intl.DateTimeFormat.p.format` to differ from `Intl.DateTimeFormat.p.format`. See https://bugs.chromium.org/p/chromium/issues/detail?id=1416538.

Until V8 fixes itself and those fixes get into CI, we'll skip the now-broken tests.

### DIFF
--- a/test/expected-failures-cldr42.txt
+++ b/test/expected-failures-cldr42.txt
@@ -1,0 +1,13 @@
+# Failures in this file are expected to fail for all Test262 tests. To record
+# expected test failures for the transpiled or optimized builds of the polyfill,
+# see expected-failures-es5.txt and expected-failures-opt.txt respectively.
+
+# # A regression in V8 (https://bugs.chromium.org/p/chromium/issues/detail?id=1416538)
+# caused these tests to fail. We don't know when a fix is expected, but when it's fixed then
+# these tests should start passing again.
+staging/Intl402/Temporal/old/date-time-format.js
+staging/Intl402/Temporal/old/datetime-toLocaleString.js
+staging/Intl402/Temporal/old/instant-toLocaleString.js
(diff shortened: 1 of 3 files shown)


test/expected-failures-cldr42.txt
```diff
@@ -0,0 +1,13 @@
+# Failures in this file are expected to fail for all Test262 tests. To record
+# expected test failures for the transpiled or optimized builds of the polyfill,
+# see expected-failures-es5.txt and expected-failures-opt.txt respectively.
+
+# A breaking change in the way CLDR v42 formats whitespace causes these to fail.
+# We don't know exactly when they will roll out the CLDR change fully, but when
+# they do these tests should start passing again.
```
justingrant, line 7:
```suggestion
# A regression in V8 (https://bugs.chromium.org/p/chromium/issues/detail?id=1416538)
# caused these tests to fail. We don't know when a fix is expected, but when it's fixed then
# these tests should start passing again.
```

justingrant (approved):
Needs to update a comment. CLDR 42 was the reason why V8 decided to poorly revert part of what CLDR 42 did, but the cause is V8 and it's a code bug.

Also you may want to change the file name to something a bit more generic, e.g. "failures after node 18" in case we get more issues like this.

test/expected-failures-cldr42.txt
12wrigja: Done.

12wrigja:
I'm going to defer changing the file name because I'm starting to think there should be a better way to specify what version a specific expected-failure list applies to, instead of having to adjust logic in runtest262.mjs.

PR #234, opened 2023-04-08
Simplify some redundant BigInt operations.

12wrigja:
ToBigInt is somewhat redundant with the JSBI.BigInt function (which implements the BigInt constructor behavior).

The ToBigInt conversion in ToTemporalInstant is unnecessary as JSBI.multiply will always return a JSBI, and this is passed to the ToBigInt AO as part of the Temporal.Instant constructor ([spec ref](https://tc39.es/proposal-temporal/#sec-temporal.instant)).


justingrant (approved):
Thanks!

Issue #236, opened 2023-04-12
Rebase tool feedback


justingrant:
After running the rebase tool (see https://github.com/js-temporal/temporal-polyfill/blob/main/tools/rebasing.md), I ran into some problems and have some feedback:

* Should document which commands to run from the tools folder and which to run from the root
* If test.sh fails, should be able to continue afterwards, and ideally a more graceful error message
* Might be nice for the docs to remind me to `chmod +x test.sh` to avoid failing the rebase later. 
* Should document how to handle updates to ecmascript.mjs, e.g. `git rm ...` and `git add ...`.
* The alias command in the docs doesn't work on MacOS zsh shell; when I run `trt` it tries to run the commit hash, not the tools script. Is there another way to write that alias that would be more compatible?

justingrant:
More: should document manual recovery procedure in case things can't continue. Here's some content from James as a starting point 

If that fails, you could continue the rebase manually (`git rebase --continue`). If you want to "finish" the rebase without actually going through all the remaining commits, that can be done by:

* `git rebase --edit-todo`, which will pop open an editor with a bunch of lines in it - these are all the actions left to do in the rebase. Delete them all, save the file, and close the editor. :)
* then, once git is happy with the current state of the working tree, run `git rebase --continue` and the rebase will finish.

Note that it will dump you into a detached head state (if i recall properly), so you'll probably want to make a branch at that commit to be able to refer to it (`git checkout -b <new branch name>`)

PR #238, opened 2023-04-13
Migrate commits from proposal-temporal 2023-04-12

justingrant:
Porting my first batch of commits from proposal-temporal using @12wrigja's [awesome rebase tool](https://github.com/js-temporal/temporal-polyfill/blob/main/tools/rebasing.md) for porting commits from proposal-temporal repo.

Lots of feedback on the tool is in #236.

If this PR looks good then I can migrate more commits later this week. 

I fixed up a mistake I made in one commit and in the process lost @ptomato's authorship of it. Whoops, sorry Philip!

### DIFF
--- a/lib/ecmascript.ts
+++ b/lib/ecmascript.ts
@@ -17,7 +17,6 @@ const ObjectAssign = Object.assign;
 const ObjectCreate = Object.create;
 const ObjectDefineProperty = Object.defineProperty;
 const ObjectGetOwnPropertyDescriptor = Object.getOwnPropertyDescriptor;
-const ObjectIs = Object.is;
 const ReflectApply = Reflect.apply;
 
 import { DEBUG } from './debug';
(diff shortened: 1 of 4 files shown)


12wrigja (approved):

lib/ecmascript.ts
```diff
@@ -5959,11 +5955,13 @@ export function CompareISODate(y1: number, m1: number, d1: number, y2: number, m
   return 0;
 }
 
-function NonNegativeModulo(x: number, y: number) {
-  let result = x % y;
-  if (ObjectIs(result, -0)) return 0;
-  if (result < 0) result += y;
-  return result;
+function NonNegativeBigIntDivmod(x: JSBI, y: JSBI) {
+  let { quotient, remainder } = divmod(x, y);
```
ptomato, line 5969: Are you sure that our JSBI `divmod` function has the same semantics as `bigInteger.divmod`? (Just checking; I have no idea)

ptomato (approved):

lib/ecmascript.ts
justingrant: Hmmm, I sure hope so given how much we use divmod in this polyfill. @12wrigja you're the JSBI expert, do you know?

justingrant: Actually, I wrote a quick test to confirm that results are identical. https://codesandbox.io/s/jsbi-divmod-test-v9zbbn?file=/src/index.js

So I'll go ahead and merge tomorrow morning.

12wrigja: I had planned to do the same thing today - I vaguely recall having done something like that when I first needed to reimplement bigInteger.divmod, but didn't have the script or results handy.

Is that script something long-lived that we could link to in a comment so we can easily refer back to it?

PR #239, opened 2023-04-13
Fix shell quoting that stopped the alias command from working in zsh.

12wrigja:


12wrigja:
As mentioned in #236, the existing command didn't work in zsh because it needs to be properly quoted there. Not really sure why it worked in bash to begin with.

ptomato:
> As mentioned in #236, the existing command didn't work in zsh because it needs to be properly quoted there. Not really sure why it worked in bash to begin with.

I think `$()` is its own quoting in bash, apparently not in zsh?

justingrant (approved):
👍

PR #240, opened 2023-04-14
Improvements to rebase tool.

12wrigja:
This addresses some of the points raised in #236:

- Includes an explanation of how to handle updates to polyfill/lib/ecmascript.mjs from upstream
- new abort and finish commands that help end rebasing prematurely, throwing away or keeping your current rebase progress respectively.
-  > When a commit only updates the spec (no polyfill files) the tool still stopped.
   This was a bug in the way calls to `git rebase --continue` were being handled - handling is now centralized to be consistent.
- The tool will attempt to clean up the `.temporal_rebase_tool` folder it creates when it thinks the rebase is done.
- Test262 no-op update handling might have improved - we check to see if the upstream submodule is changing before running the "handle test262 cases" logic, but I'm not 100% that will fix all the problems.
 - some more docs on what commands are useful during rebases
 - better delineation between what parts of the tool output are actually from underlying git command invocations and what parts are the tool itself.

### DIFF
--- a/tools/rebase-upstream-commits.ts
+++ b/tools/rebase-upstream-commits.ts
@@ -8,6 +8,7 @@
 
 import * as child_process from 'child_process';
 import * as fs from 'fs';
+import * as fsP from 'fs/promises';
 import path from 'path';
 import * as process from 'process';
 import yargs from 'yargs';
@@ -32,6 +33,13 @@ function storeInConfigDir(filename: string, content: string | undefined) {
 
 function readFromConfigDir(filename: string): string {
   return fs.readFileSync(`${currentExecOpts.rebaseToolConfigDir}/${filename}`, UTF8);
+}
+
+function cleanupConfigDir(): void {
+  log('cleaning up tool local storage');
+  return fs.rmSync(currentExecOpts.rebaseToolConfigDir, {
+    recursive: true
+  });
 }
 
 function runInRepository(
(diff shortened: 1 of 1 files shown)


tools/rebase-upstream-commits.ts
```diff
@@ -260,51 +276,86 @@ function continueRebasing() {
...
+    if (filesInUpstreamCommit.includes('polyfill/lib/ecmascript.mjs')) {
+      log(`
+Changes to the ecmascript.mjs file in the upstream repository are never
+automatically rebased onto lib/ecmascript.ts, because the files don't
+share history and git doesn't think they are at all similar.
+
+To resolve this conflict you will need to manually inspect and re-apply
+changes from polyfill/lib/ecmascript.mjs over to lib/ecmascript.ts. Once
+done, remove polyfill/lib/ecmascript.mjs (using git rm) and add the new
+changes to lib/ecmascript.ts (using git add).
```
justingrant, line 312: Might be nice to add copy/pasteable command lines here.

```suggestion
done, remove the mjs file from the commit using:
  git rm polyfill/lib/ecmascript.mjs
Then stage the TS version of that file using:
  git add lib/ecmascript.ts 
At that point, after any other changes are staged too, you can:
  trt continue
```

tools/rebase-upstream-commits.ts
```diff
@@ -503,4 +554,68 @@ yargs(hideBin(process.argv))
...
+  tool's "finish" command
+
+Prefer using this tool's commands to finish the rebase (either via "continue",
+"abort" or "finish") so the local file storage this tool uses will be cleaned
+up.
+
+This tool runs any git commands that might have interactivity directly, and so
+lots of the output printed when running tool commands will actually be from git
+, not from this tool. To make it easy to distinguish the two, any output this
+tool prints is prefixed with [Temporal Rebase Tool].
```
justingrant, line 617:
```suggestion
lots of the output printed when running tool commands will actually be from
git, not from this tool. To make it easy to distinguish the two, any output
this tool prints is prefixed with [Temporal Rebase Tool].
```

justingrant (approved):
Amazing!  Can't wait to try out the new tool.

PR #244, opened 2023-04-17
Port commits from proposal-temporal 2023-04-16

justingrant:
Big batch of migrated commits!  I think this is about 40% of the remaining commits, although the huge ones like the string slot changes are still ahead. 

Thanks again @12wrigja for the great tools to make this easier.


Ms2ger (approved):
It seems like there's a fair number of these commits that don't change expectations, though they seems to be observable. Are those indicative of missing coverage?

justingrant:
Yes, I think that every observable commit that doesn't also change the expected failures file is probably a sign off a possible test gap.

Maybe it's possible in some cases that the tests were deterred? @ptomato may know.

I'll open an issue in the proposal repo to investigate this further.

lib/ecmascript.ts
```diff
@@ -1307,12 +1336,13 @@ export function PrepareTemporalFields<
   RequiredFields extends ReadonlyArray<FieldKeys> | FieldCompleteness
 >(
   bag: Partial<Record<FieldKeys, unknown>>,
-  fields: ReadonlyArray<FieldKeys>,
+  fields: Array<FieldKeys>,
```
12wrigja, line 1339: Does this break type inference for the return type?

If it does, then I'd be tempted to lie about the type inside this function and pretend it is still readonly outside.

For the most part the field ordering probably doesn't matter (in the type-system) in every other call, and it is unfortunate there isn't a good way to model that in the typesystem.

lib/ecmascript.ts
```diff
@@ -1364,8 +1394,9 @@ export function ToTemporalTimeRecord(
   bag: Partial<Record<keyof TimeRecord, string | number | undefined>>,
   completeness: FieldCompleteness = 'complete'
 ): Partial<TimeRecord> {
-  // NOTE: Field order here is important.
+  // NOTE: Field order is sorted to make the sort in PrepareTemporalFields more efficient.
```
12wrigja, line 1397: Couldn't we just remove this comment entirely? Presumably the field order here doesn't matter at all now as it will always be sorted?

ptomato:
Here are the ones where that seems to be the case:
- c8912b05703fe6de698f29deee2cb99549d98c1f
- d1e939f1e9f5cf86020ec7e95505208dd8c2ec11
- 0613780d960baa09d3520772fc9492612b1d448f
- d28c3f8f9bf138958f97921c6a56a115c89396b9
- 92334c7dbe5224ddb9b38557f8253dda9c2abfd9
- 0f7691206e40665a3b5f427922d76f6faf3b5562
- 544f5391b446567b01e6f3267322685c09745900
- bd112438110296af1a0d91309c5ca30782ab22a8
- 11cd552d9c5daaacd9a2a73ddc54bcb19c4e8ab7

All except the last one in the list have to do with observable order of operations. My guess is that each of those individual commits were necessary but not sufficient to make one of the test262 `order-of-operations.js` tests pass, so that's why no tests were removed from the expected failures list in each of those commits.

If you can identify the normative PR in the proposal-temporal repo that corresponds to a commit in this repo, I have been fairly consistently linking the test262 PRs that correspond to them (and if I forgot, there's likely a commit in the PR labeled "Update test262" so you can see which test262 commit it points to)

lib/ecmascript.ts
```diff
@@ -6139,24 +6201,15 @@ function GetOption<
...
-    throw new RangeError(`${String(property)} must be between ${minimum} and ${maximum}, not ${value}`);
-  }
-  return MathFloor(value);
+export function IsBuiltinCalendar(id: string): id is BuiltinCalendarId {
+  return BUILTIN_CALENDAR_IDS.includes(ASCIILowercase(id));
 }
 
-export function IsBuiltinCalendar(id: string): id is BuiltinCalendarId {
-  return ArrayIncludes.call(BUILTIN_CALENDAR_IDS, id);
+export function ASCIILowercase<T extends string>(str: T): T {
```
12wrigja, line 6208: Why not just .lowercase() ? If non-ASCII unicode is supposed to be left alone, can we at least document why?

12wrigja:
@justingrant I wonder if something that might be useful would be a "test262 results" comparsion system built into the rebase tool where we could gather up the expected-failures lists from both versions and compare them to see whether there are tests that should have been re-enabled (or disabled) but didn't for some reason? It would probably mean having to maintain another expected-failures-like list of test cases we know should be different from upstream "for reasons", but might make rebasing easier overall?

lib/ecmascript.ts
justingrant: The return type is still inferred. What it \*does\* do is allow callers to call it with a non-readonly array, which might make them forget to put `as const` on the array input. But if the caller types the input properly then the output will get the correct inferred type. I admit I am a little nervous about doing mutation on an input declared to be readonly, but also fine to do the lying thing too. 

How do you want to proceed?

lib/ecmascript.ts
```diff
@@ -1120,20 +1157,19 @@ export function ToRelativeTemporalObject(options: {
       'second',
       'year'
     ] as const);
+    type FieldNamesWithTimeZoneAndOffset = ArrayWithNewKeys<typeof fieldNames, 'timeZone' | 'offset'>;
+    (fieldNames as FieldNamesWithTimeZoneAndOffset).push('timeZone', 'offset');
```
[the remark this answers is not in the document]
justingrant: This relates to our chat convo about Array and String prototype methods. Let's resolve there and then can change this (and others like it) if needed.

lib/ecmascript.ts
```diff
@@ -889,82 +956,61 @@ export function ToTemporalOffset(
...
+    if (ToString(digitsValue) !== 'auto') {
+      throw new RangeError(`fractionalSecondDigits must be 'auto' or 0 through 9, not ${digitsValue}`);
+    }
+    return 'auto';
+  }
+  const digitCount = MathFloor(digitsValue);
+  if (!NumberIsFinite(digitCount) || digitCount < 0 || digitCount > 9) {
+    throw new RangeError(`fractionalSecondDigits must be 'auto' or 0 through 9, not ${digitsValue}`);
+  }
+  return digitCount as 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9;
```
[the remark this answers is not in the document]
justingrant: Yeah, might be a good candidate for naming. I'll add one.

lib/ecmascript.ts
```diff
@@ -512,7 +562,7 @@ export function ParseTemporalTimeString(isoString: string) {
     millisecond = ToIntegerOrInfinity(fraction.slice(0, 3));
     microsecond = ToIntegerOrInfinity(fraction.slice(3, 6));
     nanosecond = ToIntegerOrInfinity(fraction.slice(6, 9));
-    annotations = match[15];
+    annotations = match[14];
```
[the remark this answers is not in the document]
justingrant: Seems like a good idea to do this upstream and then port it back here. Although (just to check) are there some cases where the same number has different meanings depending on the string?  Would want to make sure that names make things better not worse. Although can figure that out in the later PR, don't need to know now.

justingrant (commented):
Thanks for the review!  A few responses and notes below. 

Also, do you know if there's a trick to keeping the original committer's name on a commit when rebasing using `-i`?  The last few times I tried this, it lost Philip on the updated commit.

justingrant:
> I wonder if something that might be useful would be a "test262 results" comparison system built into the rebase tool where we could gather up the expected-failures lists from both versions and compare them to see whether there are tests that should have been re-enabled (or disabled) but didn't for some reason? It would probably mean having to maintain another expected-failures-like list of test cases we know should be different from upstream "for reasons", but might make rebasing easier overall?

The upstream repo has essentially no expected failures, except weird cases like V8 bugs and for those this repo should fail too.

Also, the tests are somewhat out of sync because upstream the Test262 commit is usually the last one in the PR, but here I've been removing tests from the expected failures file for each commit that needs them. 

So I'm not sure how what you describe will be helpful. But maybe I'm missing something?

FWIW, dealing with Test262 using the test runner option you built has been a huge help. I just run every Test262 with that option set. So Test262 is essentially a non-issue for the porting workflow... adds zero extra work for me, because my workflow is this:

0. Open two terminal tabs, and set up the alias in Tab 1 and start the rebase there. Note that I don't run tests using `--exec=./test.sh` because that wastes time given that I don't commit until all tests are passing. Below is what I do for each step.
1. Resolve merge conflicts in editor
2. Make ecmascript.mjs => ecmascript.ts changes in the editor. This is, by far, the most time consuming (and error-prone) part. If we could turn these changes into merge conflicts, then that would be a massive time savings!
3. Tab 2: Run `npm run prune && npm test && npm run test262 -- --update-expected-failure-files` 
4. Tab 1: While tests are running, do `git add` in terminal tab 2 for all changed files (and the expected failure file if needed). If tests fail then I can `git rm` to fix the problems.
5. Tab 2: If tests pass, then `git rm /polyfill/lib/ecmascript.mjs`
6. Tab 1: `trt continue`, GOTO 1

After doing this 50+ times I'm getting pretty fast at it. If we can fix the ecmascript thing to be merge conflicts then I could probably 2x+ my pace.

lib/ecmascript.ts
```diff
@@ -329,6 +350,35 @@ function getIntlDateTimeFormatEnUsForTimeZone(timeZoneIdentifier: string) {
...
+  source: T | undefined,
+  excludedKeys: K[],
+  excludedValues?: unknown[]
+) {
+  if (typeof source === 'undefined' || source === null) return;
+
+  const keys = Object.getOwnPropertyNames(source) as (keyof T)[];
+  for (const nextKey of keys) {
+    if (excludedKeys.some((e) => Object.is(e, nextKey))) continue;
+    if (Object.prototype.propertyIsEnumerable.call(source, nextKey)) {
```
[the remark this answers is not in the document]
justingrant: After discussing offline, we'll leave this (and similar cases above and below) as-is for now, and then when we're caught up to proposal-temporal then we'll probably remove these across the repo.

lib/ecmascript.ts
justingrant: Decided to do this comment upstream. https://github.com/tc39/proposal-temporal/pull/2559

lib/ecmascript.ts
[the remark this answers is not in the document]
justingrant: https://github.com/tc39/proposal-temporal/issues/2560

lib/ecmascript.ts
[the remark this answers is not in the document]
justingrant: Removed the numbers in favor of a reference to `Temporal.ToStringPrecisionOptions['fractionalSecondDigits']`

lib/ecmascript.ts
12wrigja: I didn't know there was a fast-path. Happy to keep the comment as-is.

lib/ecmascript.ts
12wrigja: My recollection from working on the PrepareTemporalFields typing is that if the inference degrades a bunch of stuff will break loudly, so it shouldn't be a problem in the future so long as it still happens most of the time now. Let's leave this as-is and we can always fix it up later if we need to by either breaking the readonly contract here or possibly by using a sorted copy here somehow.

12wrigja (approved):

PR #246, opened 2023-04-23
Port more commits from proposal-temporal 2023-04-22

justingrant:
Here's another batch of 35 upstream commits!  Note that these commits are rebased on upstream's [`porting`](https://github.com/tc39/proposal-temporal/commits/porting) branch. This branch was produced by running commits through a [script](https://gist.github.com/justingrant/80ec66b882059c05ef556215ba185041) that converted ecmascript.mjs to exported functions and limited commits to only files that the prod polyfill needed. This made porting much, much faster and easier because ecmascript.mjs changes now show up as merge conflicts. After upstream catches up to https://github.com/tc39/proposal-temporal/pull/2557, then this branch will no longer be needed.

I ended up reverting the perf and bundle size optimizations we'd been using previously (e.g. `calendar.day(date)` instead of `GetMethod` followed by `Call`. This will make it easier to port later commits. We can add back these optimizations once the number of upstream commits (and hence the porting workload) slows down.

This PR stops right before @ptomato's gargantuan PRs like the string slots change.  The next porting PR will include those changes.

There's only ~30 expected failures remaining (before we advance Test262 to the latest in preparation for the next batch of ports, which will add 600+ new expected failures). Some of these failures might represent mistakes in porting commit merges (instead of tests that I haven't caught up with in porting commits). From almost all of them look like tests of code that will be changed radically or removed in later upstream commits. And none of them looked like critical code paths; they're problems like RangeError vs. TypeError and order of operations. So unless anyone objects, instead of diagnosing why each one is still failing, I'm planning going to leave those expected failures as-is for now, and then if any still remain after the next batch is ported (hopefully next week), then we can figure out what went wrong.


lib/duration.ts
```diff
@@ -33,16 +33,16 @@ export class Duration implements Temporal.Duration {
...
-    const months = ES.ToIntegerIfIntegral(monthsParam);
-    const weeks = ES.ToIntegerIfIntegral(weeksParam);
-    const days = ES.ToIntegerIfIntegral(daysParam);
-    const hours = ES.ToIntegerIfIntegral(hoursParam);
-    const minutes = ES.ToIntegerIfIntegral(minutesParam);
-    const seconds = ES.ToIntegerIfIntegral(secondsParam);
-    const milliseconds = ES.ToIntegerIfIntegral(millisecondsParam);
-    const microseconds = ES.ToIntegerIfIntegral(microsecondsParam);
-    const nanoseconds = ES.ToIntegerIfIntegral(nanosecondsParam);
+    const years = yearsParam === undefined ? 0 : ES.ToIntegerIfIntegral(yearsParam);
```
12wrigja, line 36: Doesn't ToIntegerIfIntegral already return 0 if it is given undefined?

https://github.com/js-temporal/temporal-polyfill/blob/8ba2a0121ff4458eb98b00d360841df37fc0b403/lib/ecmascript.ts#L239-L240

and

https://github.com/js-temporal/temporal-polyfill/blob/8ba2a0121ff4458eb98b00d360841df37fc0b403/lib/ecmascript.ts#L177-L184

lead me to believe so.

lib/ecmascript.ts
```diff
@@ -164,6 +153,34 @@ function isZero(value: JSBI): boolean {
...
+  obj: T,
+  methodName: M
+): T[M];
+function GetMethod<T extends { [s in M]?: undefined | ((...args: any[]) => unknown) }, M extends string & keyof T>(
+  obj: T,
+  methodName: M
+): T[M] | undefined {
+  const result = obj[methodName];
+  if (result === undefined) return undefined;
+  if (typeof result !== 'function') throw new TypeError(`'${methodName}' must be a function`);
```
12wrigja, line 166: Nit: Maybe put this behind DEBUG as this should be type-checked / guaranteed by tsc?

lib/ecmascript.ts
```diff
@@ -2137,123 +2129,191 @@ function CalendarDateUntil(
...
 
-export function CalendarDayOfWeek(
-  calendar: Temporal.CalendarProtocol,
-  dateLike: CalendarProtocolParams['dayOfWeek'][0]
-) {
-  return ToPositiveIntegerWithTruncation(calendar.dayOfWeek(dateLike));
+export function CalendarDayOfWeek(calendar: Temporal.CalendarProtocol, dateLike: CalendarProtocolParams['era'][0]) {
+  const dayOfWeek = GetMethod(calendar, 'dayOfWeek');
+  const result = Call(dayOfWeek, calendar, [dateLike]);
+  if (typeof result !== 'number') {
```
12wrigja, line 2217: There is so much repetition across all these calendar methods... can we de-dupe this?

lib/duration.ts
justingrant: Nope. You're looking at an out-of-date `ecmascript.ts` above. Upstream changed (https://github.com/tc39/proposal-temporal/commit/b9579ebe33388bf1b68e3011a449cbcd30cd14ed) its implementation of `ToIntegerIfIntegral` to throw on `NaN`, so the `undefined` check is needed here. (And therefore it's what upstream is doing too.)

PR #256, opened 2023-05-30
Enable assertions in Test262 builds.

### LINKED ISSUE #255: assertExists in ecmascript.ts throws if exists
I'm pretty sure the condition on this DEBUG-only function is inverted (should be arg == null):

https://github.com/js-temporal/temporal-polyfill/blob/96c417ccdb389c38fc0da8fd4fdfcd3dcf7e94d9/lib/ecmascript.ts#L149

12wrigja:
Fixes #255, and also enables asserts in our 262 test runs so we hopefully don't run into this problem again.


justingrant (approved):
👍

PR #257, opened 2023-06-05
Port commits 2023-06-04

justingrant:
Port a few more commits from proposal-temporal.

### DIFF
--- a/lib/ecmascript.ts
+++ b/lib/ecmascript.ts
@@ -1,13 +1,12 @@
 const ArrayIncludes = Array.prototype.includes;
 const ArrayPrototypePush = Array.prototype.push;
 const IntlDateTimeFormat = globalThis.Intl.DateTimeFormat;
-const MathMin = Math.min;
-const MathMax = Math.max;
 const MathAbs = Math.abs;
 const MathFloor = Math.floor;
+const MathMax = Math.max;
+const MathMin = Math.min;
 const MathSign = Math.sign;
 const MathTrunc = Math.trunc;
-const NumberIsNaN = Number.isNaN;
 const NumberIsFinite = Number.isFinite;
 const NumberCtor = Number;
 const StringCtor = String;
@@ -17,6 +16,7 @@ const ObjectDefineProperty = Object.defineProperty;
 const ObjectGetOwnPropertyDescriptor = Object.getOwnPropertyDescriptor;
 const ReflectApply = Reflect.apply;
 const ReflectOwnKeys = Reflect.ownKeys;
+const NumberIsNaN = Number.isNaN;
 
 import { DEBUG, ENABLE_ASSERTS } from './debug';
 import JSBI from 'jsbi';
--- a/lib/ecmascript.ts
+++ b/lib/ecmascript.ts
@@ -498,7 +498,29 @@ function FormatCalendarAnnotation(id: string, showCalendar: Temporal.ShowCalenda
   return `[${flag}u-ca=${id}]`;
 }
 
-function ParseISODateTime(isoString: string) {
+// Not a separate abstract operation in the spec, because it only occurs in one
+// place: ParseISODateTime. In the code it's more convenient to split up
+// ParseISODateTime for the YYYY-MM, MM-DD, and THH:MM:SS parse goals, so it's
+// repeated four times.
+function processAnnotations(annotations: string) {
+  let calendar;
+  let calendarWasCritical = false;
+  for (const [, critical, key, value] of annotations.matchAll(PARSE.annotation)) {
+    if (key === 'u-ca') {
+      if (calendar === undefined) {
+        calendar = value;
+        calendarWasCritical = critical === '!';
+      } else if (critical === '!' || calendarWasCritical) {
+        throw new RangeError(`Invalid annotations in ${annotations}: more than one u-ca present with critical flag`);
+      }
+    } else if (critical === '!') {
+      throw new RangeError(`Unrecognized annotation: !${key}=${value}`);
+    }
+  }
+  return calendar;
+}
+
+export function ParseISODateTime(isoString: string) {
   // ZDT is the superset of fields for every other Temporal type
   const match = PARSE.zoneddatetime.exec(isoString);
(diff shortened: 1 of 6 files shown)


lib/ecmascript.ts
```diff
@@ -395,7 +397,7 @@ export function ToObject<T>(value: T): T extends Record<string, unknown> ? T : o
 
 // Adapted from https://github.com/ljharb/es-abstract/blob/main/2022/CopyDataProperties.js
 // but simplified (e.g. removed assertions) for this polyfill to reduce bundle size.
-export function CopyDataProperties<K extends string | symbol, T extends Record<K, unknown>>(
+export function CopyDataProperties<K extends string | symbol | number, T extends Record<K, unknown>>(
```
12wrigja, line 398: Nit: could `string | symbol | number` also be `PropertyKey` (built-in type for this)?

lib/ecmascript.ts
```diff
@@ -6486,10 +6488,15 @@ export function CreateOnePropObject<K extends string, V>(propName: K, propValue:
   return o;
 }
 
-function CopyOptions<T extends { [s in K]?: unknown }, K extends string & keyof T>(options: T | undefined) {
-  const optionsCopy = ObjectCreate(null) as NonNullable<T>;
-  CopyDataProperties(optionsCopy, GetOptionsObject(options), []);
-  return optionsCopy;
+export function SnapshotOwnProperties<T extends { [s in K]?: unknown }, K extends string & keyof T>(
+  source: T | undefined,
+  proto: Record<string | symbol, unknown> | null,
```
12wrigja, line 6491: Record<PropertyKey, unknown> | null ?

lib/ecmascript.ts
```diff
@@ -6486,10 +6488,15 @@ export function CreateOnePropObject<K extends string, V>(propName: K, propValue:
   return o;
 }
 
-function CopyOptions<T extends { [s in K]?: unknown }, K extends string & keyof T>(options: T | undefined) {
-  const optionsCopy = ObjectCreate(null) as NonNullable<T>;
-  CopyDataProperties(optionsCopy, GetOptionsObject(options), []);
-  return optionsCopy;
+export function SnapshotOwnProperties<T extends { [s in K]?: unknown }, K extends string & keyof T>(
```
12wrigja, line 6489: Should k just be PropertyKey?

lib/ecmascript.ts
```diff
@@ -603,12 +617,7 @@ export function ParseTemporalTimeString(isoString: string) {
     millisecond = ToIntegerOrInfinity(fraction.slice(0, 3));
     microsecond = ToIntegerOrInfinity(fraction.slice(3, 6));
     nanosecond = ToIntegerOrInfinity(fraction.slice(6, 9));
-    annotations = match[14];
-    for (const [, critical, key, value] of annotations.matchAll(PARSE.annotation)) {
-      if (key !== 'u-ca' && critical === '!') {
-        throw new RangeError(`Unrecognized annotation: !${key}=${value}`);
-      }
-    }
+    processAnnotations(match[14]); // ignore found calendar
```
12wrigja, line 629: This is being done for the checking side-effects, right? Can you please be a bit more explicit about that in a comment (so it doesn't get cleaned up on accident)?

lib/ecmascript.ts
```diff
@@ -2920,75 +2930,71 @@ interface ToStringOptions {
   roundingMode: ReturnType<typeof ToTemporalRoundingMode>;
 }
 
+function formatAsDecimalNumber(num: number) {
+  if (num <= NumberMaxSafeInteger) return num.toString(10);
+  return JSBI.BigInt(num).toString();
```
12wrigja, line 2939: Do either the bigint ctor or JSBI.BigInt work properly for numbers larger than MAX_SAFE_INTEGER?

12wrigja: Another maybe silly question: negative numbers? Or is this guaranteed to only receive positive numbers?

12wrigja (approved):

lib/ecmascript.ts
justingrant: > Do either the bigint ctor or JSBI.BigInt work properly for numbers larger than MAX_SAFE_INTEGER?

Yep, this is kinda the whole point of big integer types!

> Another maybe silly question: negative numbers? Or is this guaranteed to only receive positive numbers?

It's only called with positive values. We could perhaps add an assertion here, although not sure if that's needed. What do you think?

lib/ecmascript.ts
justingrant: I replaced `string` above with `PropertyKey` and didn't see any ill effects, so I checked in that. The `& keyof T` is required though. Without it, a bunch of TS errors come up.

lib/ecmascript.ts
12wrigja: An assertion seems reasonable to me.

I wasn't sure if the bigint ctor parsing a number that is larger than MAX_SAFE_INTEGER always yields the same value as parsing the equivalent base-10 string, but maybe the safe-ness only applies if you are doing operations across the boundary, not just declaring the numbers?

12wrigja: `BigInt(12345678901234568000) === BigInt('12345678901234568000') // false`

Declaring Number values that are above the MAX_SAFE_INTEGER value can cause problems, so I think we should probably throw an error here and force callers to declare their values as bigints properly?

Does the upstream polyfill have this problem?

ptomato: I think the semantics here are intended to be "format the number exactly as `String(BigInt(number))` would do" because that's the only thing that doesn't produce scientific notation for large numbers. But note that this will be obsolete once we reform Duration calculations, because larger than MAX_SAFE_INTEGER will be disallowed.

justingrant: It sounds like this problem:
a) also exists upstream, so should probably be fixed there first
b) is going away with the upcoming Duration-as-mathematical-values change upstream

So I think I'll just leave it as-is for now, and then when the Duration change is ported we can decide if we need to do anything additional here in this repo.

PR #259, opened 2023-06-11
Support Node 20 and work around Node 20 / ICU 73.1 bug

justingrant:
Work around an ICU 73.1 bug (https://bugs.chromium.org/p/chromium/issues/detail?id=1173158) that was causing Test262 tests to fail. The workaround simply avoids using Intl.DateTimeFormat for Gregorian-based calendars, which enables these calendars to work properly for dates before the 1582-10-15 Gregorian switchover while also using less RAM and CPU.

This PR is stacked on top of #258, so you only need to review the top commit in this PR.

We'll wait to merge this until the corresponding Test262 PR (https://github.com/tc39/test262/pull/3848) is merged and this PR can be rebased on top of that one.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        version: [19.x, 18.x, 16.x, 14.x]
+        version: [20.x, 18.x, 16.x, 14.x]
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -9,7 +9,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        version: [19.x, 18.x, 16.x, 14.x]
+        version: [20.x, 19.x, 18.x, 16.x, 14.x]
```
ptomato, line 12: Maybe we should drop 19.x since it is now neither current nor LTS?

ptomato (approved):
Thanks!

12wrigja (approved):
LGTM - thanks for simplifying the behavior here.

.github/workflows/test.yml
justingrant: Good idea. I just removed 19.x in the latest push.

PR #266, opened 2023-10-13
Rebase some more upstream changes

12wrigja:


ptomato (approved):
These all look correct. :+1:

PR #274, opened 2024-01-12
Rebase upstream changes

12wrigja:


tsconfig.json
```diff
@@ -7,7 +7,7 @@
     "importHelpers": true,
     "inlineSources": true,
     "inlineSourceMap": false,
-    "lib": ["dom", "ES2020"],
+    "lib": ["dom", "ES2022"],
```
justingrant, line 10: What's the benefit of making this change?  Are there new JS APIs that we want to use? 

Also, does this mean that browsers need to support ES2022 to run this polyfill?

justingrant (approved):
Looks great!  So happy that you're cranking through these commits. Soon we'll be all caught up!  Thanks again, James!

tsconfig.json
12wrigja: We need this in order for TS to type-check / recognize the Intl.supportedValuesOf API - without this, TS complains that it doesn't exist on the Intl namespace.

12wrigja: AFAICT, this just changes the types available to TS, and it doesn't change the runtime requirements any more than we already have by using Intl.supportedValuesOf in the polyfill sources (which seems fairly new? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/supportedValuesOf).

12wrigja: Actually, having re-read the code we do fallback if the API isn't available, so I don't think we've bumped any runtime requirements here.

12wrigja:
@justingrant I've noticed a couple backwards-compat issues with some of our tests:
 - earlier versions of Node don't have Intl.supportedValuesOf. I made some changes (see the latest commit) to the Demitasse tests to work around this - are those right?
  - It seems one of these test cases fails on earlier versions of Node because of the `America/Port-au-Prince` IANA identifier. Any ideas on what to do here? This doesn't repro on later versions of Node, so does it count as "too new" similar to the existing carve-out in the test, or should something else happen here? https://github.com/js-temporal/temporal-polyfill/actions/runs/7675453088/job/20921701127?pr=274

justingrant:
@12wrigja I'll take a look later this week. Thanks!

justingrant:
> earlier versions of Node don't have Intl.supportedValuesOf. I made some changes (see the latest commit) to the Demitasse tests to work around this - are those right?

IMO these tests should be skipped instead of just returning success. For example, we can use something like `itOrSkip` function from [this commit.](https://github.com/tc39/proposal-temporal/pull/1245/commits/ee9ee9d642477e1f9c69f8126f3ea8caf00275d1#) 

```js
const itOrSkip = (id) => ((id === 'chinese' || id === 'dangi') && hasOutdatedChineseIcuData ? it.skip : it);
```


> * one of these test cases fails on earlier versions of Node because of the `America/Port-au-Prince` IANA identifier.

I think the fix would be to add this time zone as a special case in `GetAvailableNamedTimeZoneIdentifier`. Specifically, add this line: 
```ts
    Dar_Es_Salaam: 'Dar_es_Salaam',
    Port_Of_Spain: 'Port_of_Spain',
    'Port-Au-Prince': 'Port-au-Prince',  // add this line 
    Isle_Of_Man: 'Isle_of_Man',
    Comodrivadavia: 'ComodRivadavia',
```
This is not Node-version-dependent... it should work in all Node versions. If not, let me know!

12wrigja:
Thanks Justin.

The addition to GetAvailableNamedTimeZoneIdentifier worked perfectly - thanks. I also added the itOrSkip-like function to disable the two new test-cases for older Node runtimes.

PR #280, opened 2024-03-19
Fix relativeTo example values in typings.

12wrigja:


justingrant (approved):
Whoops!  This was my fault from 3 years ago, and no one noticed until now. Nice catch. 

This is present upstream too, so I'll file a PR to fix it there too.

Issue #231, opened 2023-03-28
[Help] JSBI to BigInt in webpack


celluj34:
Hello;

I am not using babel, but webpack, to compile my TS code into JS. My total app size is gigantic due to how I have to create the entrypoints for my app. I am curious if not having to bundle JSBI would improve my bundle size. Is there a way to include the jsbi-to-bigint package in webpack?

celluj34:
Ah, yes, I should have inspected the output first. Sorry about that.

It looks like out of the 293.91KB from `node_modules`, 144.36KB is from `@js-temporal/polyfill/dist/index.esm.js` and 33.2KB is from `jsbi/dist/jsbi-umd.js`.

So it appears that Temporal + JSBI accounts for ~52% of my dependencies. Is this expected? I don't know how complex the polyfill is, but that seems excessive. Is there any advice for reducing the package size?

12wrigja:
> I don't know how complex the polyfill is

We do run the polyfill library source through an optimizer (Terser) before shipping it to try and reduce our bundle size, but:
 - this polyfill does aim to be faithful to the Temporal spec, which itself is large and complicated
 - even when publishing es modules, I don't believe that a bundler/optimizer is able to tree-shake away unused pieces due to the way the polyfill is structured - see  because many Temporal classes can create each other. I explained that somewhat here: https://github.com/js-temporal/temporal-polyfill/issues/235#issuecomment-1504040023.
 - One potentially large chunk of code that you aren't using is the non-ISO calendar support. The Temporal spec only mandates that there be calendar support for the ISO calendar, but we also ship implementations for various other calendars. We could potentially do better here.
 
 If you don't need bigint compatibility, I would be interested to see what code-size numbers you get with the Babel plugin configured. That should drop code-size a fair bit both by removing JSBI from the bundle entirely but I'd expect the source attributed to Temporal to maybe drop in size a bit as well as the rewritten version of all the JSBI operations is likely smaller.

celluj34:
> One potentially large chunk of code that you aren't using is the non-ISO calendar support.

Actually I am using an extremely small chunk of the library - basically just converting strings and numbers into `PlainDates`, doing a few comparisons, adding days, calling .toString, that's about it.

> I would be interested to see what code-size numbers you get with the Babel plugin configured.

[...]

PR #283, opened 2024-05-08
export ./lib folder from npm package

espretto:
Hello 👋 ,
the `lib` folder is part of the npm package as configured in the package.json field `files`. However it is not accessible unless listed in the `exports` field. Currently I have to override my bundlers dependency resolution to access the folder's contents.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,8 @@
         "default": "./dist/index.cjs"
       },
       "./dist/index.cjs"
-    ]
+    ],
+    "./lib/": "./lib/"
   },
   "types": "./index.d.ts",
   "scripts": {


12wrigja:
We originally include our sources in the bundle so they can be referenced using source maps when debugging, not to be consumed directly by bundlers.

Can you help me understand why you can't apply the transpilation to the relevant bundled ESM in `dist`, which is the intended artifact to consume? My recollection is that running the Babel transform for JSBI -> native bigint over that file will get you want you want (based on my previous comments [here](https://github.com/js-temporal/temporal-polyfill/issues/231#issuecomment-1509398519)), though the thread there implies that configuring webpack in practice (vs my toy example) is non-trivial.

espretto:
I didn't realize the ts sources where included for debugging purposes. Makes sense.

Also, I didn't think of using the babel plugin on the ESM build but yes, that should definitely work. Using ESM modules however, is still a little tricky to configure when targeting non-ESM environments.

My intention is to create a partial polyfill by picking only those parts of the polyfill that I need in my codebase. Since tree-shaking is not really an option w/ Temporal's OOP design and its cross-referencing conversion methods, this approach seems to hold the best balance between flexibility/bundle-size and spec compliance.

I am aware that this is branching out quite far from where I started.

espretto:
Here is what my partial polyfill looks like. This would become possible if the `./lib` folder were accessible.
```ts
import * as PlainDateLib from "@js-temporal/polyfill/lib/plaindate";

export namespace Temporal {
  /** @see https://www.typescriptlang.org/docs/handbook/namespaces.html#aliases */
  export import PlainDate = PlainDateLib.PlainDate;
}
```
Of course I'll have to setup my toolchain to be compatible w/ the sourcecode.

12wrigja:
> My intention is to create a partial polyfill by picking only those parts of the polyfill that I need in my codebase. Since tree-shaking is not really an option w/ Temporal's OOP design and its cross-referencing conversion methods, this approach seems to hold the best balance between flexibility/bundle-size and spec compliance.

The intent seems reasonable, but doing this by depending on a released artifact seems like the wrong approach. Instead, given you are effectively forking the polyfill, I would suggest you depend on our repository directly using a git URL dependency: https://docs.npmjs.com/cli/v10/configuring-npm/package-json#dependencies

espretto:
I understand that you wouldn't want to support this approach as the official way of doing things. I'd have to try if using a git-based npm dependency is compatible w/ our build-pipeline and/or if adding the repo as a git submodule would be a viable solution (though I doubt it).

espretto:
Currently the most convenient

12wrigja:
> UPDATE: I am now using npm package [patch-package](https://www.npmjs.com/package/patch-package) to apply this change as part of my postinstall routine. Closing this PR.

Are you able / willing to link to that on GitHub (or is it closed source)? It might be a good example for others who may want to do the same thing.

Furthermore, I'm wondering if you've run into any issues relying on the implementation this way as a result of the implicit cyclic dependencies / intrinsic class registrations and usage in various class methods that transform one Temporal type to another (e.g. https://github.com/js-temporal/temporal-polyfill/blob/44ea889133e0da5b062bbc7660ed77d995430d90/lib/plaindatetime.ts#L388 - I'm wondering if this fails at runtime because ZonedDateTime was never evaluated and so trying to lookup ZonedDateTime at runtime will fail)?

PR #287, opened 2024-06-20
Set package sideEffects to `false`

federicocarboni:
Allows bundlers to tree-shake out the temporal polyfill when it is imported but not used.

If (or when) in the future the polyfill patches the global object `"sideEffects"` may be changed to an array to declare the side effects of a specific entry (e.g. `@js-temporal/polyfill/global` has side effects while other entry points do not).

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
       "./dist/index.cjs"
     ]
   },
+  "sideEffects": false,
   "types": "./index.d.ts",
   "scripts": {
     "test": "npm run build && node --no-warnings --experimental-modules --experimental-specifier-resolution=node --icu-data-dir node_modules/full-icu --loader ./test/resolve.source.mjs ./test/all.mjs",


12wrigja:
There are package-local side-effects when modules are evaluated (example: https://github.com/js-temporal/temporal-polyfill/blob/8fd0dead40de7c31398f4d2d41e145466ca57a16/lib/timezone.ts#L172-L174).

It is possible to convert from one Temporal type to another today using prototype methods which require references to the other module's classes being registered using these side-effects (e.g. `toZonedDateTime`: https://github.com/js-temporal/temporal-polyfill/blob/8fd0dead40de7c31398f4d2d41e145466ca57a16/lib/instant.ts#L151). Would this method still work with this `sideEffect` flag if user code only ever imported `Instant` (as `Instant` does not and can not have a direct import on `ZonedDateTime`)?

federicocarboni:
`sideEffects` does not affect function calls. It is a hint to module bundlers that whole modules may be ignored if none of their exports are used. 

Just in case, I checked built files and they are bit-by-bit identical with and without `sideEffects`.

12wrigja:
Ok - thank you for checking. Seems reasonable to me.

12wrigja (approved):

PR #291, opened 2025-01-11
rebase-tool: Allow 'continue' and 'finish' to work in exec step

ptomato:
I went to do some rebasing and found the rebase tool didn't really work with `--exec` — whenever the tests would fail in the exec step, it was impossible to run `trt continue` or `trt finish`; the tool would just think it was all done, and exit immediately. Here's a fix.

Note that to move on after an exec failure, you now have to have run `trt exec` to mark the last commit as having succeeded. Ideally you wouldn't have to do this and would just be able to run `trt continue`, but I didn't immediately see a way to implement that without a large refactor.

### DIFF
--- a/tools/rebase-upstream-commits.ts
+++ b/tools/rebase-upstream-commits.ts
@@ -217,11 +217,19 @@ function editSequenceFile(file: string) {
 }
 
 function getConflictHash() {
-  const REBASE_CONFLICT_HASH_FILE = `${currentExecOpts.repository}/.git/rebase-merge/stopped-sha`;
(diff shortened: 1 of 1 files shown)


justingrant (approved):
Great!

12wrigja (approved):
Thanks - I never realized that the done list also contained the active step. Now you mention it, I'm not really sure what it was I did to continue rebasing if the automated exec failed. Maybe just manually continued the rebase, and then once that inevitably failed on a non-exec step retain `trt continue`? Regardless, this seems cleaner.

PR #294, opened 2025-02-24
Batch of rebases - early October 2023

ptomato:
This brings the polyfill up to date as of early October 2023. The changes are mainly around removing unnecessary user calls.


justingrant (approved):
Looks great!

One thing to think about is whether `GetMethod` should be replaced by simple property access. Not sure it adds much value and I assume it slows down perf. But to make rebasing easier, we can save changes like this for after we're all caught up to current.

ptomato:
Yep, in fact I'm running into type signature problems with GetMethod a bit down the line from here. But it's not worth going through and replacing; in the current state of the Temporal proposal, GetMethod isn't used at all.

PR #295, opened 2025-02-25
Batch of rebases - late October 2023

ptomato:
Brings up to date as of late October 2023. These are mostly updates to the removal of user code calls, in preparation for the introduction of calendar/timezone method records.


justingrant (approved):
Nice!  That's a lot of commits for so few changes.  Good ROI!  Thanks so much for catching us up.

PR #296, opened 2025-02-26
Update as of November 2023

ptomato:
This one is very few commits but a lot of changed lines. It corresponds pretty much to https://github.com/tc39/proposal-temporal/pull/2519

I'll be the first to admit that the Typescript annotations in `methodrecord.ts` are not very good, and there's a lot of `@ts-expect-error` needed to get it to compile. However, note that method records were removed from the proposal in June 2024. So I didn't want to spend a lot of time getting the types just right, since it'll be deleted in a forthcoming PR anyway.

If there are any easy-to-implement suggestions to improve `methodrecord.ts` I'm happy to incorporate them, but if it's complicated I'd prefer to just leave it. The sloppy code will be gone soon enough!

### DIFF
--- a/lib/methodrecord.ts
+++ b/lib/methodrecord.ts
@@ -1,0 +1,92 @@
+import { GetIntrinsic } from './intrinsicclass';
+import type { Temporal } from '..';
+import type { TimeZonePrototypeKeys } from './intrinsicclass';
+
+type TimeZoneRecordInfo = {
+  Name: 'TimeZone';
+  Protocol: Temporal.TimeZoneProtocol;
+  MethodName: TimeZonePrototypeKeys;
+};
+
+// We switch off type checking when accessing the cached methods on this object.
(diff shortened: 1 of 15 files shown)


lib/methodrecord.ts
```diff
@@ -0,0 +1,155 @@
...
+  MethodName: CalendarRecordMethodNames;
+};
+
+type TimeZoneRecordInfo = {
+  Name: 'TimeZone';
+  Protocol: Temporal.TimeZoneProtocol;
+  MethodName: TimeZonePrototypeKeys;
+};
+
+// We switch off type checking when accessing the cached methods on this object.
```
justingrant, line 29: Agree, no need to worry about making this work if you're gonna remove them soon.

justingrant (approved):
Cool! Thanks for moving the ball fwd.

PR #297, opened 2025-02-27
Various maintenance

ptomato:
- Update temporal-test262-runner dep
- Update deps with security vulnerabilities
- Tweak ignored files in rebase tool


justingrant (approved):
👍

justingrant:
Something wrong with the test running?  Are you sure you don't want to use a release instead of a commit hash for the test runner repo?

ptomato:
Yeah, looks like commit hashes are not yet supported in package.json in Node 14. I was hoping to get https://github.com/js-temporal/temporal-test262-runner/pull/19 but it's not in a release yet. Given the discussion in https://github.com/js-temporal/temporal-test262-runner/issues/17 it sounded like we were leaning towards a commit hash anyway, but it looks like we can't do that if we still want to support Node 14.

I've updated to test262-runner 0.10.1 now, but we really should get a new release out there so we can run the DTF tests. The DTF fixes are coming up soon in the series of rebases.

I also realized that some of the expected DTF failures are actually due to bugs in Node <20, not in the polyfill's DTF implementation. So I've added another commit moving those to a new expectations file.

Mind having another look to see if there's anything that could be improved here?

ptomato:
I'll merge this now, as it's needed for #302.

PR #298, opened 2025-02-27
Rebase through end of 2023

ptomato:
Main change during this period is removing some BigInt arithmetic in Duration calculations.


justingrant (approved):
👍

ptomato:
Thanks for the review!

PR #299, opened 2025-03-01
Rebase up to Februrary 2024

ptomato:
Fewer commits this time, but a much larger diff. This is where the proposal changed to using time durations internally.

### DIFF
--- a/lib/ecmascript.ts
+++ b/lib/ecmascript.ts
@@ -14,13 +14,14 @@ const MathSign = Math.sign;
 const MathTrunc = Math.trunc;
 const NumberIsFinite = Number.isFinite;
 const NumberCtor = Number;
+const NumberIsNaN = Number.isNaN;
+const NumberIsSafeInteger = Number.isSafeInteger;
 const NumberMaxSafeInteger = Number.MAX_SAFE_INTEGER;
 const ObjectCreate = Object.create;
 const ObjectDefineProperty = Object.defineProperty;
 const ObjectGetOwnPropertyDescriptor = Object.getOwnPropertyDescriptor;
 const ReflectApply = Reflect.apply;
 const ReflectOwnKeys = Reflect.ownKeys;
-const NumberIsNaN = Number.isNaN;
 const StringCtor = String;
 const StringPrototypeSlice = String.prototype.slice;
 
@@ -51,6 +52,7 @@ import type {
   TimeZoneSlot
 } from './internaltypes';
 import { GetIntrinsic } from './intrinsicclass';
+import { TruncatingDivModByPowerOf10 } from './math';
 import { CalendarMethodRecord, TimeZoneMethodRecord } from './methodrecord';
 import {
   CreateSlots,
@@ -3900,7 +3902,7 @@ function NanosecondsToDays(
   const dtEnd = GetPlainDateTimeFor(timeZoneRec, end, 'iso8601');
   const date1 = TemporalDateTimeToDate(dtStart);
   const date2 = TemporalDateTimeToDate(dtEnd);
-  let daysNumber = DaysUntil(date1, date2);
+  let days = DaysUntil(date1, date2);
 
   const timeSign = CompareTemporalTime(
     GetSlot(dtStart, ISO_HOUR),
@@ -3917,13 +3919,13 @@ function NanosecondsToDays(
     GetSlot(dtEnd, ISO_NANOSECOND)
   );
 
-  if (daysNumber > 0 && timeSign > 0) {
-    daysNumber--;
-  } else if (daysNumber < 0 && timeSign < 0) {
-    daysNumber++;
+  if (days > 0 && timeSign > 0) {
+    days--;
+  } else if (days < 0 && timeSign < 0) {
+    days++;
   }
 
-  let relativeResult = AddDaysToZonedDateTime(start, dtStart, timeZoneRec, calendar, daysNumber);
+  let relativeResult = AddDaysToZonedDateTime(start, dtStart, timeZoneRec, calendar, days);
   // may disambiguate
 
   // If clock time after addition was in the middle of a skipped period, the
@@ -3934,11 +3936,10 @@ function NanosecondsToDays(
   // back inside the period where it belongs. Note that this case only can
   // happen for positive durations because the only direction that
   // `disambiguation: 'compatible'` can change clock time is forwards.
-  let daysBigInt = JSBI.BigInt(daysNumber);
   if (sign === 1) {
-    while (JSBI.greaterThan(daysBigInt, ZERO) && JSBI.greaterThan(relativeResult.epochNs, endNs)) {
-      daysBigInt = JSBI.subtract(daysBigInt, ONE);
-      relativeResult = AddDaysToZonedDateTime(start, dtStart, timeZoneRec, calendar, JSBI.toNumber(daysBigInt));
+    while (days > 0 && JSBI.greaterThan(relativeResult.epochNs, endNs)) {
+      days--;
+      relativeResult = AddDaysToZonedDateTime(start, dtStart, timeZoneRec, calendar, days);
       // may do disambiguation
     }
   }
@@ -3964,10 +3965,10 @@ function NanosecondsToDays(
     if (isOverflow) {
       nanosecondsBigInt = JSBI.subtract(nanosecondsBigInt, JSBI.BigInt(dayLengthNs));
       relativeResult = oneDayFarther;
-      daysBigInt = JSBI.add(daysBigInt, JSBI.BigInt(sign));
+      days += sign;
     }
   } while (isOverflow);
-  if (!isZero(daysBigInt) && signJSBI(daysBigInt) !== sign) {
+  if (days !== 0 && MathSign(days) != sign) {
     throw new RangeError('Time zone or calendar converted nanoseconds into a number of days with the opposite sign');
   }
   if (!isZero(nanosecondsBigInt) && signJSBI(nanosecondsBigInt) !== sign) {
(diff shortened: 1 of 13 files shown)


lib/ecmascript.ts
```diff
@@ -3956,204 +3899,137 @@ function NanosecondsToDays(
...
-    minutesParam,
-    secondsParam,
-    millisecondsParam,
-    microsecondsParam,
-    nanosecondsParam,
-    largestUnit
-  );
-  if (result === 'positive overflow' || result === 'negative overflow') {
-    throw new RangeError('Duration out of range');
+  const daylen = JSBI.toNumber(dayLengthNs.abs().totalNs);
```
justingrant, line 3923: Why is this use of JSBI not moved into its own module like most other uses are?

Also, should this operation (total ns as a number) be encapsulated in the TimeDuration class?

lib/ecmascript.ts
```diff
@@ -4493,6 +4358,37 @@ export function RejectDuration(
...
+  const timeSign = norm.sign();
+  if (dateSign !== 0 && timeSign !== 0 && dateSign !== timeSign) {
+    throw new RangeError('mixed-sign values not allowed as duration fields');
+  }
+}
+
+function ISODateToEpochDays(y: number, m: number, d: number) {
+  // This is inefficient, but we use GetUTCEpochNanoseconds to avoid duplicating
+  // the workarounds for legacy Date. (see that function for explanation)
+  return JSBI.toNumber(JSBI.divide(GetUTCEpochNanoseconds(y, m, d, 0, 0, 0, 0, 0, 0), DAY_NANOS_JSBI));
```
justingrant, line 4391: Copying questions from similar comment above: 

> Why is this use of JSBI not moved into its own module like most other uses are?
>
> Also, should this operation (total ns as a number) be encapsulated in the TimeDuration class?

lib/ecmascript.ts
```diff
@@ -4807,18 +4577,13 @@ function DifferenceZonedDateTime(
   // dateUntil must be looked up if the instants are not identical, the date
   // parts are not identical, and largestUnit is greater than 'day'
   const nsDiff = JSBI.subtract(ns2, ns1);
```
justingrant, line 4579: Same here: why is this not moved into the new JSBI module?

lib/ecmascript.ts
```diff
@@ -6180,8 +5839,8 @@ export function AddDurationToOrSubtractDurationFromZonedDateTime(
 function RoundNumberToIncrement(quantity: JSBI, increment: JSBI, mode: Temporal.RoundingMode) {
   if (JSBI.equal(increment, ONE)) return quantity;
```
justingrant, line 5840: Same comment as above: why are these JSBI uses not combined into the JSBI module?

lib/ecmascript.ts
```diff
@@ -6229,13 +5888,56 @@ function RoundNumberToIncrement(quantity: JSBI, increment: JSBI, mode: Temporal.
...
+
 export function RoundInstant(
   epochNs: JSBI,
   increment: number,
   unit: keyof typeof nsPerTimeUnit,
   roundingMode: Temporal.RoundingMode
 ) {
-  let { remainder } = NonNegativeBigIntDivmod(epochNs, DAY_NANOS);
+  let { remainder } = NonNegativeBigIntDivmod(epochNs, DAY_NANOS_JSBI);
   const wholeDays = JSBI.subtract(epochNs, remainder);
```
justingrant, line 5941: same

lib/ecmascript.ts
```diff
@@ -6584,20 +6230,12 @@ export function RoundDuration(
...
-        days: deltaDays,
-        nanoseconds,
-        dayLengthNs: dayLengthNumber
-      } = NanosecondsToDays(nanoseconds, intermediate, timeZoneRec));
-      dayLengthNs = JSBI.BigInt(dayLengthNumber);
+      ({ days: deltaDays, norm, dayLengthNs } = NormalizedTimeDurationToDays(norm, intermediate, timeZoneRec));
     } else {
-      ({ quotient: deltaDays, remainder: nanoseconds } = divmod(nanoseconds, DAY_NANOS));
-      deltaDays = JSBI.toNumber(deltaDays);
+      ({ quotient: deltaDays, remainder: norm } = norm.divmod(JSBI.toNumber(DAY_NANOS_JSBI)));
```
justingrant, line 6235: same

lib/ecmascript.ts
```diff
@@ -6754,112 +6374,58 @@ export function RoundDuration(
...
-      allNanoseconds = JSBI.add(allNanoseconds, JSBI.multiply(JSBI.BigInt(milliseconds), MILLION));
-      allNanoseconds = JSBI.add(allNanoseconds, JSBI.multiply(JSBI.BigInt(microseconds), THOUSAND));
-      allNanoseconds = JSBI.add(allNanoseconds, nanoseconds);
-      total = BigIntDivideToNumber(allNanoseconds, JSBI.BigInt(divisor));
-      const rounded = RoundNumberToIncrement(allNanoseconds, JSBI.BigInt(divisor * increment), roundingMode);
-      hours = JSBI.toNumber(JSBI.divide(rounded, JSBI.BigInt(divisor)));
-      nanoseconds = ZERO;
-      minutes = seconds = milliseconds = microseconds = 0;
+      total = norm.fdiv(divisor);
+      norm = norm.round(JSBI.BigInt(divisor * increment), roundingMode);
```
justingrant, line 6394: Should `round` take a Number instead of a JSBI as its first argument?

lib/instant.ts
```diff
@@ -39,17 +39,17 @@ export class Instant implements Temporal.Instant {
   get epochSeconds(): Return['epochSeconds'] {
     if (!ES.IsTemporalInstant(this)) throw new TypeError('invalid receiver');
     const value = GetSlot(this, EPOCHNANOSECONDS);
-    return JSBI.toNumber(ES.BigIntFloorDiv(value, BILLION));
+    return JSBI.toNumber(BigIntFloorDiv(value, BILLION));
   }
   get epochMilliseconds(): Return['epochMilliseconds'] {
     if (!ES.IsTemporalInstant(this)) throw new TypeError('invalid receiver');
     const value = JSBI.BigInt(GetSlot(this, EPOCHNANOSECONDS));
```
justingrant, line 46: Why is value being wrapped by `JSBI.BigInt` here, but it's not in `epochSeconds` above?

lib/math.ts
```diff
@@ -0,0 +1,57 @@
...
+  const mod = sign * NumberParseInt(ReflectApply(StringPrototypeSlice, xStr, [xDigits - p]), 10);
+
+  return { div, mod };
+}
+
+// Computes x * 10**p + z with precision loss only at the end, by string
+// manipulation. If the result is a safe integer, then it is exact. x must be
+// an integer. p must be a non-negative integer. z must have the same sign as
+// x and be less than 10**p.
+export function FMAPowerOf10(xParam: number, p: number, zParam: number) {
```
justingrant, line 40: Is this code ever called in a perf-sensitive context?  Seems like it'd be a lot slower than normal math. Not sure there's anything we can do about it, just wondering,

lib/zoneddatetime.ts
```diff
@@ -105,17 +106,17 @@ export class ZonedDateTime implements Temporal.ZonedDateTime {
   get epochSeconds(): Return['epochSeconds'] {
     if (!ES.IsTemporalZonedDateTime(this)) throw new TypeError('invalid receiver');
     const value = GetSlot(this, EPOCHNANOSECONDS);
-    return JSBI.toNumber(ES.BigIntFloorDiv(value, BILLION));
+    return JSBI.toNumber(BigIntFloorDiv(value, BILLION));
```
justingrant, line 109: This pattern is repeated many times. Should it be encapsulated into bigintmath.ts?

justingrant (approved):
Looks good!  I added a bunch of comments, all are optional suggestions and/or questions for you.

lib/ecmascript.ts
ptomato: I didn't really intend to move _all_ JSBI operations into their own module, just the ones that I'd consider nontrivial utility functions like `abs` and `divmod` because I needed to import those in `timeduration.ts` without introducing a circular dependency.

Looking at `bigintmath.ts` now with fresh eyes, I'm thinking that probably `isZero` and `isNegative` don't really meet that bar of being nontrivial. If we compiled out JSBI to native BigInt ops, they'd reduce to `=== 0n` and `< 0n`. So I think I'll actually remove those from the module and inline them at the call sites.

About total ns as a number, at first glance I think it's best to leave it like this, because total ns is fundamentally a bigint quantity; we can only safely convert it to a number because we know it's less than 48 hours in this specific case. But I'll check where else it's used and if we use it in enough places, that might make me change my mind.

lib/ecmascript.ts
ptomato: I agree. I thought it'd be better to accept a JSBI since we use JSBI inside the `round` method, but on reflection accepting a Number and converting will encapsulate the bigint usage inside TimeDuration, which is good.

lib/instant.ts
ptomato: Probably a mistake. Since this line is already a diff due to JSBI usage, I'll add another commit to remove it.

lib/math.ts
ptomato: Probably pretty rare. It's called in Duration.p.round and since/until methods of types with times, only when rebalancing a duration to a largestUnit of ms, µs, or ns.

FWIW, while it is certainly slower than normal Number math, and likely slower than native BigInt math, it is actually faster than emulating BigInt math with the big-integer library in the proposal polyfill. I haven't checked whether that's true for JSBI.

I would prefer not to try to optimize it until we've caught up with the current state of the proposal. However, when we do decide to optimize it, we could make a fast path where Number arithmetic is used if the quantities are under certain limits where it's impossible that the multiplication hits the safe integer limit.

Alternatively, I also have an implementation of this math operation in a gist, from back when I was writing this code, that only uses Numbers, no bigint or string manipulation, which I cribbed from a C++ numerics library: https://gist.github.com/ptomato/d1ac949bbd3e5719766e770b7c0415bb
It's pretty impenetrable but I'd expect it to be really fast and it'd certainly be an option to use in the future if this was showing up in profiles.

lib/zoneddatetime.ts
ptomato: :shrug: ... I think having BigIntFloorDiv in bigintmath.ts is DRY enough. We only use it a few times anyway, and only some of those want a Number result.

ptomato:
Thanks for the review. The comments were food for thought, much appreciated!
I fixed up some things and will merge this now.

justingrant:
Cool, sounds good. Thanks for churning through these changes!

PR #302, opened 2025-03-03
Rebase up to May 2024

ptomato:
Another batch of 20 mostly small commits. There are a few questionable things inline that I will flag with additional comments.


lib/slots.ts
```diff
@@ -313,3 +362,19 @@ export function SetSlot<KeyT extends SlotKey>(
...
+  const slots = GetSlots(container);
+
+  if (slots === undefined) throw new TypeError('Missing slots for the given container');
+
+  const existingSlot = slots[id];
+
+  if (existingSlot === undefined) throw new TypeError(`tried to reset ${id} which was not set`);
+
+  slots[id] = value;
+}
```
ptomato, line 380: This is new because we made SetSlot in this polyfill work only once per object per slot, which makes sense because Temporal objects are immutable. But the Intl.DateTimeFormat monkeypatch uses lazily computed slots, because of the slow creation time of Intl.DateTimeFormat objects, which doesn't fit with that. So I added ResetSlot and intend to only use it in getSlotLazy.

lib/timeduration.ts
```diff
@@ -122,8 +123,9 @@ export class TimeDuration {
     return JSBI.equal(this.totalNs, ZERO);
   }
 
-  round(increment: number, mode: Temporal.RoundingMode) {
-    if (increment === 1) return this;
+  round(incrementParam: JSBI | bigint, mode: Temporal.RoundingMode) {
+    const increment = ensureJSBI(incrementParam);
+    if (JSBI.equal(increment, ONE)) return this;
```
ptomato: Unfortunately I had to change the type of this parameter back to JSBI, because of a test262 test where a custom time zone reported a very long day length. I should be able to change this back to Number when custom time zones are removed.

lib/ecmascript.ts
```diff
@@ -384,22 +384,25 @@ const BUILTIN_DEFAULTS = new Map([
...
+  ['milliseconds', 'millisecond', 'time', 1e6],
+  ['microseconds', 'microsecond', 'time', 1e3],
+  ['nanoseconds', 'nanosecond', 'time', 1]
 ] as const;
 const SINGULAR_FOR = new Map(TEMPORAL_UNITS.map((e) => [e[0], e[1]] as const));
 const PLURAL_FOR = new Map(TEMPORAL_UNITS.map(([p, s]) => [s, p]));
 const UNITS_DESCENDING = TEMPORAL_UNITS.map(([, s]) => s);
+const NS_PER_TIME_UNIT = new Map(
+  TEMPORAL_UNITS.map(([, s, , l]) => [s, l] as const).filter(([, l]) => l)
+) as ReadonlyMap<Temporal.TimeUnit | 'day', number>;
```
ptomato, line 409: This is pretty annoying that we have to do `castExists` on every value from this Map. I'm not sure if there's a way to tell TypeScript that there's a value for every possible key allowed by the type. Happy to update it if there is a good way to express that, otherwise I guess we should leave it for later.

lib/intrinsicclass.ts
```diff
@@ -6,10 +6,19 @@ import { DEBUG } from './debug';
...
+  ['Temporal.Instant']: OmitConstructor<Temporal.Instant> &
     (new (epochNanoseconds: JSBI) => Temporal.Instant) & { prototype: typeof Temporal.Instant.prototype };
-  ZonedDateTime: OmitConstructor<Temporal.ZonedDateTime> &
+  ['Temporal.PlainDate']: typeof Temporal.PlainDate;
+  ['Temporal.PlainDateTime']: typeof Temporal.PlainDateTime;
+  ['Temporal.PlainMonthDay']: typeof Temporal.PlainMonthDay;
+  ['Temporal.PlainTime']: typeof Temporal.PlainTime;
+  ['Temporal.PlainYearMonth']: typeof Temporal.PlainYearMonth;
+  ['Temporal.TimeZone']: typeof Temporal.TimeZone;
+  ['Temporal.ZonedDateTime']: OmitConstructor<Temporal.ZonedDateTime> &
```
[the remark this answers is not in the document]
justingrant: Yep, makes sense. I remember having to do something similar before with Instant.

That said, why is the argument type `JSBI` instead of `JSBI | bigint`?  If it's a public API then won't users be calling it with either type?

lib/ecmascript.ts
justingrant: Your problem here is that `get` is typed so that it could possibly return undefined. 

You could work around this by replacing `ReadonlyMap` with your own utility type that re-types the `get` method. 

```typescript
// Utility type whose `get` is only callable using valid keys, and that therefore
// omits `undefined` from its result.
type ConstMap<K, V> = Omit<ReadonlyMap<K, V>, 'get'> & { get(key: K): V };
const NS_PER_TIME_UNIT = new Map(
  TEMPORAL_UNITS.map(([, s, , l]) => [s, l] as const).filter(([, l]) => l)
) as ConstMap<Temporal.TimeUnit | 'day', number>;
```

Another alternative would be to forgo `Map` and just use an object literal declared with `as const`. 

There's also probably a way to type a non-`Map` object even without declaring an `as const` literal, and casting it to `Record<Temporal.TimeUnit | 'day', number>` or something like that.

If upstream is using `Map` (unnecessarily?) then the first solution is probably the cleanest.

lib/bigintmath.ts
```diff
@@ -11,14 +11,18 @@ export const MILLION = JSBI.BigInt(1e6);
 export const BILLION = JSBI.BigInt(1e9);
 const HOUR_SECONDS = 3600;
 export const HOUR_NANOS = JSBI.multiply(JSBI.BigInt(HOUR_SECONDS), BILLION);
-export const MINUTE_NANOS = JSBI.multiply(SIXTY, BILLION);
+export const MINUTE_NANOS_JSBI = JSBI.multiply(SIXTY, BILLION);
```
justingrant, line 14: One minor suggestion: should we call these constants `*_BIG` instead of `*_JSBI`?  Then when we finally rip JSBI out and replace it with normal `bigint` then we won't have to update all uses of these constants?

Also fine to leave as-is.

lib/ecmascript.ts
```diff
@@ -4526,7 +4529,7 @@ function DifferenceInstant(
   ns1: JSBI,
   ns2: JSBI,
   increment: number,
-  smallestUnit: keyof typeof nsPerTimeUnit,
+  smallestUnit: Parameters<typeof NS_PER_TIME_UNIT.get>[0],
```
justingrant, line 4532: Minor thing, but IMO it'd make this more readable to use a custom type alias here, e.g. `TIME_OR_DAY_UNIT` or something like that.

lib/ecmascript.ts
```diff
@@ -5936,115 +5934,56 @@ export function AddDurationToOrSubtractDurationFromZonedDateTime(
...
+  let r1: JSBI, r2: JSBI;
+  if (JSBI.lessThan(quantity, ZERO)) {
+    r1 = JSBI.subtract(quotient, ONE);
+    r2 = quotient;
+  } else {
+    r1 = quotient;
+    r2 = JSBI.add(quotient, ONE);
   }
-  return quotient * increment;
+  const cmp = compare(abs(JSBI.multiply(remainder, TWO)), increment) * (JSBI.lessThan(quantity, ZERO) ? -1 : 1);
```
justingrant, line 5971: It'd be nice to have a comment on this line explaining what this line's magic is doing.

lib/ecmascript.ts
```diff
@@ -5936,115 +5934,56 @@ export function AddDurationToOrSubtractDurationFromZonedDateTime(
...
+  const even = isEven(r1);
```
justingrant, line 5972: Very minor, but seems like the call to `isEven` could be inlined in the call to `ApplyUnsignedRoundingMode` to avoid paying for it when we don't need to.

lib/ecmascript.ts
```diff
@@ -5936,115 +5934,56 @@ export function AddDurationToOrSubtractDurationFromZonedDateTime(
...
+  const rounded = JSBI.equal(JSBI.multiply(quotient, increment), quantity)
```
justingrant, line 5973: Why is this test not `JSBI.equal(remainder, ZERO)` ?

lib/ecmascript.ts
```diff
@@ -5936,115 +5934,56 @@ export function AddDurationToOrSubtractDurationFromZonedDateTime(
...
+    : ApplyUnsignedRoundingMode(r1, r2, cmp, even, unsignedRoundingMode);
+  return JSBI.multiply(rounded, increment);
 }
 
-export function RoundInstant(
+export function RoundTemporalInstant(
   epochNs: JSBI,
   increment: number,
-  unit: keyof typeof nsPerTimeUnit,
+  unit: Parameters<typeof NS_PER_TIME_UNIT.get>[0],
```
justingrant, line 5982: Same comment as above: replace with a type alias?

lib/ecmascript.ts
```diff
@@ -6086,34 +6023,32 @@ export function RoundTime(
   microsecond: number,
   nanosecond: number,
   increment: number,
-  unit: keyof typeof nsPerTimeUnit | 'day',
-  roundingMode: Temporal.RoundingMode,
-  dayLengthNs = 86400e9
+  unit: Parameters<typeof NS_PER_TIME_UNIT.get>[0],
```
justingrant, line 6026: Same

lib/ecmascript.ts
```diff
@@ -6700,10 +6605,16 @@ export function ASCIILowercase<T extends string>(str: T): T {
   // values. For example, Turkish's "I" character was the source of a security
   // issue involving "file://" URLs. See
   // https://haacked.com/archive/2012/07/05/turkish-i-problem-and-why-you-should-care.aspx/.
-  return str.replace(/[A-Z]/g, (l) => {
-    const code = l.charCodeAt(0);
-    return String.fromCharCode(code + 0x20);
-  }) as T;
+  let lowercase = '';
```
justingrant, line 6608: What's the difference in behavior between `String.prototype.lowercase` and this code?  I'm wondering if it's really needed for a polyfill.

lib/math.ts
```diff
@@ -55,3 +58,50 @@ export function FMAPowerOf10(xParam: number, p: number, zParam: number) {
...
   return sign * NumberParseInt(resStr, 10);
 }
+
+type UnsignedRoundingMode = 'half-even' | 'half-infinity' | 'half-zero' | 'infinity' | 'zero';
+
+export function GetUnsignedRoundingMode(
+  mode: Temporal.RoundingMode,
+  sign: 'positive' | 'negative'
+): UnsignedRoundingMode {
+  const index = +(sign === 'negative'); // 0 = positive, 1 = negative
```
justingrant, line 68:
```suggestion
  const index = +(sign === 'negative') as 0 | 1; // 0 = positive, 1 = negative
```

TS doesn't know what you do! :-)

lib/math.ts
```diff
@@ -55,3 +58,50 @@ export function FMAPowerOf10(xParam: number, p: number, zParam: number) {
...
+  switch (mode) {
+    case 'ceil':
+      return (['infinity', 'zero'] as const)[index];
```
justingrant, line 71: I don't love this complexity here. Seems like a conditional would be clearer, e.g. `return isNegative ? 'infinity' : 'zero'`.  If we're keeping it this way to align with upstream, then OK, but if not then simpler would be nicer!

justingrant (approved):
Nice! A bunch of minor comments, but nothing required.

lib/bigintmath.ts
ptomato: I'm hoping that most of these will still go away (or at least not need to be exported) when we're more up to date, so I'll leave it as is for now.

lib/math.ts
ptomato: Makes sense. This is how upstream does it but I can't remember why I did it that way; probably to make it work more like looking it up in a table, which is how it's expressed in the spec. We already have to deviate from upstream to add `as const` to make it compile, so may as well use the conditional instead.

lib/ecmascript.ts
ptomato: `toLowerCase` uses [Unicode Default case mapping](https://www.unicode.org/versions/Unicode16.0.0/core-spec/chapter-3/#G33992) and on a casual reading I cannot say with any certainty what that algorithm does. _Probably_ it produces identical results if the input is guaranteed to be ASCII, so we could get rid of this, but let's save that for when we detach from usptream.

lib/ecmascript.ts
ptomato: Nice find.

lib/intrinsicclass.ts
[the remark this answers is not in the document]
ptomato: I'm not sure why that is; it was already that way so I just left it alone. I think there is probably some weirdness with what types are exposed publicly or privately. The public type definitions file says `bigint`, `zoneddatetime.ts` says `JSBI | bigint`, and the intrinsics file says `JSBI`. Maybe because when you get the intrinsic %Temporal.ZonedDateTime%, it's internal, so you're only going to call it with JSBI.

ptomato:
Thanks for the reviews! The comments have been very helpful.

lib/math.ts
justingrant: Cool. I guess the other option would be to make it completely table-driven, like an object where the keys are the rounding mode and the values are either a single string or a tuple. Then we could forego the switch statement completely.

justingrant:
Lol, I'm doing the easy part. Thanks so much for rocketing through catching up with upstream. Amazing work.

PR #303, opened 2025-03-04
Rebase through mid-May 2024

ptomato:
This is only a small number of commits, but a large diff to go through. It includes the big refactor of how durations are rounded relative to ZonedDateTime.


justingrant (approved):
Great!

ptomato:
Thanks for the quick review!

PR #305, opened 2025-03-05
Rebase through end of June 2024

ptomato:
Another stack of 25 commits. Most of them are either the more straightforward removals from the June 2024 changes, or improvements to the ISO 8601 grammar which were reflected in test/validStrings.mjs. (We have not yet gotten to the removal of calendars and time zones.)


justingrant (approved):
Love it!

👍  image

PR #307, opened 2025-03-12
Rebase: Removal of time zone and calendar objects

ptomato:
This is the big one, removing Temporal.Calendar and Temporal.TimeZone objects.

I guess I'd suggest particularly looking at `lib/calendar.ts` in the review. A lot of the code in the other `lib/` files is mechanical removal of user calls and method records.


index.d.ts
```diff
@@ -1085,57 +891,9 @@ export namespace Temporal {
...
-      options?: ToInstantOptions
-    ): Temporal.Instant;
-    getPossibleInstantsFor(dateTime: Temporal.PlainDateTime | PlainDateTimeLike | string): Temporal.Instant[];
-    toString?(): string;
-    toJSON?(): string;
-  }
-
-  /**
-   * Any of these types can be passed to Temporal methods instead of a Temporal.TimeZone.
+   * Any of these types can be passed to Temporal methods instead of a time zone ID.
```
justingrant, line 894:
```suggestion
   * To provide a time zone to any non-constructor Temporal method, any of the following are allowed:
   * - An IANA time zone ID string like "Europe/Paris"
   * - An RFC 9557 string that includes a time zone annotation like "2025-03-13T05:40:57.499+01:00[Europe/Paris]"
   * - a `Temporal.ZonedDateTime` object
```

justingrant (approved):
Boom!  Was a long road to get here....

ptomato:
Thanks for the review!

PR #309, opened 2025-03-15
Rebase through mid-September 2024

ptomato:
For once, a batch with no upstream commits from me!

This PR does bring in the pattern of capturing intrinsics and calling them directly, instead of using things like Array methods. I understand that's not a goal of this polyfill, but I'd suggest that we adopt this coding pattern temporarily to speed up the upcoming rebases, then move back to directly invoking methods when we're done with the rebases.


justingrant (approved):
Looks good. If you're confident that merging this (which we'll need to revert most of it later) is the best way to get to the end state we want, then go for it. Otherwise we could consider merging part of it?

ptomato:
I am pretty sure that going with the flow for now is less work overall.

PR #310, opened 2025-03-17
Rebase: More from September 2024

ptomato:
A batch of commits increasing conciseness using Duration Records.


justingrant (approved):
Looks good. Thanks!

PR #312, opened 2025-03-17
Rebase through early October 2024

ptomato:
Includes a refactor of calendar operations.

### DIFF
--- a/lib/ecmascript.ts
+++ b/lib/ecmascript.ts
@@ -917,21 +917,6 @@ export function RegulateTime(
   return { hour, minute, second, millisecond, microsecond, nanosecond };
 }
 
-export function RegulateISOYearMonth(yearParam: number, monthParam: number, overflow: Overflow) {
-  let year = yearParam;
-  let month = monthParam;
-  const referenceISODay = 1;
-  switch (overflow) {
-    case 'reject':
-      RejectISODate(year, month, referenceISODay);
-      break;
-    case 'constrain':
-      ({ year, month } = ConstrainISODate(year, month));
-      break;
-  }
-  return { year, month };
-}
-
 export function ToTemporalPartialDurationRecord(temporalDurationLike: Temporal.DurationLike | string) {
   if (!IsObject(temporalDurationLike)) {
     throw new TypeErrorCtor('invalid duration-like');
--- a/lib/ecmascript.ts
+++ b/lib/ecmascript.ts
@@ -1347,6 +1347,10 @@ export function TemporalObjectToFields(
   return ISODateToFields(calendar, isoDate, type);
 }
 
+function calendarImplForID(calendar: BuiltinCalendarId) {
+  return GetIntrinsic('%calendarImpl%')(calendar);
+}
+
 export function calendarImplForObj(
   temporalObj:
     | Temporal.PlainDate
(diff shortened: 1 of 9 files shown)


lib/ecmascript.ts
```diff
@@ -1861,6 +1886,43 @@ export function InterpretISODateTimeOffset(
     return GetEpochNanosecondsFor(timeZone, dt, disambiguation);
   }
 
+  // The caller wants the offset to always win ('use') OR the caller is OK
+  // with the offset winning ('prefer' or 'reject') as long as it's valid
+  // for this timezone and date/time.
+  if (offsetBehaviour === 'exact' || offsetOpt === 'use') {
```
justingrant, line 1892: I forget: did we agree to standardize on British or American spellings in our code?  I have no preference, other than we should match one standard throughout if possible.

lib/ecmascript.ts
```diff
@@ -3391,8 +3421,8 @@ export function UnbalanceDateDurationRelative(dateDuration: DateDuration, plainR
   // balance years, months, and weeks down to days
   const isoDate = TemporalObjectToISODateRecord(plainRelativeTo);
   const later = CalendarDateAdd(GetSlot(plainRelativeTo, CALENDAR), isoDate, yearsMonthsWeeksDuration, 'constrain');
-  const epochDaysEarlier = ISODateToEpochDays(isoDate.year, isoDate.month, isoDate.day);
-  const epochDaysLater = ISODateToEpochDays(later.year, later.month, later.day);
+  const epochDaysEarlier = ISODateToEpochDays(isoDate.year, isoDate.month - 1, isoDate.day);
```
justingrant, line 3424: I don't know why these -1s are here, so I'll assume they're correct!

lib/plaindatetime.ts
```diff
@@ -110,57 +110,57 @@ export class PlainDateTime implements Temporal.PlainDateTime {
...
     if (!ES.IsTemporalDateTime(this)) throw new TypeErrorCtor('invalid receiver');
     const isoDate = ES.TemporalObjectToISODateRecord(this);
-    return ES.CalendarMonthsInYear(GetSlot(this, CALENDAR), isoDate);
+    return ES.calendarImplForObj(this).isoToDate(isoDate, { monthsInYear: true }).monthsInYear;
   }
   get inLeapYear(): Return['inLeapYear'] {
     if (!ES.IsTemporalDateTime(this)) throw new TypeErrorCtor('invalid receiver');
     const isoDate = ES.TemporalObjectToISODateRecord(this);
-    return ES.CalendarInLeapYear(GetSlot(this, CALENDAR), isoDate);
+    return ES.calendarImplForObj(this).isoToDate(isoDate, { inLeapYear: true }).inLeapYear;
```
justingrant, line 163: There's a lot of repetitive code here. Could you explain a bit about this pattern?  If only one scalar is needed, I wonder if there's a simpler way to get it than this `{ fieldName: true }` parameter followed by a property getter. Also the same pattern is repeated in every method, could it be DRY-ed? Unless I'm missing something, could all these 2-line repeated calls be coalesced into a single shared function with a signature like this?

```js
ES.getCalendarProperty(this, 'inLeapYear')
```

justingrant (approved):
Overall looks good. Nice work with the compex TS. 

I had a few optional suggestions and questions.

lib/ecmascript.ts
ptomato: TC39 uses British spelling in the specifications, so that's what I've been using.

lib/ecmascript.ts
ptomato: Just to match the spec text: https://tc39.es/proposal-temporal/#sec-isodatetoepochdays

lib/plaindatetime.ts
ptomato: There are a couple of other places where we fetch multiple scalars at once (e.g. `ISODateToFields`, `calendarDateWeekOfYear`.) But I like the suggestion to reduce the repetition in the many public property getters. I've added it to my to-do list of cleanups after the rebase is done.

ptomato:
Thanks for the review!

PR #313, opened 2025-03-18
Two fixes that are not rebases

ptomato:
One fix is for a merge conflict resolution that I messed up. The other is a solution for a weird error that I got while rebasing, where `git rebase` could not find my `.gitconfig` file.


justingrant (approved):
👍

PR #314, opened 2025-03-19
Remove usage of private fields

ptomato:
Downleveling this with the TypeScript compiler would add a runtime dependency on tslib. The TimeDuration object is private anyway and not exported to client code, so we don't strictly need this method to be private.


ptomato:
Looks like CI's node 20.x got upgraded from 20.18 to 20.19, that brought in a new CLDR and broke a test. https://github.com/js-temporal/temporal-polyfill/pull/315 adds a new expected failures file until we can update to a version of test262 that takes CLDR 46 into account.

justingrant (approved):
👍

PR #315, opened 2025-03-19
Add expected failures for CLDR >= 46

ptomato:
ICU 76.1 / CLDR 46 got backported to Node 20.19 LTS, which is now in the GitHub CI. We need a CLDR 46 expected failures file until we can update to a version of test262 where this failure has been fixed.


justingrant (approved):
👍

PR #316, opened 2025-03-20
Rebase: ISO Date-Time Record refactoring

ptomato:
Here is a batch of commits from October 2024 all dealing with refactoring the way ISO Date-Time Records are used in the spec. It made the spec text more concise, so should do the same with this JS code. (Net -1k lines)


justingrant (approved):
Nice!

PR #317, opened 2025-03-22
Rebase up to December 2024

ptomato:
Second-to-last batch of rebases.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -494,6 +494,9 @@ export namespace Temporal {
    */
   export type TransitionDirection = 'next' | 'previous' | { direction: 'next' | 'previous' };
 
+  // Handle Intl.LocalesArgument regardless of whether the TS version has it yet
+  type LocalesArgument = ConstructorParameters<typeof Intl.DateTimeFormat>[0];
+
   export type DurationLike = {
     years?: number;
     months?: number;
@@ -552,7 +555,7 @@ export namespace Temporal {
     subtract(other: Temporal.Duration | DurationLike | string): Temporal.Duration;
     round(roundTo: DurationRoundTo): Temporal.Duration;
     total(totalOf: DurationTotalOf): number;
-    toLocaleString(locales?: string | string[], options?: globalThis.Intl.DateTimeFormatOptions): string;
+    toLocaleString(locales?: LocalesArgument, options?: globalThis.Intl.DateTimeFormatOptions): string;
     toJSON(): string;
     toString(options?: ToStringPrecisionOptions): string;
     valueOf(): never;
@@ -600,7 +603,7 @@ export namespace Temporal {
       roundTo: RoundTo<'hour' | 'minute' | 'second' | 'millisecond' | 'microsecond' | 'nanosecond'>
     ): Temporal.Instant;
     toZonedDateTimeISO(tzLike: TimeZoneLike): Temporal.ZonedDateTime;
-    toLocaleString(locales?: string | string[], options?: globalThis.Intl.DateTimeFormatOptions): string;
+    toLocaleString(locales?: LocalesArgument, options?: globalThis.Intl.DateTimeFormatOptions): string;
     toJSON(): string;
     toString(options?: InstantToStringOptions): string;
     valueOf(): never;
@@ -678,7 +681,7 @@ export namespace Temporal {
     ): Temporal.ZonedDateTime;
     toPlainYearMonth(): Temporal.PlainYearMonth;
     toPlainMonthDay(): Temporal.PlainMonthDay;
-    toLocaleString(locales?: string | string[], options?: globalThis.Intl.DateTimeFormatOptions): string;
+    toLocaleString(locales?: LocalesArgument, options?: globalThis.Intl.DateTimeFormatOptions): string;
     toJSON(): string;
     toString(options?: ShowCalendarOption): string;
     valueOf(): never;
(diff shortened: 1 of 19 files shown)


lib/intl.ts
```diff
@@ -71,7 +72,7 @@ function getSlotLazy(obj: DateTimeFormatImpl, slot: LazySlot) {
 
 function createDateTimeFormat(
   dtf: DateTimeFormatImpl,
-  locale: Params['constructor'][0],
+  locale: Intl.LocalesArgument,
   optionsParam: Params['constructor'][1]
 ) {
   const hasOptions = typeof optionsParam !== 'undefined';
```
justingrant, line 78: Should the type of `options` one line below here be `typeof optionsParam` ?

lib/intl.ts
```diff
@@ -114,7 +115,7 @@ function createDateTimeFormat(
   } else {
     options = ObjectCreate(null);
   }
-  const original = new IntlDateTimeFormat(locale, options);
+  const original = new IntlDateTimeFormat(locale as Params['constructor'][0], options);
```
justingrant, line 118: Why is this cast necessary?  Is the type of `constructor` different from the type of `Intl.DateTimeFormat`'s constructor?

justingrant (approved):
👍  🙏

index.d.ts
```diff
@@ -600,7 +600,7 @@ export namespace Temporal {
       roundTo: RoundTo<'hour' | 'minute' | 'second' | 'millisecond' | 'microsecond' | 'nanosecond'>
     ): Temporal.Instant;
     toZonedDateTimeISO(tzLike: TimeZoneLike): Temporal.ZonedDateTime;
-    toLocaleString(locales?: string | string[], options?: globalThis.Intl.DateTimeFormatOptions): string;
+    toLocaleString(locales?: globalThis.Intl.LocalesArgument, options?: globalThis.Intl.DateTimeFormatOptions): string;
```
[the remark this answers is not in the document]
ptomato: It's a port of https://github.com/tc39/proposal-temporal/commit/12012602de791e1399a67e9361a90cd3f0de0d44 so I guess we could just drop this commit?

lib/intl.ts
ptomato: Without this, we get:

> Argument of type 'LocalesArgument' is not assignable to parameter of type 'string | string[] | undefined'.

I guess because we are indeed currently on a version of TypeScript that doesn't have LocalesArgument. But if we drop the LocalesArgument commit, this will go away.

lib/ecmascript.ts
```diff
@@ -333,11 +334,26 @@ export function RequireString(value: unknown): string {
...
+  if (
+    value.length < 3 ||
+    value.length > 4 ||
+    value[0] !== 'M' ||
+    Call(StringPrototypeIndexOf, '0123456789', [value[1]]) === -1 ||
+    Call(StringPrototypeIndexOf, '0123456789', [value[2]]) === -1 ||
+    (value[1] + value[2] === '00' && value[3] !== 'L') ||
+    (value[3] !== 'L' && value[3] !== undefined)
+  ) {
+    throw new RangeError(`bad month code ${value}; must match M01-M99 or M00L-M99L`);
```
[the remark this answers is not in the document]
ptomato: I think it was Richard :smile:

index.d.ts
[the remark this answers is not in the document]
justingrant: It looks like this is ES2020 which is probably long enough ago that it's OK. If enough users of this library run into a problem, then we could reconsider.

lib/intl.ts
justingrant: I dug in further and it looks like a problem is that TS's own types in 5.1.3 (the version in this repo) are out-of-date relative to the latest TS lib types.  5.1.3 was released on June 1st, 2023, but the PR https://github.com/microsoft/TypeScript/pull/52996 to update lib types to include `LocalesArgument` wasn't merged until TS 5.4.

We could update the TS version in this repo, but I think a cleaner change could be to avoid using `Intl.LocalesArgument` in the implementation of the polyfill. Instead, we can just use `Params['constructor'][0]` which seems like it should work for any version of TS because we're just referring back to whatever declaration is used for the `Intl.DateTimeFormat` constructor.

This might cause our usage of `Intl.LocalesArgument` in the polyfill's .d.ts to cause an error, and if so then we could I guess apply the same trick in our .d.ts: refer back to whatever parameter types that the `Intl.DateTimeFormat` constructor is using.

lib/duration.ts
```diff
@@ -281,9 +282,20 @@ export class Duration implements Temporal.Duration {
       throw new RangeErrorCtor(`a starting point is required for ${largestUnit}s balancing`);
     }
     assert(!ES.IsCalendarUnit(smallestUnit), 'smallestUnit was larger than largestUnit');
-    let duration = ES.NormalizeDurationWith24HourDays(this);
-    duration = ES.RoundTimeDuration(duration, roundingIncrement, smallestUnit, roundingMode);
-    return ES.UnnormalizeDuration(duration, largestUnit);
+    let internalDuration = ES.ToInternalDurationRecordWith24HourDays(this);
+    if (smallestUnit === 'day') {
+      // First convert time units up to days
+      const DAY_NANOS = 86400 * 1e9;
```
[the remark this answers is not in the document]
ptomato: Not exported, but I went ahead and fixed that; it's very minor even though it's a departure from upstream.

lib/intl.ts
ptomato: This is good advice, thanks. In the end I added
```ts
type LocalesArgument = ConstructorParameters<typeof Intl.DateTimeFormat>[0];
```
to the index.d.ts. This solution is much cleaner.

lib/intl.ts
ptomato: As far as I can tell it should not, because that could be undefined.

lib/ecmascript.ts
```diff
@@ -3769,31 +3795,29 @@ function NudgeToZonedTime(
 
   // Compute time parts of the duration to nanoseconds and round
   // Result could be negative
-  let roundedNorm = duration.norm.round(
-    JSBI.BigInt(Call(MapPrototypeGet, NS_PER_TIME_UNIT, [unit]) * increment),
-    roundingMode
-  );
+  const unitIncrement = JSBI.BigInt(Call(MapPrototypeGet, NS_PER_TIME_UNIT, [unit]) * increment);
```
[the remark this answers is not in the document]
ptomato: Yep, will add this to my list of minor changes post-rebasing.

PR #319, opened 2025-03-26
Various post-rebase improvements

ptomato:
Here is a PR with a number of improvements based on review comments during the rebase.


lib/ecmascript.ts
```diff
@@ -1184,18 +1185,18 @@ export function GetTemporalUnitValuedOption<
...
+    const plural = PLURAL_FOR[singular];
     if (plural !== undefined) Call(ArrayPrototypePush, allowedValues, [plural]);
   }
   let retval = GetOption(options, key, allowedValues, defaultVal);
   if (retval === undefined && requiredOrDefault === REQUIRED) {
     throw new RangeErrorCtor(`${key} is required`);
   }
   // Coerce any plural units into their singular form
-  if (Call(MapPrototypeHas, SINGULAR_FOR, [retval as Temporal.PluralUnit<Temporal.DateTimeUnit>])) {
+  if ((retval as Temporal.PluralUnit<Temporal.DateTimeUnit>) in SINGULAR_FOR) {
```
justingrant, line 1103: `in` is a type guard, so you can avoid casts.  Like this: 

```ts
  return (retval && retval in SINGULAR_FOR ? SINGULAR_FOR[retval] : retval) as R;
```

lib/plaindate.ts
```diff
@@ -124,63 +122,63 @@ export class PlainDate implements Temporal.PlainDate {
...
   equals(otherParam: Params['equals'][0]): Return['equals'] {
-    if (!ES.IsTemporalDate(this)) throw new TypeErrorCtor('invalid receiver');
+    if (!ES.IsTemporalDate(this)) throw new TypeError('invalid receiver');
     const other = ES.ToTemporalDate(otherParam);
     if (ES.CompareISODate(GetSlot(this, ISO_DATE), GetSlot(other, ISO_DATE)) !== 0) return false;
     return ES.CalendarEquals(GetSlot(this, CALENDAR), GetSlot(other, CALENDAR));
   }
   toString(options: Params['toString'][0] = undefined): string {
-    if (!ES.IsTemporalDate(this)) throw new TypeErrorCtor('invalid receiver');
+    if (!ES.IsTemporalDate(this)) throw new TypeError('invalid receiver');
```
justingrant, line 125: We duplicate this line 100+ times, with only the test varying. Should we have `ES.RejectReceiver(this, ES.IsTemporalDate)` ? 

This code is highly compressible so I'm not sure how much bundle size win that'd be, but it can't hurt and it'd probably make these files easier to read.

lib/assert.ts
```diff
@@ -1,10 +1,8 @@
-import { Error as ErrorCtor } from './primordials';
-
 export function assert(condition: boolean, message: string): asserts condition {
-  if (!condition) throw new ErrorCtor(`assertion failure: ${message}`);
+  if (!condition) throw new Error(`assertion failure: ${message}`);
```
justingrant, line 2: OMG the changes in this commit make the library so much easier to read! Thanks for doing this.

justingrant (approved):
Love to see it. Really nice savings in both lines (and hopefully some bundle size) and complexity.

lib/plaindate.ts
ptomato: Surprisingly large win: -4k minified.

justingrant: Cool. I suspect the diff in compressed & minified won't be nearly as large, but any improvement is good!

PR #324, opened 2025-03-28
Delete assertions using rollup strip plugin

ptomato:
Produces a bundle size savings of about -3k minified, -1k minified and gzipped.

(I unfortunately didn't think of this before preparing the release, otherwise I'd have tried to include it.)

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -2,6 +2,7 @@ import commonjs from '@rollup/plugin-commonjs';
 import nodeResolve from '@rollup/plugin-node-resolve';
 import babel from '@rollup/plugin-babel';
 import replace from '@rollup/plugin-replace';
+import strip from '@rollup/plugin-strip';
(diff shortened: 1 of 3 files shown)


justingrant (approved):
👍

rollup.config.js
```diff
@@ -44,6 +45,13 @@ function withPlugins(
     }
     basePlugins.push(babel(options.babelConfig));
   }
+  if (!options.enableAssertions) {
+    basePlugins.push(
+      strip({
+        functions: ['assert', 'assertNotReached', 'assertExists', 'ES.assertExists']
```
justingrant, line 51: Do we need to do anything to remove the emission of the source code for these functions from prod builds? Or is this automatic via this plugin?

ptomato: We pass `enableAssertions: !isProduction` later in the file. So it's not automatic, but already taken care of.

PR #327, opened 2025-03-28
Minify function names in prod builds

justingrant:
Changex rollup's `keep_fnames` option from `true` to `false`.

My understanding is that this option will minify function names, for example the often-long internal exports in `ecmascript.ts`.

This change (and a minor change to run terser's compression 2x to save 85 bytes) reduces minified & gzipped size of `dist/index.cjs` to 36.3K. (a 3.7K reduction). ESM reductions are similar. I didn't apply this optimization to ES5 UMD builds because it caused all ES5 Test262 tests for `Temporal.prototype.*.name` to fail.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@js-temporal/polyfill",
-  "version": "0.4.4",
+  "version": "0.5.0",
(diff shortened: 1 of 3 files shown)


ptomato (approved):
Makes sense.

PR #331, opened 2025-03-30
Support Node 22 & 23

### LINKED ISSUE #323: Run tests with Node 22 and 23
> Node 20 is already old. We should probably test with 22 and 23 too.

justingrant:
Support Node 22 & 23. In addition to changing expected failures files for bugs that Node fixed, there were two Node changes that required polyfill code changes:
* Now that Node 22 supports offset time zones in Intl.DTF, we needed to stop throwing when these IDs were supplied, and we needed to guard against U+2212 (unicode minus) in offset time zone IDs because Node doesn't catch these (yet).
* Node 23 adds `Intl.DurationFormat`, and this PR fixes an infinite recursive loop (accidentally introduced in 19b00b2) that occurred if a native `Intl.DurationFormat` was present.

Fixes #323.

### DIFF
--- a/test/expected-failures.txt
+++ b/test/expected-failures.txt
@@ -6,21 +6,12 @@
 # canonicalizing the timezone name.
 staging/Intl402/Temporal/old/date-time-format.js
 
-# temporal-test262-runner doesn't support the $262 global object
-intl402/DateTimeFormat/proto-from-ctor-realm.js
-
 # These are caught by the default test glob, but are unrelated to Temporal.
 # They rely on Intl.DateTimeFormat supporting offset time zones.
 intl402/DateTimeFormat/prototype/format/offset-timezone-gmt-same.js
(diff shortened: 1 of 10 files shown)


test/expected-failures.txt
```diff
@@ -6,22 +6,13 @@
 # canonicalizing the timezone name.
 staging/Intl402/Temporal/old/date-time-format.js
 
-# temporal-test262-runner doesn't support the $262 global object
-intl402/DateTimeFormat/proto-from-ctor-realm.js
```
ptomato, line 10: Note, I added this in https://github.com/js-temporal/temporal-polyfill/pull/282 in order to be able to support https://github.com/js-temporal/temporal-test262-runner/pull/19, but then we decided to stick to npm releases of temporal-test262-runner only. When we update temporal-test262-runner, we'll need to add this back in.

ptomato (approved):
Much appreciated! I'll cut the new release now.

PR #333, opened 2025-03-31
Temporal is stage 3.

a-eid:


ptomato:
Thanks for taking the time to contribute! I think those sentences are correct as is, though. Temporal is currently stage 3, and the text is describing things that may or may not happen before it gets to stage 4.

PR #339, opened 2025-05-15
Update test262

ptomato:
I removed usage of Object.hasOwn from the era helper, which means these tests should pass now on old Node runtimes that don't have it.


justingrant (approved):
Cool, thanks!

PR #340, opened 2025-05-16
Test Node 24 in CI

justingrant:
Start testing in Node 24


justingrant:
Amazingly, no tests broke from Node 24. There's a first time for everything!  ;-)

ptomato (approved):
Glad it was so simple!

PR #354, opened 2026-01-31
docs(README): Point to MDN for documentation

ferdnyc:
Update documentation references to point to MDN and removed outdated links.

Language blatantly stolen from tc39/proposal-temporal#3086 by @ptomato.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,14 +41,7 @@ If you've found a bug in this polyfill&mdash;meaning that the implementation her
 
 ## Documentation
 
-Reference documentation and examples for the Temporal API can be found [here](https://tc39.es/proposal-temporal/docs/index.html).
-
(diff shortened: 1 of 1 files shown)


justingrant (approved):
Thanks!

ptomato:
I would actually like to keep the link to the cookbook, since that doesn't exist on MDN.

justingrant:
Sorry for not checking more carefully... I assumed we were just mirroring your language from the other repo. Should have checked!

ferdnyc:
> I would actually like to keep the link to the cookbook, since that doesn't exist on MDN.

Mmm, I didn't realize there was significant content there that wasn't migrated to MDN.

Given that's the case, I wonder if MDN should _also_ have a link to the Cookbook? (At least in the [See Also section of the main `Temporal` documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal#see_also).)

PR #357, opened 2026-03-26
fix: add updated era codes to EthiopicHelper

kabaros:
This adds the updated era codes for Ethiopic calendar to reflect the updates in CLDR48 https://cldr.unicode.org/downloads/cldr-48#identifiers

Dealing with Ethiopian calendars started failing in the polyfill in versions of node after 20.19 (and after recent browser updates in Firefox) as ICU updated the era codes: https://unicode-org.atlassian.net/browse/ICU-23167. In the polyfill code, the `ecamscript` module seems to delegate to call the `calendar` module passing the updated era code, but the `calendar` module then can't match the era code failing with `RangeError: Era am (ISO year 2016) was not matched by any era`. 

We observed the failure on CI first as node updated the ICU, but then it started failing on Firefox after an update. It currently works on Chrome stable, but fails with the same error on Canary: https://developers.dhis2.org/demo/?path=/story/calendarinput--ethiopic-with-amharic

### DIFF
--- a/lib/calendar.ts
+++ b/lib/calendar.ts
@@ -1934,8 +1934,13 @@ class CopticHelper extends OrthodoxBaseHelper {
 class EthiopicHelper extends OrthodoxBaseHelper {
   constructor() {
     super('ethiopic', [
-      { code: 'ethioaa', names: ['ethiopic-amete-alem', 'mundi'], isoEpoch: { year: -5492, month: 7, day: 17 } },
-      { code: 'ethiopic', names: ['incar'], isoEpoch: { year: 8, month: 8, day: 27 }, anchorEpoch: { year: 5501 } }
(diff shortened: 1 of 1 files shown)


kabaros:
@ptomato I'd appreciate a review please for this one as it completely broke the Ethiopian calendar for us (or a workaround if there is one!) .. thank you in advance 🙏🏿

fabon-f:
@kabaros 
Sorry to advertise my project here, but FYI I made another polyfill supporting the latest spec including era codes of Ethiopian. You can use it for now:
https://github.com/fabon-f/temporal-polyfill-lite (You need to use `temporal-polyfill-lite/calendars-full` entry point for non-ISO calendars)

ptomato (approved):
Thanks. I intend to get the polyfill updated in the next few days to match the status of the proposal at the time of moving to stage 4, and I'll rebase and merge this afterwards.

PR #358, opened 2026-04-02
Rebase part 1

ptomato:
Part 1 of a rebase to the latest proposal state, in preparation for a release.


lib/ecmascript.ts
```diff
@@ -1046,70 +1034,67 @@ type UnitTypeMapping = {
...
-// This signature of the function is NOT used in type-checking, so restricting
-// the default value via generic binding like the other overloads isn't
-// necessary.
-export function GetTemporalUnitValuedOption<
-  T extends keyof UnitTypeMapping,
-  D extends typeof REQUIRED | UnitTypeMapping[T] | 'auto' | undefined,
-  E extends 'auto' | Temporal.DateTimeUnit,
-  R extends UnitTypeMapping[T] | Exclude<D, typeof REQUIRED> | E
->(
+export function GetTemporalUnitValuedOption(
```
ptomato, line 1027: The typing of GetTemporalUnitValuedOption and ValidateTemporalUnitValue below is probably the most iffy part of this PR, worth looking at carefully.

GetTemporalUnitValuedOption becomes simpler but requires 2 overloads: the 2-argument version can return undefined, while the 3-argument version (3rd argument is `ES.REQUIRED`) can't.

ValidateTemporalUnitValue narrows the type of the unit according to the `unitGroup` argument, and can return undefined if its input type can be undefined. There are 2 extra overloads for the `extraValues` argument as well. This could probably be done with parametrization, given more time, but I couldn't get it to work.

justingrant: I didn't dig in too deeply, but these types look OK to me.

justingrant (approved):
Looks good! Thanks for catching up the polyfill!

lib/ecmascript.ts
ptomato: OK, thanks for checking!

PR #359, opened 2026-04-06
April 2026 rebase, part 2

ptomato:
This batch of commits mostly deals with speeding up calendar calculations.


lib/calendar.ts
```diff
@@ -1100,11 +1137,20 @@ abstract class HelperBase {
...
   ): FullCalendarDate {
+    let years = yearsParam;
+    let months = monthsParam;
     const { year, day, monthCode } = calendarDate;
+    const monthInfo = monthCodeInfo[this.id];
+    const cycleInfo = monthInfo ? monthInfo.cycleInfo : { years: 1, months: 12 };
+    if (cycleInfo && Math.abs(months) > cycleInfo.months) {
+      const cycleCount = Math.trunc(months / cycleInfo.months);
+      years += cycleCount * cycleInfo.years;
+      months %= cycleInfo.months;
```
justingrant, line 1152: TIL that `%=` was a thing.

lib/monthcode.ts
```diff
@@ -1,14 +1,21 @@
...
 export function ParseMonthCode(argument: unknown) {
   const value = ToPrimitiveRequireString(argument);
-  const match = MONTH_CODE_REGEX.exec(value);
-  if (!match) throw new RangeError(`bad month code ${value}; must match M01-M99 or M00L-M99L`);
-  return {
-    monthNumber: +(match[1] ?? match[3] ?? match[5]),
-    isLeapMonth: (match[2] ?? match[4] ?? match[6]) === 'L'
-  };
+  const digits = value.slice(1, 3);
+  const monthNumber = digits.length === 2 ? +digits | 0 : -1; // -1 ensures failure
```
justingrant: `| 0` is a nice trick but it's obscure enough that I'd love to see a comment here explaining what this code is doing.

Also, why is the `| 0` there in the first place?  Specifically why do we want to treat `NaN` as zero but shorter/missing month codes as -1? Shouldn't both be -1?

Honestly it seems like this function could be clearer either by hardcoding the parsing of these two digits or just using a regex. This is just a port from upstream, so it's fine not to make any changes here in this PR, but perhaps we should simplify this code upstream?

justingrant (approved):
Sorry it took so long to review. This looks great!  Thanks for porting these commits. I'll try to jump on the next one sooner.

lib/monthcode.ts
ptomato: This comes from https://github.com/tc39/proposal-temporal/pull/3160#discussion_r2383494083

I can't remember anymore how much of this was performance-sensitive. That said, I think if we just remove the `| 0` the code will still work fine.

I think if we make any further simplifications, we should just do them here. I don't think we should continue working on the upstream polyfill unless it's necessary for test262. I'd like to decouple the two once we finish bringing this one up to speed.

PR #362, opened 2026-05-14
Update description to reflect Stage 4 proposal

jdalton:
This pull request updates the project description in `package.json` to reflect that the Temporal proposal is now at ECMA TC39 Stage 4, instead of Stage 3.

- Updated the `description` field in `package.json` to state that Temporal is now a Stage 4 proposal.


ptomato (approved):
Thanks!

Issue #352, opened 2025-12-29
TypeScript: `ZonedDateTime.from()` doesn't require `timeZone` property, but runtime throws error


MOhhh-ok:
### Current Behavior
When calling `Temporal.ZonedDateTime.from()` with an object that's missing the required `timeZone` property, TypeScript does not report a type error. However, at runtime, the code throws:

```/dev/null/error.txt#L1-1
Uncaught TypeError: required property 'timeZone' missing or undefined
```

### Example Code
```/dev/null/example.tsx#L1-9
import { Temporal } from "@js-temporal/polyfill";

// TypeScript does NOT error on this, but it should
const zdt = Temporal.ZonedDateTime.from({
  year: 2026,
  month: 1,
  day: 1
  // missing timeZone!
});
```

### Expected Behavior
TypeScript should report a type error indicating that `timeZone` is required.

### Root Cause
In `index.d.ts`, the `ZonedDateTimeLike` type defines `timeZone` as optional:

```/dev/null/type-definition.ts#L1-5
export type ZonedDateTimeLike = {
  // ... other properties
  timeZone?: TimeZoneLike;  // ← should NOT be optional
  // ... other properties
};
```

### Suggested Fix
The `timeZone` property should be required in `ZonedDateTimeLike`:

```/dev/null/fix.ts#L1-5
export type ZonedDateTimeLike = {
  // ... other properties
  timeZone: TimeZoneLike;  // ← remove the `?`
  // ... other properties
};
```

### Environment
- `@js-temporal/polyfill` version: 0.5.1
- TypeScript version: 5.9.3

PR #365, opened 2026-09-11
Fix TypeScript types so ZonedDateTime.from() requires timeZone

cpruijsen:
`from`, `compare`, `equals`, `until`, and `since` now require `timeZone` on property bags. Strings and `ZonedDateTime` instances are unchanged.

`Temporal.ZonedDateTime.from({ year, month, day })` type-checked but threw `TypeError: required property 'timeZone' missing or undefined` at runtime (issue #352).

`with()` still takes `ZonedDateTimeLike` (all fields optional). A partial update uses the receiver's time zone; passing `timeZone` throws.

Chose to require `timeZone` on the converter signatures (`ZonedDateTimeLike & { timeZone: TimeZoneLike }`), not on `ZonedDateTimeLike` itself. The patch suggested in #352 makes `timeZone` required on `ZonedDateTimeLike`.

`with({ hour: 12 })` is valid and must not require `timeZone`. TypeScript's own Temporal lib uses the same from/with split.

Can switch to the issue's patch plus a separate `with()` bag type (`Omit` / `Partial`, excluding `timeZone` and `calendar`) if you prefer that shape.

## Test plan

- [x] `test/zoneddatetime-types.ts`: `@ts-expect-error` on bags missing `timeZone` for `from` / `compare` / `equals` / `until` / `since`
- [x] Same file: `from` still accepts strings, instances, and `timeZone: zdt`; `with({ hour: 12 })` still type-checks; `relativeTo` without `timeZone` still type-checks
- [x] Unused `@ts-expect-error` fails if `index.d.ts` is reverted (tsc TS2578)
- [x] `npm run tscheck` (now includes that file)
- [x] `npx tsc --noEmit` (implementation still typechecks against the tighter public signatures)
- [x] `npm run lint` (eslint + prettier + postlint tscheck)


justingrant:
Thanks for the PR. When we next rev this polyfill, we're likely to swap in the official TS types from the Typescript repo rather than continuing to evolve TS types here. If you want to see changes from those types, you may want to file a PR there.

cpruijsen:
Makes sense, thanks for the quick steer. Closing this one so it does not sit in your queue. If I pick it up again I will take it to the TypeScript repo instead.

PR #366, opened 2026-09-11
Fix TypeScript types so with() rejects Temporal objects

### LINKED ISSUE #206: Passing Temporal object to with() should cause compile-time error
Moved from https://github.com/tc39/proposal-temporal/issues/1214

e.g.
```js
const now = Temporal.Now.plainDateTimeISO();
const time = Temporal.PlainTime.from('09:10');
now.with(time);
```
The above code throws a TypeError at runtime because Temporal objects should not be passed to with(). In this case, withPlainTime() should be used instead.

It would be better if the TypeScript binding disallowed this at compile-time, as well.

cpruijsen:
## Summary

`with()` on PlainDate, PlainDateTime, PlainMonthDay, PlainTime, PlainYearMonth, and ZonedDateTime now takes `WithLike<…>`: a property bag without `calendar`/`timeZone`, and not a Temporal instance. Use `withPlainTime()`, `withCalendar()`, or `withTimeZone()` instead.

Runtime already throws for a Temporal object or a `calendar`/`timeZone` field (`RejectTemporalLikeObject`). The TypeScript types accepted those arguments because `*Like` bags are structurally compatible with Temporal instances, and `PlainTime.with` explicitly listed `Temporal.PlainTime`.

`from()` / `equals()` / `until()` / `Duration.with()` are unchanged. Duration.with does not go through `RejectTemporalLikeObject`, and `duration.with(otherDuration)` is valid at runtime.

Fixes https://github.com/js-temporal/temporal-polyfill/issues/206
(Moved from https://github.com/tc39/proposal-temporal/issues/1214.)

## What I chose and the alternative

A `WithLike<T>` mapped type that omits `calendar`/`timeZone` and uses `[Symbol.toStringTag]?: never` so Temporal instances are not assignable. `calendar?: never` / `timeZone?: never` also reject those fields on variables, not only on object literals.

Alternative: TypeScript's `PartialTemporalLike` (omit `calendar`/`timeZone` only). That still accepts `now.with(time)`, which is the example in the issue, because `PlainTime` is structurally a time-field bag and has no `calendar`/`timeZone`. The reporter's example is a Temporal.PlainTime instance, not an object-literal `calendar` field. Can switch to a generic conditional (`T extends TemporalObject ? never : T`) or drop instance rejection for a closer match to the TS 6 lib types.

`WithLike` is exported so the `with()` signatures stay readable. It could be inlined instead.

## Test plan

- [x] `npm run tscheck` (includes `test/with-types.ts`: `@ts-expect-error` on Temporal objects and on `calendar`/`timeZone` bags; valid bags, `withPlainTime`, and `from()` still type-check)
- [x] Revert `index.d.ts` only: the `@ts-expect-error` directives become unused (TS2578); restore: clean
- [x] `tsc --noEmit -p tsconfig.json` (implementation `Params['with']` still matches)
- [x] `npm run lint` (eslint ignore for `test/*.ts` because the fixture is outside `tsconfig.json`'s `lib/*.ts` include)
- [x] Issue example still throws at runtime (types-only change)
- [x] `npm test` (607 passing; Jest only loads `test/*.mjs`)
- [ ] test262 not required (no runtime code touched)


justingrant:
Thanks for the PR. When we next rev this polyfill, we're likely to swap in the official TS types from the Typescript repo rather than continuing to evolve TS types here. If you want to see changes from those types, you may want to file a PR there.

cpruijsen:
Makes sense, thanks for the quick steer. Closing this one so it does not sit in your queue. If I pick it up again I will take it to the TypeScript repo instead.

PR #369, opened 2026-09-18
Polyfill: Fix extreme PlainYearMonth crash in non-ISO calendars

craig-o-curtis:
Ports two tc39/proposal-temporal commits, one commit each:

- a41eb67 "Polyfill: Clamp ISO date to range in iterative calendar-to-ISO conversion" (from tc39/proposal-temporal#3199)
- af0cb4b "Polyfill: Fix extreme PlainYearMonth crash in non-ISO calendars" (tc39/proposal-temporal#3279, which fixes tc39/proposal-temporal#3251)

### What was wrong

To convert a non-ISO calendar date to an ISO date, `calendarToIsoDate()` guesses an ISO date, converts the guess back to the calendar with `Intl.DateTimeFormat`, and corrects it. Near either end of the supported range, a guess can fall outside the range of the JavaScript `Date` object. `Intl.DateTimeFormat` can't format such a date, so the conversion threw `RangeError: Invalid ISO date`, even though the requested date itself is in range. On `main`, all of these throw:

```js
Temporal.PlainDate.from({ calendar: 'hebrew', year: 279517, month: 10, day: 11 }, { overflow: 'reject' });
Temporal.PlainDate.from({ calendar: 'islamic-civil', year: -280804, month: 3, day: 21 }, { overflow: 'reject' });
Temporal.PlainYearMonth.from({ calendar: 'islamic-civil', year: 283583, monthCode: 'M06' });
Temporal.PlainYearMonth.from({ calendar: 'ethioaa', year: 281247, monthCode: 'M06' });
Temporal.PlainYearMonth.from({ calendar: 'indian', year: 275682, monthCode: 'M07' });
```

### What this changes

- **First commit:** the first guess, and the correction after it, are clamped to the supported range (-271821-04-19 to +275760-09-13). This fixes the first two lines above.
- **Second commit:** for the islamic calendars, coptic/ethiopic/ethioaa, and indian, a guess outside the `Date` range is moved into range by whole calendar cycles (islamic 10631 days = 30 years, coptic/ethiopic/ethioaa 1461 days = 4 years, indian 4 years), converted, and the year moved back. This fixes the last three lines. It also fixes islamic-civil `until()` near both ends of the range, which threw on `main`. Dates inside the range take the same code path as before.

### Not fixed by this PR

These behave the same on `main` and on this branch:

[...]

PR #370, opened 2026-09-18
Don't format dates past the range limit when counting months in until()

craig-o-curtis:
Depends on #369.

This is not a port: the fix isn't in tc39/proposal-temporal yet. The same bug is there, and I reported it as tc39/proposal-temporal#3327 with the same change, so a later sync can pick it up.

### What was wrong

In non-ISO calendars, `untilCalendar()` counts months one at a time until it passes the second date. It built each next month with `addMonthsCalendar()`, which converts through an ISO date and formats it with `Intl.DateTimeFormat`. When the second date is in the last month of the supported range (or the first, counting backwards), the next month is outside the range, so the call throws. On `main`:

```js
const d = (s, cal) => Temporal.PlainDate.from(s).withCalendar(cal);
d('+275760-07-01', 'persian').until(d('+275760-09-13', 'persian'), { largestUnit: 'months' });
// RangeError: Invalid ISO date: +275760-10-02T00:00Z (expected P2M12D)
d('-271821-07-01', 'hebrew').until(d('-271821-04-19', 'hebrew'), { largestUnit: 'months' });
// RangeError: Invalid ISO date: -271821-04-04T00:00Z (expected -P2M14D)
Temporal.Duration.from('P40D').round({ largestUnit: 'months', relativeTo: { calendar: 'hebrew', year: 279517, month: 8, day: 15 } });
// RangeError: Invalid ISO date: +275760-09-17T00:00Z (expected P1M10D)
```

The same happens with `largestUnit: 'years'` and with `since()`. The expected values are Chrome 153's native Temporal.

### What this changes

The loop only compares the next month's year and month; its day is always reset to the first date's day. So the next month is now computed arithmetically: add or subtract 1 from the ordinal month, and wrap the year using `monthsInYear()`, which `addMonthsCalendar()` already calls. `addMonthsCalendar(date, ±1)` always lands in the adjacent ordinal month, so ordinary dates give the same results as before (see Testing).

### Why it depends on #369

Some of the dates the loop still converts are near the minimum, and converting them from fields throws on `main` (the bug #369 fixes). The old loop avoided some of these by luck, because the dates it had already formatted were in its cache. So with this PR alone, 32 `until()` calls near the minimum (persian and islamic-\*) that `main` answers now throw. For example, islamic-civil `-271821-04-19` until `-271821-05-19` by months gives `P1M` on `main` and `Invalid ISO date: -271821-01-01T00:00Z` on this branch alone. With #369 underneath, all of them work. The branch is based on `main` and applies cleanly on top of #369 either way.

### Not fixed by this PR

These behave the same with and without this PR:

- **chinese and dangi near either limit:** `until()` throws `TypeError: Internal error. Icu error.` near the max, and `Missing property: year` or `Unexpected leap month suffix` near the min. tc39/proposal-temporal#3277 and #361 cover these.
- **indian before year 1:** throws `calendar 'indian' is broken for ISO dates before 0001-01-01`. #361 fixes it.
- **coptic before the coptic epoch:** reading a date and converting it back from its own fields gives a different date on `main`. #361 fixes it (the coptic `ERA0` change). This PR does change some of those `until()` results (next section).

### Testing

Node 24.21.0 (ICU 78.3): jest 607 passed; eslint, prettier and `tscheck` clean. test262 at the pinned version: intl402/Temporal 330 passed with 8 expected failures, built-ins/Temporal 4330 passed. Both match `main`.

**Edge scan.** For every ISO day within 60 days of each limit, I ran `day.until(limit)` and `limit.until(day)`, by months and by years, and checked each result by adding it back. That's 240 calls per calendar per limit. The table shows how many succeeded:

| Calendar, limit | `main` | this PR | #369 | #369 + this PR |
| --- | --- | --- | --- | --- |
| hebrew, max | 110 | 112 | 120 | **240** |
| hebrew, min | 120 | **240** | 120 | **240** |
| persian, max | 120 | **240** | 120 | **240** |
| persian, min | 2 | 0 | 120 | **240** |
| islamic-civil, max | 104 | 106 | 240 | 240 |
| islamic-civil, min | 10 | 0 | 238 | **240** |
| indian, max | 120 | **240** | 240 | 240 |
| coptic, min | 4 | 124 | 8 | 124 |

islamic-tbla and islamic-umalqura match islamic-civil. ethiopic, ethioaa and coptic near the max go from 120 on `main` to 240 with this PR. gregory, buddhist, japanese and roc are 240 everywhere. With #369, #361 and this PR together, every calendar I scanned is 240/240 at both limits except chinese and dangi (above).

**Ordinary dates.** I ran 10,200 `until()`/`since()` calls (15 calendars; month ends, hebrew leap months, both directions; spans of 1 day to 7000 days). The output with this PR is identical to `main` except for 18 coptic calls starting at `-000500-06-15`, before the coptic epoch. On `main` those were wrong or threw, for example `P1M573078D` or `mixed-sign values not allowed`. With this PR they give `P1M1D`, `-P6M3D` and so on, which round-trip. With #361 underneath, the output is identical with and without this PR. I also ran a grid of 126,720 `until()` calls on #369 + #361, with and without this PR (hebrew, islamic-civil, persian, chinese, coptic, indian; start dates 2022–2025). The output is identical, and every result round-trips.

This PR applies cleanly with #369 and with #361, in either order.

### Note for a later sync of tc39/proposal-temporal 196a31919d (tc39/proposal-temporal#3245)

That commit rewrote this loop to set `next.day = calendarOne.day` on the result of `addMonthsCalendar()`. That object is the one stored in the calendar cache, so the change corrupts cached dates and gives wrong results at ordinary dates upstream. For example, hebrew `2022-05-01` until `2022-03-31` by months gives `-P1M2D` there (Chrome: `-P1M1D`). This PR removes that line's reason to exist. When 196a31919d is ported, its hunk for this loop should be left out in favour of this change. tc39/proposal-temporal#3327 has the details.


ptomato:
Thanks! I'll plan to apply this in proposal-temporal and pull it in here after it's been merged there, when we get to that point with the rebase. Will close it for now, as otherwise the rebases get unmanageable.

PR #371, opened 2026-09-18
Search the last partial window when looking for the next time zone transition

craig-o-curtis:
This is not a port: the fix isn't in tc39/proposal-temporal yet. The same bug is there, and I reported it as tc39/proposal-temporal#3328 with the same change, so a later sync can pick it up. It doesn't depend on #367–#370 or #361.

### What was wrong

`GetNamedTimeZoneNextTransition()` steps forward one search window (19 days by default) at a time. When the next window would end past the maximum instant, it returned `null` without looking at the part of the window that is still in range. So a transition shortly before the maximum was missed. America/Santiago changes offset at `+275760-09-07T04:00:00Z` (`-04:00` → `-03:00`, so local midnight is skipped), six days before the maximum. On `main` (production build):

| Call | `main` | Expected |
| --- | --- | --- |
| `Temporal.Instant.from('+275760-09-01T00:00:00Z').toZonedDateTimeISO('America/Santiago').getTimeZoneTransition('next')` | `null` | `+275760-09-07T01:00:00-03:00[America/Santiago]` |
| `Temporal.PlainDate.from('+275760-09-07').toZonedDateTime('America/Santiago')` | `TypeError: Cannot read properties of null (reading 'sign')` | `+275760-09-07T01:00:00-03:00[America/Santiago]` |
| `Temporal.ZonedDateTime.from('+275760-09-07T12:00[America/Santiago]').startOfDay()` | same TypeError | `+275760-09-07T01:00:00-03:00[America/Santiago]` |
| `Temporal.ZonedDateTime.from('+275760-09-07T12:00[America/Santiago]').hoursInDay` | same TypeError | `23` |
| `Temporal.ZonedDateTime.from('+275760-09-06T12:00[America/Santiago]').hoursInDay` | same TypeError | `24` |
| control: `new Temporal.ZonedDateTime(8640000000000000000000n, 'America/Santiago').getTimeZoneTransition('previous')` | `+275760-09-07T01:00:00-03:00[America/Santiago]` | same |

The expected values are Chrome 153's native Temporal. They also follow from the spec: GetNamedTimeZoneNextTransition returns null only if there is no transition t with t ≤ nsMaxInstant. The TypeError comes from `GetStartOfDay()`, which uses this function when midnight is skipped. Its `castExists()` is removed in production builds, so the `null` reaches JSBI. Development builds throw `Error: Expected arg to be set.` instead.

It isn't limited to the last 19 days. Whether a lookup fails depends on where the 19-day steps from the starting point land. Starting from 00:00Z on each of the last 61 days, `getTimeZoneTransition('next')` in America/Santiago returns `null` on 39 of them, the earliest being `+275760-07-19`.

### What this changes

The last window is clamped to the maximum instant instead of the function returning early (`lib/ecmascript.ts`, a 3-line change plus a comment). `leftMs` still increases on every pass and stops at the maximum, and the existing check after the loop returns `null` as before if no transition was found. The window passed to `bisect()` never ends past the maximum, so neither does the result. Away from the maximum the clamp changes nothing (see Testing).

### Not fixed by this PR

These behave the same with and without this PR:

- **Transitions before 1847:** the transition search starts at `BEFORE_FIRST_DST` (1847-01-01), but Asia/Manila, Pacific/Guam, Pacific/Saipan, Pacific/Kosrae and Pacific/Palau have a transition at the start of 1845. For example, `Temporal.Instant.from('1800-01-01T00:00:00Z').toZonedDateTimeISO('Asia/Manila').getTimeZoneTransition('next')` gives `1899-09-06T12:00:00+08:00[Asia/Manila]`, while Chrome 153 gives `1845-01-01T00:00:00+08:04[Asia/Manila]`. tc39/proposal-temporal `main` behaves the same.
- **`GetNamedTimeZonePreviousTransition()`** has the mirror-image early return at `BEFORE_FIRST_DST`. It has no visible effect today: the earliest transition after 1847-01-01 is Europe/London's on 1847-12-01, far more than one window later.

### Testing

Node 24.21.0 (ICU 78.3): jest 607 passed; eslint, prettier and `tscheck` clean. test262 at the pinned version: intl402/Temporal 330 passed and 8 expected failures, built-ins/Temporal 4330 passed.

**Near the maximum.** For nine zones (America/Santiago, America/New_York, Europe/London, Australia/Sydney, Pacific/Chatham, America/Asuncion, Asia/Tokyo, UTC, Asia/Kolkata), I checked `getTimeZoneTransition('next')` and `('previous')`, the start of day and `hoursInDay` for each of the last 61 days (549 days in all), and compared them with Chrome 153:

| | `main` | this PR |
| --- | --- | --- |
| 19 repro rows (the table above, plus transition, `null` and start-of-day controls in America/Santiago, America/New_York and Asia/Tokyo) | 6 wrong | 0 wrong |
| Last 61 days, 9 zones | 39 wrong (all America/Santiago) | 0 wrong |

The only other differences from Chrome are in the wording of RangeError messages on the last two days, where both throw.

**Ordinary dates.** `next` and `previous` from every day of 2020–2030 in the same nine zones (36,162 lookups) give byte-identical output with and without this PR, and they match Chrome 153. The first transition after 1800 and `previous` from just after it, for all 418 zones in `Intl.supportedValuesOf('timeZone')`, are also byte-identical.

I made the same change in tc39/proposal-temporal at e8cc03fc97 and got the same results there.


ptomato:
Same as the other PR, I'll apply the code in https://github.com/tc39/proposal-temporal/issues/3328 and then it'll get pulled in here in a future rebase. Thanks again and sorry that it's taking a while.

